PR #191, opened 2014-11-23
Promise fixes

juandopazo:
There are still a few issues with the current Promise implementation:
1. It's lacking `Promise.reject`
2. It's not passing two A+ tests:

Resolving a promise with another one is not working correctly:

``` js
var p1 = new Promise(function (resolve, reject) {
  reject(new Error('hello'));
});
var p2 = new Promise(function (resolve) {
  resolve(p1);
});
// p2 should be rejected with an error
```

An infinite loop is caused when resolving a promise with itself:

``` js
var p = Promise.resolve(42).then(function () {
  return p;
});
// p should be rejected
```

I'm trying to fix the resolution of promises, but I'm having a hard time understanding the "promiseChain" logic. @jonathantneal can you explain a bit how it works?

PS: this is obviously failing and should not be merged yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "grunt-cli": "~0.1.6",
     "grunt-mocha": "^0.4.11",
     "grunt-simple-mocha": "^0.4.0",
+    "promises-aplus-tests": "^2.1.0",
     "sauce-tunnel": "^2.1.0",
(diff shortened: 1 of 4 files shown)


juandopazo:
Closing in favor of #257

PR #194, opened 2014-11-26
Update iterables

jonathantneal:
- consistently calculate whole iterable lengths
- replace most references to arraylike with iterable
- correct arraylike bug in Promise
- add Number.EPSILON
- add Number.MAX_SAFE_INTEGER
- add Number.MIN_SAFE_INTEGER


polyfills/Number/detect.js
```diff
@@ -0,0 +1 @@
+Number.MAX_SAFE_INTEGER
```
jonathantneal, line 1: Number will always be available.

jonathantneal:
Whatever happened with this request?

triblondon:
Sorry Jonathan, it's just a bit beyond my understanding, and it seems Juan had an issue with it.  It does also need conflicts resolved.  But I've got not problem with it in principle.

triblondon:
@juandopazo @jonathantneal I'm going to close this request shortly, let me know if you'd like to keep it open.

PR #195, opened 2014-11-26
Update promise

jonathantneal:
This is an alternate pull for #191 that only addresses to the polyfill itself, with a dependency on the update to iterables #194. This update will:
- add Promise.reject
- allow reject on Promise.race
- reference setImmediate from global

All of these fixes were generously provided by @juandopazo. Additionally, I have made some changes so that this update will:
- support promises resolved with promises
- throw promises resolved with themselves


triblondon:
Superceded by #255 then #257

PR #201, opened 2014-11-27
Add test262 tests for Array.isArray

juandopazo:
Obviously everything passes here, with one exception that is commented in the code.

I'm opening this to discuss how to actually integrate this in the repository. As you can see, these are _a lot_ of tests. Other APIs have much more. Should we copy them in the `tests.js` file or should we use a folder?


triblondon:
Ah, now I can.

juandopazo:
Closing this. I think the way forward is running the test262 harness. I'll experiment with that soon.

PR #227, opened 2014-12-19
Support ES6 polyfill sources with 6to5, closes #179

### LINKED ISSUE #179: Transpile ES6->ES5 in browsers that don't support ES6
Since we might conceivably write polyfills or output other code in ES6 syntax, it would be good if the service automatically transpiled code that requires ES6 syntax into ES5 for browsers that don't understand ES6.

Use one of these transpilers:
https://github.com/addyosmani/es6-tools#transpilers

triblondon:
This was actually surprisingly easy, though I cheated by assuming we don't want to serve untranspiled ES6 to any browser.  There's an argument that Firefox 36 might already qualify, having ES6 support that's already better than the 6to5 transpiler.  However, while we can guarantee the consistent logic of the transpiler and cope with language features it doesn't support, browsers with varying degrees of ES6 support are harder to cope with so we should probably have a higher bar.  At this point I don't think there are any browsers that we should serve raw ES6 to.

I've jotted down a few thoughts in a comment about how we could model the more complex set of sources that would result from having ES version as a dimension along with whether to minify and whether to gate (resulting in a minimum of eight versions of every polyfill variant!).  But for now, I think we have a licence to assume we want to level down everything to ES5 before splitting into minified and gated variations.

I also added some better error handling on startup, so that it's easier to spot misconfigured or syntactically invalid polyfills.

### DIFF
--- a/lib/sources.js
+++ b/lib/sources.js
@@ -37,7 +37,7 @@ fs.readdirSync(polyfillSourceFolder).forEach(function (featureName) {
 				browsers: config.browsers,
 				dependencies: config.dependencies || [],
 				licence: config.licence || "",
-				ESVersion: config.ESVersion || undefined
+				esversion: config.esversion || undefined
 			}
 			delete config.browsers;
 			delete config.dependencies;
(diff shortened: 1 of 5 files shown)


lib/sources.js
```diff
@@ -1,99 +1,151 @@
...
-		if (fs.existsSync(detectPath)) {
-			v.detectSource = fs.readFileSync(detectPath, 'utf8').replace(/\s*$/, '') || null;
-			v.rawGatedSource = "if (!(" + v.detectSource + ")) {\n" + v.rawSource + "}\n";
+		// If there is still no default variant, create one
+		if (!('default' in config.variants)) {
+			config.variants.default = {
+				browsers: {},
+				dependencies: [],
+				licence: "",
+				rawSource: "/* This polyfill has no generic form, and no browser specific variant applies to the current user agent. */"
```
samgiles: Can we create an `ESVersion` entry here (even if it is undefined)?

lib/sources.js
```diff
@@ -1,99 +1,151 @@
...
+
+		// Move any default variant (outside of 'variants' object, into the variant object against the 'default' key)
+		// Allows for simple configuration in cases where there is only one variant
+		config.variants = config.variants || {};
+		if (config.browsers) {
+			config.variants['default'] = {
+				browsers: config.browsers,
+				dependencies: config.dependencies || [],
+				licence: config.licence || "",
+				ESVersion: config.ESVersion || undefined
```
samgiles: Can we create a `rawSource` entry here (even if it is undefined)?

polyfills/Math.acosh/config.json
```diff
@@ -14,5 +14,6 @@
 		"op_mob": "*",
 		"op_mini": "*",
 		"safari": "4 - 7"
-	}
+	},
+	"ESVersion": 6
```
samgiles: Can we keep the config keys lower case for consistency?

samgiles:
Need to update the Shrinkwrap to include 6to5.

samgiles:
The config would need to include two pieces of information:  for each variant it would need to specify whether it was written using ES6 syntax, and additionally we'd need to specify the range of browsers that need compilation to ES5 and those that don't.  Take the config for rAF as an example:

``` JS
{
    "aliases": [
        "default",
        "caniuse:requestanimationframe",
        "modernizr:requestanimationframe"
    ],
    "browsers": {
        "bb": "7",
        "chrome": "1 - 9",
        "ie": "6 - 9",
        "ios_saf": "3.2 - 5.1",
        "firefox": "3.6",
        "opera": "9 - 12.1",
        "op_mini": "5 - 7",
        "op_mob": "10 - 12.1",
        "safari": "3.1 - 5.1"
    },
    "esversion": "5",
    "variants": {
        "moz": [{
            "browsers": {
                "firefox": "4 - 22"
            },
            "esversion": "6",
            "requiresversion": "5"
        },
        {
            "browsers": {
                "firefox": "23 - 36"
            },
            "esversion": "6",
            "requiresversion": "6"
        }],
        "webkit": {
            "browsers": {
                "chrome": "10 - 23",
                "safari": "6"
            },
            "esversion": "5"
        }
    },
    "dependencies": [
        "Date.now",
        "performance.now"
    ],
    "license": "requestAnimationFrame polyfill by Erik Möller, fixes from Paul Irish, Tino Zijdel, and Jonathan Neal"
}
```

In this example, the variant for `moz` is written using ES6 syntax and must be compiled to ES5 for '4 - 22' and can remain as ES6 for '23 - 36' (example values).   This doesn't seem too clean to me.

triblondon:
My intention there is to have a separate data file that says which ES syntax each browser understands, something like:

``` json
{
  "ie": {
    "6": "11 - *",
    "5": "<=10"
  },
  "firefox": {
    ...
  }
}
```

I believe there are no syntax incompatibilities between versions 3 and 5, and there are no browsers with sufficient ES6 support to leave ES6 syntax unconverted.  The first stage would be something like the above, and then later we'd need to add 7 too (since there are syntax incompatibilities between 6 and 7)

triblondon:
This was successfully merged via CLI

PR #236, opened 2015-01-06
Faking high precisiong timing in requestAnimationFrame

juandopazo:
Hi! I just got back from some pretty long vacations, so I'm starting to slowly get my feet wet.

The `requestAnimationFrame` polyfill wasn't passing the right argument to the callback function. Vendor versions also send the current timestamp in milliseconds instead of using high precision time. For example, see:

image

This PR fixes this, adds a couple of minor perf improvements (`Date.now` is much faster than `new Date().getTime()`) and adds some basic tests.

### DIFF
--- a/polyfills/performance.now/config.json
+++ b/polyfills/performance.now/config.json
@@ -1,0 +1,15 @@
+{
+	"browsers": {
+		"bb": "<10",
+		"chrome": "<24",
+		"ie": "<10",
+		"ios_saf": "<8 || 8.1",
+		"firefox": "<15",
+		"opera": "<15",
+		"op_mob": "<24",
+		"safari": "<8"
+	},
+	"dependencies": [
+		"Date.now"
+	]
+}
--- a/polyfills/performance.now/detect.js
+++ b/polyfills/performance.now/detect.js
@@ -1,0 +1,1 @@
+window.performance && window.performance.now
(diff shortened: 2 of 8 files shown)


paulirish:
AFAIK calling performance.now was slower than calling Date.now.. i'm not sure if that's still the case.
If it is, can you just use the date.now() polyfill that's already in this repo?

also the repo is lacking a perf.now polyfill.

juandopazo:
I still feel a little weird about polyfill dependencies. But you're right, I'll switch to dependencies and a performance.now polyfill.

Do you think performance is an issue? We're shooting for correctness first.

triblondon:
@juandopazo can you create a branch within the base repo rather than your fork? Then Travis will be able to run the CI tests (it needs credentials to log into Sauce Labs, which are in the FT Travis environment)

juandopazo:
@triblondon I'm getting `Fatal error: Cannot read property 'browserName' of undefined` in the test code.

triblondon:
There's something weird going on at Sauce's end and sometimes the tests hang up before finishing.  I've not looked into it yet because it usually works if you just run the build again.

juandopazo:
It looks like it finally passed.

triblondon:
I'm a bit confused by how your new branch's CI has flagged this branch's PR, but works for me :-)

juandopazo:
@triblondon I didn't do anything. It was all Github magic! :open_mouth:

Issue #233, opened 2014-12-24
Move config, syntax checks, gating and minification to build step


triblondon:
Following #230, it seems sensible to move the following processes to a build step:
- Syntax checks on config.json
- Standardising config format (eg creating default variant)
- Transpiling ES6 -> ES5
- Minification

We could then write out a single file for each polyfill containing the processed sources, which would (currently) include four copies of each polyfill - gated and not, minified and not.  These files can then be read on startup.

Or we could even combine all of that into one monster JSON file.

samgiles:
This seems sensible.  We should run the build process (transpiling, minification, etc) using the 'install' script in the `package.json`.

`config.json` syntax checks should be a separate process run as part of the CI test.

azproduction:
> We should run the build process (transpiling, minification, etc) using the 'install' script in the package.json.

That would be awesome :+1:

PR #243, opened 2015-01-11
Shrinkwrapping, closes #242

### LINKED ISSUE #242: Allow version-locking / shrinkwrapping of polyfill library version in request
Various people including FT and Yahoo devs have requested the ability to ensure that polyfills served by the polyfill service do not change in a manner that is outside of their release process.  This is of course achievable by running your own version of the service, or using it as a library within your own Node stack, but if you are using the public CDN instance, or multiple projects are sharing the same private CDN instance, it is beneficial to be able to upgrade that instance without having to retest integrations with all projects that use the service.

In npm land, this is called shrinkwrapping, so I'll use that term.  There are two possible approaches, depending on the level of granularity required:
1. Version-lock at the module level.  Modifies the existing `features` query param, eg: `features=WeakSet@^1.0.0&Array.prototype.forEach@3.1.2`.
2. Version lock at the library level.  Adds a new querystring param, `libVersion`, eg. `features=default&libVersion=^1.2.0`.

Option 1 allows the most flexibility, and is how the Origami build service works.  However, implementation is significantly complicated by possible dependency conflicts: it's possible that a request for two different features A and B at specific versions will be unsatisfiable because A and B share a dependency and are unable to accept a common version of it.  We would also need to version each polyfill separately, because tracking versions only at the libary level but specifying them at the feature level would likely lead to confusion over the semantics.

I don't really want to touch that solution with a bargepole, so I propose option 2, which is fairly simple to implement:
1. Add a grunt task to clone older versions of the polyfill library into the file tree.  Propose `/polyfills/_versions/{version}/`.  Run this as part of pre-deploy build.
2. Load sources from all versions, with the ones in the root of `/polyfills` being considered 'latest'
3. Add support for a querystring argument `libVersion` on the service, which will pass through to the library
4. Support an option `libVersion` on the `getPolyfillString` API method, which switches to the appropriate library version (or uses latest if unable to find one
5. Report which version of the library is being used in the bundle header comment.

triblondon:
See #242.


tasks/installcollections.js
```diff
@@ -0,0 +1,53 @@
...
+		.then(function() {
+			grunt.log.writeln('Cloning polyfill service source from '+repourl);
+			return exec('git clone '+ repourl +' '+ repodir);
+		})
+		.then(versions.reduce.bind(versions, function(soFar, version) {
+			var dest = path.join(versionsdir, version);
+			return soFar
+			.then(function() {
+				grunt.log.writeln('Installing version '+version);
+				return exec('cd '+repodir+'; git checkout ' + version);
```
matthew-andrews: I think in the options of child_process exec you can specify which direction to run the command in

triblondon: Thanks, wilco

tasks/installcollections.js
```diff
@@ -0,0 +1,51 @@
+
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var Promise = require('es6-promise').Promise;
```
samgiles: Need to use `require('es6-promise').polyfill();` or `Promise` is undefined in denodeify.

samgiles:
Looks good to me.   We will need make the loading of sources not sync (and mostly offline), it would simplify some of this code I think, see #233.

triblondon:
@juandopazo @matthew-andrews what do you think?

Basic gist is:
1. New grunt task `installcollections` lists tags in your local repo, then clones a fresh copy, checks out each of those tags in turn (provided they match a v0.0.0 pattern) and copies the polyfills collection into `/polyfills/__versions/{version}`.
2. New grunt task `shipit` replaces the deploy.sh bash script, and `deploy` wraps installcollections and shipit.
3. Sources module now reads the versions directory and loads all versions of the polyfills.  New public method `getCollection()` returns a reference to the library version that you want, and the latest is statically available at `.latest`.
4. When parsing sources on startup, ignore errors in the older libs (since we improve source validation in later releases)
5. Add `libVersion` param to service and library
6. Document `libVersion` in docs, and add a list of installed versions.

juandopazo:
When is the `deploy` task run? If it's run by your CI, doesn't that conflict with people using the npm package as a library?

triblondon:
Deploy is run manually. People who want to use it as a library probably won't want the shrinkwrap feature since they'll probably just install the version of the module that contains the polyfills that they want.

triblondon:
Deploy wraps installcollections and shipit. For your use you might replace shipit with a task to deploy to your platform.

triblondon:
@matthew-andrews @juandopazo I'll merge this on Tuesday if there are no objections to this implementation.

juandopazo:
I was actually hoping you'd do it in `prepublish` so that I could use the service as an NPM dependency and do `getPolyfillString({ version: '1.0.0', features: ... })`.

triblondon:
Still planning to do that, but thought we could do that as a subsequent step.

juandopazo:
Ok!

PR #244, opened 2015-01-14
Can't set window properties to undefined in IE7/8: "Not implemented", closes #239

### LINKED ISSUE #239: IE7: error thrown
SCRIPT16385: Not implemented
7, line 3 character 11588

Then the code that uses it:
SCRIPT438: Object doesn't support property or method 'addEventListener'

triblondon:


triblondon:
@jonathantneal this fixes a bug I think you introduced in a commit you made in https://github.com/Financial-Times/polyfill-service/commit/d59c718728f04ee8805725f2a419f1fe3e3333e0.  I have found that IE7 and 8 don't allow assignment of undefined to a window property, so this is my suggestion, but I am not sure why you needed to set the property to undefined in the first place, so would be grateful for your review.

jonathantneal:
The undefined assignment is there to trip checks for `'onhashchange' in window`, so this will suffice. Bummer about that bug, though!

triblondon:
Ah, makes sense.  Thanks for the review!

PR #245, opened 2015-01-14
String.prototype.contains is now includes, closes #240

### LINKED ISSUE #240: String.prototype.contains -> String.prototype.includes
String.prototype.contains has been changed in the ES6 spec to includes: http://people.mozilla.org/~jorendorff/es6-draft.html#sec-string.prototype.includes

More information here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#String.prototype.contains

triblondon:


samgiles:
The spec says `String.prototype.contains` should throw an exception if the first argument is a RegExp.  Other than that the algorithm is identical to `string.indexOf(searchString) !== -1`

triblondon:
@samgiles updated, can you review again?

samgiles:
Looks good.  I think we should include a contributing guide somewhere that documents all the polyfills that require tests and rank them based on difficulty, these `String` polyfills would be good candidates for beginners.

PR #246, opened 2015-01-14
Polyfill fetch API, closes #109

### LINKED ISSUE #109: Fetch API Polyfill
It would be great if this service was able to serve a polyfill for the [Fetch API](https://fetch.spec.whatwg.org/) (XHR with Promise interface).

Unfortunately I highly suspect one doesn't yet exist.  Am I wrong, does one already exist?  (Maybe @annevk could confirm?)

If not, then I'm not sure how to proceed...

triblondon:
Early draft.  Don't merge yet.

### DIFF
--- a/polyfills/fetch/config.json
+++ b/polyfills/fetch/config.json
@@ -1,0 +1,22 @@
+{
+	"aliases": [
+		"es6"
+	],
+	"browsers": {
+		"ie": "*",
+		"firefox": "*",
+		"opera": "*",
+		"safari": "*",
+		"chrome": "*",
+		"ios_saf": "*"
+	},
+	"dependencies": [
+		"Array.prototype.forEach",
+		"Object.getOwnPropertyNames",
+		"Promise",
+		"XMLHttpRequest"
+	],
+	"notes": [
+		"Uses FileReader and Blob if they exist, and depends on FormData"
+	]
+}
--- a/polyfills/fetch/config.json
+++ b/polyfills/fetch/config.json
@@ -1,7 +1,4 @@
 {
-	"aliases": [
-		"es6"
-	],
 	"browsers": {
 		"ie": "*",
 		"firefox": "*",
--- a/polyfills/fetch/config.json
+++ b/polyfills/fetch/config.json
@@ -16,6 +16,8 @@
 		"Promise",
 		"XMLHttpRequest"
 	],
+	"author": "https://github.com/github/fetch/graphs/contributors",
+	"licence": "https://github.com/github/fetch/blob/master/LICENSE",
 	"notes": [
 		"Uses FileReader and Blob if they exist, and depends on FormData"
 	]
--- a/polyfills/fetch/polyfill.js
+++ b/polyfills/fetch/polyfill.js
@@ -1,0 +1,258 @@
+(function() {
+  'use strict';
+
+  if (self.fetch) {
+    return
+  }
+
+  function Headers(headers) {
+    this.map = {}
+
+    var self = this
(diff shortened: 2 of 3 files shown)


triblondon:
A few questions here for @matthew-andrews or @josh:
- Seems like this polyfill depends on `FormData` which is only in IE as of 10. Are you using a polyfill for that?  (eg https://github.com/francois2metz/html5-formdata)
- There's also a soft dependency on `Blob` and `FileReader`.  Is it safe to say these basically aren't polyfillable?  What's the expected impact of using fetch to load blobs in a browser that doesn't support them?
- Testing this is mega hard.  Josh has written a server just to do it, which makes for a very bespoke test harness.  The polyfill service can run Mocha / expect tests.  Do you have any suggestions on how we should test fetch?

josh:
> Seems like this polyfill depends on FormData which is only in IE as of 10. Are you using a polyfill for that?

`FormData` is a soft dependency like `Blob`. If you're host environment doesn't support it, `response.formData()` will just be `undefined`.

I don't know if I would actually recommend using that polyfill since it can't actually provide file upload support. If you need IE 9, you'll probably just not supporting the fancy form upload stuff in that case. Thats basically how github.com deals with it.

> Is it safe to say these basically aren't polyfillable? 

`Blob` _maybe_, though people are reporting that they using some `Blob` polyfill that corrupts utf8 data. So clearly bad ones exist.

> What's the expected impact of using fetch to load blobs in a browser that doesn't support them?

Similar to `FormData`. `response.blob()` will just be `undefined`. And since you have no way to construct a blob, theres no way you could pass a blob to fetch `fetch(body: someBlobWhereWouldItComeFromInIE9)`.

> Testing this is mega hard. Josh has written a server just to do it, which makes for a very bespoke test harness. The polyfill service can run Mocha / expect tests. Do you have any suggestions on how we should test fetch?

Yeah, pretty much anything doing requests will need an actual server. No simple static page. Maybe considering pushing the tests to a Heroku instance just so people can hit the test page publicly. What do you think about that?

polyfills/fetch/polyfill.js
```diff
@@ -0,0 +1,281 @@
+/*
+Copyright (c) 2014 GitHub, Inc.
```
josh, line 2: I regret not making this public domain to begin with.

polyfills/fetch/config.json
```diff
@@ -0,0 +1,24 @@
+{
+	"aliases": [
+		"es6"
```
matthew-andrews: i don't think `fetch` is ES6?

juandopazo:
You can launch a local server using Grunt in the `test` task.

monolithed:
What are we waiting for, guys? :)

juandopazo:
BTW I recenly tried something like this with grunt-simplemocha and [grunt-run](https://www.npmjs.org/package/grunt-run) and it worked pretty well. You can define a `functional` task as:

``` js
grunt.registerTask('functional', [
  'start:polyfill-service',
  'simplemocha:functional',
  'stop:polyfill-service'
]);

grunt.registerTask('test', ['unit', 'functional']);
```

polyfills/fetch/config.json
monolithed: `fetch` is not a part of ES6. 
It's a [host object](https://fetch.spec.whatwg.org/) like WebSoket, FormData and so on.

triblondon:
OK, since this is a straight copy of Josh's polyfill, let's merge this and create a new issue (#267) for testing it.

PR #249, opened 2015-01-21
Matches ie mob

matthew-andrews:
Caught by error aggregation on Next:-
https://app.getsentry.com/nextftcom/ft-next-frontend/group/44609192/

Also IE11 Mobile doesn't seem to return very many polyfills, have we tested it on a real device?

https://cdn.polyfill.io/v1/polyfill.js?ua=Mozilla/5.0%20(Mobile;%20Windows%20Phone%208.1;%20Android%204.0;%20ARM;%20Trident/7.0;%20Touch;%20rv:11.0;%20IEMobile/11.0;%20NOKIA;%20Lumia%20520)%20like%20iPhone%20OS%207_0_3%20Mac%20OS%20X%20AppleWebKit/537%20(KHTML,%20like%20Gecko)%20Mobile%20Safari/537


polyfills/Element.prototype.matches/config.json
```diff
@@ -22,6 +22,11 @@
 				"ie": "9 - *"
 			}
 		},
+		"ie_mob": {
+			"browsers": {
+				"ie": "*"
```
triblondon: Should this be `ie_mob`?

triblondon:
Probably lazy targeting.  No, we don't have any IE mobile test browsers, either readily to hand or on Sauce Labs.

matthew-andrews:
Oops — check now?

PR #254, opened 2015-01-24
Update replace method as replaceWith

### LINKED ISSUE #248: Element.prototype.replace has been renamed to `replaceWith`
https://github.com/whatwg/dom/commit/b7563aaf0864c8d104d18c36a9eda036c5205131

/cc @jonathantneal

jonathantneal:
Follow change from whatwg/dom@b7563aa
Update reference comments

Resolves #248

### DIFF
--- a/polyfills/Element.prototype.mutation/detect.js
+++ b/polyfills/Element.prototype.mutation/detect.js
@@ -3,4 +3,4 @@ Element.prototype.prepend &&
 Element.prototype.append  &&
 Element.prototype.before  &&
 Element.prototype.after   &&
-Element.prototype.replace
+Element.prototype.replaceWith
--- a/polyfills/Element.prototype.mutation/polyfill-blink.js
+++ b/polyfills/Element.prototype.mutation/polyfill-blink.js
@@ -46,8 +46,8 @@
 		}
 	};
 
-	// http://dom.spec.whatwg.org/#dom-childnode-replace
-	Element.prototype.replace = function replace() {
+	// http://dom.spec.whatwg.org/#dom-childnode-replacewith
+	Element.prototype.replaceWith = function replaceWith() {
 		if (this.parentNode) {
 			this.parentNode.replaceChild(mutation(arguments), this);
 		}
--- a/polyfills/Element.prototype.mutation/polyfill.js
+++ b/polyfills/Element.prototype.mutation/polyfill.js
@@ -46,8 +46,8 @@
 		}
 	};
 
-	// http://dom.spec.whatwg.org/#dom-childnode-replace
-	Element.prototype.replace = function replace() {
+	// http://dom.spec.whatwg.org/#dom-childnode-replacewith
+	Element.prototype.replaceWith = function replaceWith() {
 		if (this.parentNode) {
 			this.parentNode.replaceChild(mutation(arguments), this);
 		}


jonathantneal:
I would like to address another issue, too. These polyfills should target childnode and parentnode elements, which in HTML roughly translate to:

[ParentNode](https://dom.spec.whatwg.org/#interface-parentnode) => `Document`, `DocumentFragment` (`HTMLDocument` in IE8), `Element`

[ChildNode](https://dom.spec.whatwg.org/#childnode) => `Element`, `Text`, `HTMLCommentElement`, `DocumentType` (`HTMLCommentElement` in IE8)

mislav:
How does polyfill-service handle backwards compatibility? I.e. developers were getting `replace()` method polyfilled and possibly started to use it in their app code. When you ship this rename, their `replace()` calls will suddenly stop working.

triblondon:
Jonathan, I'm stuck on your comment https://github.com/Financial-Times/polyfill-service/pull/254#issuecomment-71334297, can you give a bit more detail?

Also, do you think we should split out these individual APIs into separate features within the polyfill service?  Otherwise it's hard to know that they're actually covered, hence my confusion in #248 (though as a maintainer of the service I should have known better)

@mislav We have a feature being released in the next version that enables shrinkwrapping of library versioning, so you can keep the same polyfills through a service upgrade, which will help risk-averse users who want to use the service.  But I take the point that in this case it would be pretty easy to support both, and I wonder if we should keep a `Element.prototype.replace` definition, with an alias to `Element.prototype.replaceWith`.  @jonathantneal any thoughts on that?

jonathantneal:
Splitting out each method would be helpful from an organizational standpoint. However, there is a [low-level API](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Element.prototype.mutation/polyfill.js#L3-L23) required by all of these polyfills that is not exposed to the browser. How do we solve this problem?

One solution might be to expose these low-level APIs to polyfills as private methods.

``` json
"private": {
    "mutation": "mutationMacro"
}
```

``` js
Element.prototype.prepend = function prepend() {
    this.insertBefore(mutation(arguments), this.firstChild);
};
```

Which compiles to something like:

``` js
(function (mutation) {
    Element.prototype.prepend = function prepend() {
        this.insertBefore(mutation(arguments), this.firstChild);
    };
})(function theMutationMacroMethod() {});
```

Array polyfills would benefit from separating themselves from the api that calculates the maximum value of a `length` property. The Promise polyfill would benefit from separating themselves from the api that creates asynchronous tasks. Any polyfill wrapped by a function could be unwrapped, improving its readability.

What do you think?

I brought up a similar issue earlier. There are also low-level objects. My earlier comment about the `ParentNode` object may have confused you. `ParentNode` is a kind of low-level object that does not actually exist in the browser. It represents `Document`, `DocumentFragment` and `Element` simultaneously.

``` js
ParentNode.prototype.prepend = function prepend() {};
```

Except I see no sane way to compile that into what is required:

``` js
Document.prototype.prepend = DocumentFragment.prototype.prepend = Element.prototype.prepend = function prepend() {};
```

Ideas? Do my previous comments make more sense now?

triblondon:
OK, I think I get it.  So on ParentNode, you're saying that the current form of the polyfill doesn't work when invoked on Document, but the native impl would do, so we need to support that.  You'd like to be able to do that via a single assignment, but there's no native global that refs all the constructors you want, so you'd have to use multiple assignments.  If that's right, couldn't we just do the multiple assignments?  Seems easiest, clearest and most robust to me.  Opened a new issue for it: #281.

On splitting the polyfill, the shared private method is a tad annoying, and it's true that this is a common pattern.  I think this is important enough to start a separate issue about, rather than holding up this PR. See #280.

Finally, on the subject of `replace` vs `replaceWith`, I'm moving towards the view that since `replace` was never adopted by any browser natively, we should not continue to support it.  Instead, people who want to use bleeding edge APIs may prefer to shrinkwrap the version of the polyfill library that they use.

Issue #126, opened 2014-10-17
Promise polyfill is not a promise polyfill


juandopazo:
Returning `this` from `then` is not a valid behavior for a promise polyfill. It´s also lacking `Promise.race`, `Promise.resolve` and `Promise.reject`. I suggest using https://github.com/jakearchibald/es6-promise.

monolithed:
#### jonathantneal's:

Promises/A+ 266 passing / 593 failing

#### Octane's

https://github.com/Octane/Promise
Promises/A+ 872 passing + 55 own tests

 I suggest using Octane's polyfill

jonathantneal:
I’m game with that, but your numbers don’t add up: 266 + 593 = 859 ... vs 872? + 55 own tests?

Would you setup something like you’ve done with https://octane.github.io/promise/ to show us the 859 to 872 plus 55 tests? Depending on the severity difference (like, if all your failing tests hang on one feature) we can fix ours or adopt yours.

Octane:
Hello.
The [Promise polyfill](https://github.com/Octane/Promise) is passing 872 tests of the [Promises/A+ compliance test suite](https://github.com/promises-aplus/promises-tests) [![Build Status](https://travis-ci.org/Octane/Promise.svg?branch=master)](https://travis-ci.org/Octane/Promise)

![passing 872 tests](https://dl.dropboxusercontent.com/u/8763532/screenshots/promise872tests.png)
try it

```
$ git clone https://github.com/Octane/Promise.git
$ cd Promise
$ npm install
$ npm test
```

I wasn't trying to run mocha-based Promise/A+ tests in a browser, but I have [55 browser-based tests](https://octane.github.io/promise/tests/browser.html)  in which there are tests for `Promise.all` and `Promise.race` (missing in the test suite).

Supports IE8+

––––––––––––––––––––––––––––––
UPD: I replaced my polyfil to your ([modified for tests](https://dl.dropboxusercontent.com/u/8763532/promise.js)) and got the result:
![your polyfill results](https://dl.dropboxusercontent.com/u/8763532/screenshots/promise593failing.png)

jonathantneal:
That’s awesome. I’m sure Juan or someone will setup some tests so we can take a look. Great work, both of you.

juandopazo:
Octane's promise looks pretty good. The implementation is straightforward and readable. Its approach is similar to my own version, so I obviously like it! :wink: 

[...]

Issue #88, opened 2014-09-12
Avoid duplicating polyfills already maintained elsewhere?


triblondon:
@jdalton raised on https://github.com/Financial-Times/polyfill-service/issues/86#issuecomment-55415462 the issue that we are duplicating work by reimplementing polyfills maintained by others elsewhere, such as the es6shim project.

The reasons this project exists are that things like es6shim:
- are not consistently packaged
- polyfill not just the thing they are for but all their dependent APIs too, resulting in multiple polyfills on the same page potentially duplicating the polyfilling of lower-level APIs
- require developers to know that a polyfill is needed
- typically do not offer an easy method of only shipping the polyfill to browsers that lack a native implementation.

However, it makes sense to avoid duplicating work, and we don't want to upset anyone who has the generosity to publish and maintain these polyfills, so I'm open to suggestions for how we could reach our objectives in a way that is more effective for the whole community.

triblondon:
There's now an updated policy on this, see https://github.com/Financial-Times/polyfill-service/pull/338

PR #255, opened 2015-01-24
Update/promise

jonathantneal:
- Update polyfill
- Update license
- Address #126
- Add strict type checking
- Use existing setImmediate fallback
- Follow voice of library (totally minor, only concerned if it causes problems with the actual polyfill):
  - whitespace consistency (so very minor, the original has an instance where, like, one of the `catch` statements did not have a space before it)
  - use `index` over `i` (the original actually used `index` once, too, but `i` otherwise)
  - combine `var` statements
  - prefix ++ on increment operator
  - use `while` over `for`
  - add line breaks between thought changes

I am concerned about a [for loop](https://github.com/Octane/Promise/blob/master/promise.js#L271-L294) in @Octane’s original polyfill. What should `promiseCount` be at the time of then’s execution? Could this be written to avoid the reference to either the synchronous or asynchronous value of `promiseCount` within the while loop?

### DIFF
--- a/polyfills/Promise/config.json
+++ b/polyfills/Promise/config.json
@@ -15,7 +15,6 @@
 		"opera": "* - 19",
 		"safari": "* - 7"
 	},
-	"dependencies": [
-		"Object.defineProperty"
-	]
+	"license": "MIT",
+	"author": "Dmitry Korobkin (https://github.com/Octane/Promise)"
 }
--- a/polyfills/Promise/polyfill.js
+++ b/polyfills/Promise/polyfill.js
@@ -1,28 +1,17 @@
-(function (global) {'use strict';
-
-	var STATUS = '[[PromiseStatus]]';
-	var VALUE = '[[PromiseValue]]';
-	var ON_FUlFILLED = '[[OnFulfilled]]';
-	var ON_REJECTED = '[[OnRejected]]';
-	var ORIGINAL_ERROR = '[[OriginalError]]';
-	var PENDING = 'pending';
-	var INTERNAL_PENDING = 'internal pending';
-	var FULFILLED = 'fulfilled';
-	var REJECTED = 'rejected';
-	var NOT_ARRAY = 'not an array.';
-	var REQUIRES_NEW = 'constructor Promise requires "new".';
-	var CHAINING_CYCLE = 'then() cannot return same Promise that it resolves.';
-
-	var setImmediate = global.setImmediate || function (func) {
-		var args = slice.call(arguments, 1);
-
-		return setTimeout(function () {
-			func.apply(null, args);
-		});
-	};
-	var isArray = Array.isArray || function (anything) {
-		return Object.prototype.toString.call(anything) == '[object Array]';
-	};
+(function (global, setImmediate, isArray) {
+	var
+	STATUS = '[[PromiseStatus]]',
+	VALUE = '[[PromiseValue]]',
+	ON_FUlFILLED = '[[OnFulfilled]]',
+	ON_REJECTED = '[[OnRejected]]',
+	ORIGINAL_ERROR = '[[OriginalError]]',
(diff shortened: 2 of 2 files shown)


polyfills/Promise/config.json
```diff
@@ -15,7 +15,6 @@
 		"opera": "* - 19",
 		"safari": "* - 7"
 	},
-	"dependencies": [
-		"Object.defineProperty"
```
juandopazo: I'm guessing this should include `setImmediate` and `Array.isArray`.

juandopazo:
@jonathantneal no issue in that `for` loop. At the time the promise calls the `then` callback, it checks if all the promises have been fulfilled. It does that by counting how many were fulfilled and checking against `promiseCount` which is the count of promises in the array (which is not the length of the array because of holes). Since all `then` callbacks are called asynchronously, the value of `promiseCount` will be the same for all of them.

jonathantneal:
@juandopazo, good, that’s what I hoped. I think it would be better to move the function outside the loop to remove the ambiguity. Any concerns?

``` js
Promise.all = function (values) {
    return new Promise(function (resolve, reject) {
        var
        fulfilledCount = 0,
        promiseCount = 0,
        index = -1,
        anything, length, value;

        function onFulfilled(index) {
            return function (value) {
                values[index] = value;

                ++fulfilledCount;

                if (fulfilledCount === promiseCount) {
                    resolve(values);
                }
            };
        }

        if (isArray(values)) {
            values = values.slice(0);
            length = values.length;

            while (++index < length) {
                value = values[index];
                anything = toPromise(value);

                if (isPromise(anything)) {
                    ++promiseCount;

                    anything.then(onFulfilled(index), reject);
                } else if (isInternalError(anything)) {
                    reject(anything[ORIGINAL_ERROR]);
                } else {
                    //[1, , 3] → [1, undefined, 3]
                    values[index] = value;
                }
            }

            if (!promiseCount) {
                resolve(values);
            }
        } else {
            reject(new TypeError(NOT_ARRAY));
        }
    });
};
```

juandopazo:
None at all. But wouldn't you rather copypaste the code or import it using a Grunt task or something like that instead of maintaining your own fork of it?

triblondon:
Looks OK to me and passes our existing promise tests, such as they are.  I agree with Juan that these changes to @Octane's code are probably not worth the additional maintenance responsibility that they impose on us, especially when we want to pull in newer versions of the source.  There was a lengthy discussion of this problem in https://github.com/Financial-Times/polyfill-service/issues/88, where Mathias and JDD were strongly of the view that we should not even have a copy of the polyfill locally, but it should come in as a dependency at build time.

I think the compromise for these third party libs might be:
- Avoid editing (but by all means submit PRs upstream)
- Extend config.json with the following new properties:
  - `repo`: The GitHub (or other) URL of the root of the repo
  - `version`: The tag at which the current source was captured
- Add a grunt task to update all the sources that have these two properties

This would mean we can continue to do QA through our own test suite, it's compatible with our new shrinkwrapping/version-locking feature, and updates from the original source are automatable.  If we accept this principle though, in the case of this PR we'd have to be happy to take @Octane's polyfill exactly as it comes.

But maybe that would mean that we could have more time to spend on caring for the style and performance of the polyfills that are not imported from third parties?

triblondon:
There's a bug in the setImmediate polyfill included in this - I've pushed a fix and also added @matthew-andrews's test case to the test suite.

However, we already have a setImmediate polyfill in the collection, and @Octane's source doesn't embed it, so could we use our polyfill here by listing `setImmediate` as a dependency?

triblondon:
Seems I can't push my commits to this branch because it's capitalised, and my git client has lowercased it, and now my Mac refuses to tell the difference because it has a case insensitive filesystem.  Can we continue this in #257.

PR #257, opened 2015-01-26
New promise polyfill

triblondon:
Resumed from #255

### DIFF
--- a/polyfills/Promise/config.json
+++ b/polyfills/Promise/config.json
@@ -15,7 +15,6 @@
 		"opera": "* - 19",
 		"safari": "* - 7"
 	},
-	"dependencies": [
-		"Object.defineProperty"
-	]
+	"license": "MIT",
+	"author": "Dmitry Korobkin (https://github.com/Octane/Promise)"
 }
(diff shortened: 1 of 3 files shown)


triblondon:
So I think actions here primarily on @jonathantneal:
- Do you want to do [your Promise.all refactor](https://github.com/Financial-Times/polyfill-service/pull/255#issuecomment-71342545)?  (and could we maybe consider pushing it upstream into @Octane's version?)
- Is my fix for the setImmediate embedded polyfill OK with you (and should we drop it entirely and use the canonical setImmediate polyfill as a dependency?)

And any more general feedback on the ideas for tracking 3rd party sources that I put in https://github.com/Financial-Times/polyfill-service/pull/255#issuecomment-71439382 before we adopt this PR?

juandopazo:
If you're not going to use ASAP, then using the dependency is probably the best idea. I mean, the way it's written people can include the `setImmediate` polyfill in their request and they'll get relatively fast promises. But it's not documented anywhere and few people will know how to speed them up, so you might get some cases of people complaining "why is this soo slooow?".

triblondon:
@juandopazo I don't have an opinion on ASAP.  As this PR currently stands it seems an improvement on what we already have, so can we merge this and then we can resolve the issue of spawning async tasks as a separate issue (which I'm happy to open)?

juandopazo:
@triblondon sure!

PR #262, opened 2015-01-29
Fix API docs

### LINKED ISSUE #256: Where polyfills?
Hi,

``` js
var polyfill_service = require('polyfill-service');


var polyfills = polyfill_service.getPolyfillString({
    uaString: 'MSIE 9.0',

    features: [
        {
            name : 'Object.assign',
            flags: ['always', 'gated']
        }
    ]
})
```

```
➜ node polyfill-service.js 
true

/* For detailed credits and licence information see http://github.com/financial-times/polyfill-service.
 * 
 * UA detected: ie/9.0.0
 * Features requested: 0
 *  */
```

What am I doing wrong? :)

`"polyfill-service"  : "^1.1.0"`

juandopazo:
The "features" option was actually supposed to be an object, not an array. Fixes #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ For an HTTP API reference, see the [hosted service documentation](http://polyfil
 
 Polyfill service can also be used as a library in NodeJS projects.  To do this:
 
-1. Add this repo as a dependency in your package.json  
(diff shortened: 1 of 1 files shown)


samgiles:
Thanks :)

PR #263, opened 2015-01-29
Error handling for Object.defineProperty

juandopazo:
`Object.defineProperty` was not checking if accessors were available in the object. It also was not throwing when both an accessor and a value were specified in the descriptor.

### DIFF
--- a/polyfills/Object.defineProperty/polyfill-ie7.js
+++ b/polyfills/Object.defineProperty/polyfill-ie7.js
@@ -10,20 +10,37 @@ Object.defineProperty = function defineProperty(object, property, descriptor) {
 	}
 
 	var
+	ERR_ACCESSORS_NOT_SUPPORTED = 'getters & setters can not be defined on this javascript',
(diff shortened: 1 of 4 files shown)


jonathantneal:
:+1:

juandopazo:
I'm waiting for the Sauce tests. I'm guessing it needs some changes in the browser specific variants too.

juandopazo:
Yeah just what I thought.

capture

juandopazo:
I'm confused by the approach taken for this polyfill. In browsers that are not IE but don't have getters and setter it tries to use `__defineGetter__`. If there's no `__defineGetter_` it'll throw. However, in old IE if it doesn't support getters and setters it'll fail silently. What would be the correct behavior? Personally I think it should throw when it finds an accessor and the browser doesn't support them. It should be helpful for debugging.

juandopazo:
Ok. Updated everything. There's a lot of error checking code. So much that I'm wondering if it even makes sense to have different versions for IE7 and IE 8...

PR #264, opened 2015-01-29
Use es5-shim's Object.create

juandopazo:
Ok, this one might be controversial :sweat_smile:.

The current `Object.create` implementation covers fewer edge cases. But my main motivation for updating it is that it´s using `eval` which won't work in CSP enabled environments.

So I'm copying the one from es5-shim. What do  you think?


triblondon:
I think this is one for @jonathantneal...

triblondon:
@juandopazo I'm happy with this and am intending to ship a 1.2 release today, if you're in a hurry for this feature let me know and I'll merge it.

juandopazo:
Not in a hurry. I think it's much less controversial than #263. I'd love to get some feedback on that one.

triblondon:
Pushed an update to this branch so we can see if it passes the current test suite.

triblondon:
This doesn't pass the current test suite.  Will stick with current impl - Juan feel free to pick this up if you'd like to fix the CSP issue and can follow up the test failures.

PR #269, opened 2015-01-31
Move source compilation to a build step, closes #233

### LINKED ISSUE #233: Move config, syntax checks, gating and minification to build step

triblondon:
OK, here's what we have here:
- Created new grunt task 'buildsources' which reads the polyfills dir and produces two files, sources.json and aliases.json in the polyfill directory
- Run that task on install, and also as part of the test and deploy tasks (so a failure to build sources is a CI fail)
- Moved the Promise polyfill so that promises are available to grunt tasks and when the service is being used as a library (this was working before because I was using a 0.11 version of Node)
- Refactored and updated gitignore so that the historic polyfills and the new compiled sources and aliases files are not committed
- Moved the feature-gating logic to run on demand rather than when sources are compiled, to reduce the footprint of and avoid repetition of code in the compiled sources file.  A side effect of this is that minified outputs contain some line breaks because the feature-detect code is added post-minification.  Not a big deal.

@samgiles suggested that the compilation be two tasks, one for install, one for test, but I tend to disagree - the only impact of keeping it in one is that if the polyfill code is invalid, the service won't start, which actually seems like a reasonably good thing, as well as being a bit simpler.

What do you think?


triblondon:
@samgiles could you review this?

samgiles:
Ah, sorry, I did review this but forgot to actually press 'Comment'.

This looks good, I think I still prefer the idea of having the tasks separate but can't remember the concrete reason why.  This makes sense as it is now.

PR #275, opened 2015-02-04
Add 'all' alias mapping to entire feature set, closes #258

### LINKED ISSUE #258: Create a feature group that includes all polyfills
Since `http://cdn.polyfill.io/v1/polyfill.js?features=es6` is a thing, it would be nice to have `http://cdn.polyfill.io/v1/polyfill.js?features=all` which includes all polyfills available instead of just the default or es6 groups.

You should also create a list of all the feature groups and what they include.

triblondon:


triblondon:
@samgiles code review?

PR #276, opened 2015-02-04
Support 'unknown' param, closes #252.  Tidy and reorg service/lib interface

### LINKED ISSUE #252: Unsupported browsers
The service is returning an empty response for UA strings it can´t parse. What´s the rationale behind this? Shouldn't it default to serving the polyfills?

triblondon:


triblondon:
@juandopazo could you let me know if this works for you?

juandopazo:
Looks great! I'll try to set up monitoring to see if we run into those kinds of issues you mentioned with big bundles.

PR #278, opened 2015-02-06
Removing the protocol from the script call in the docs so it can be served over http and https respectively

danielcgold:


matthew-andrews:
does it fail if you load over https on a http page?

triblondon:
This code was originally as you are suggesting, but it was changed to be always-https, because there's no reason not to.  You can load an HTTP page with HTTPS subresources, and I don't know of any browser that complains about this.  Feel free to reopen with an example if you know of one!

PR #279, opened 2015-02-07
Set separate jshint standards for the node app and the polyfills library

triblondon:
I haven't had jshint turned on up to now, and I wanted to fix that, but I realised that the current jshintrc config targets polyfills, not server code.  So I created a second jshintrc in the polyfills directory to target polyfill code, and modified the original to be more appropriate to the node app.

The settings for the app are FT standard, and I've pretty much kept Jonathan's suggested settings for the polyfills.


triblondon:
@samgiles could you review?  Most important is the two jshintrc files, the rest is updating the service code to conform to the new standard.

triblondon:
@jonathantneal said on chat this evening that this has his OK.

jonathantneal:
Yes, it does. I especially love the `globalstrict`.

.jshintrc
```diff
@@ -1,6 +1,17 @@
...
+	"unused": "vars",
 	"boss": true,
+	"esnext": true,
+	"expr": true,
+	"sub": true,
+	"node": true,
 	"strict": false,
-	"predef": ["Window", "HTMLDocument", "HTMLElement"]
+	"globalstrict": true,
+	"curly": false
```
samgiles: This doesn't seem to be in the spec: http://origami.ft.com/docs/syntax/js/#syntax-convention-rules

Disclaimer: I'm anti-single line branching statements.

samgiles:
~I only see one `.jshintrc`?~

Ignore me

samgiles:
Bar inline comment, LGTM.

Issue #280, opened 2015-02-09
Support private dependencies


triblondon:
In #254, @jonathantneal makes the point that while our current `Element.prototype.mutation` polyfill would benefit from being split up into separate features for replace, append, prepend, before and after, all of these would require exactly the same private mutation code.

Therefore it may be worth considering a mechanism by which functions that are not standardised web APIs can be treated as separate features by the service on which real features can depend.

An easy way to do this would be to:
1. Wrap the entire polyfill output in a closure and define a `PolyfillService` variable at the top of it, initialised to `{shared:{}}`.
2. Create private features by using the `PolyfillService.shared` namespace
3. In the polyfill.js for the private feature, assign the exported function to `PolyfillService`, eg `PolyfillService.shared.mutation = function(...) { ... };`
4. In the config.js for the real feature, add the private feature to the normal dependency list
5. In the polyfill.js for the real feature, use `PolyfillService.shared.mutation`

In terms of the service functionality, the only change needed here is the outer wrapper closure, which is only needed if you want to avoid polluting the global scope with the `PolyfillService` object.

Jonathan has a slightly different proposal in #254.

jonathantneal:
Initial thoughts:
- We should avoid polluting the global scope.
- An automated task can assemble closures / Polyfills should be closure-free.
- We could use `config.json` or create a `closure.js` file to assign private methods and variables.

If the polyfill service wraps every response in a closure, then this becomes even easier.

triblondon:
@jonathantneal OK, so are you satisfied with the approach I outlined in the original post above?  So there's one big closure around the whole polyfill bundle, rather than individual closures around each polyfill.  Dependencies can just be named clearly using a prefix: `PolyfillService.shared.thing`, and listed as normal dependencies - don't see a strong argument for a new config property, and if we use dependencies it will just flow through the existing dep resolution logic.

triblondon:
Just to be clear, the `PolyfillService` variable lives inside the outer closure, so is not a global.

jonathantneal:
That’s great. Could you explain the reason behind the `PolyfillService` object?

We have an example of the wrapped closure in [Array.isArray](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Array.isArray/polyfill.js).

``` js
(function (toString) {
    Object.defineProperty(Array, 'isArray', {
        configurable: true,
        value: function isArray(object) {
            return toString.call(object) === '[object Array]';
        },
        writable: true
[...]

PR #287, opened 2015-02-09
Add enqueueMicrotask

### LINKED ISSUE #266: Promise polyfill is too slow
Following discussions in https://github.com/Financial-Times/polyfill-service/pull/255 and https://github.com/Financial-Times/polyfill-service/issues/126, we adopted a new Promise polyfill, but left the following issue unresolved: The polyfill relies on [setImmediate](https://developer.mozilla.org/en-US/docs/Web/API/Window.setImmediate) which is only implemented by IE and not expected to be adopted by any other browser.  It contains an embedded implementation of setImmediate which is slower than alternatives such as [ASAP](https://github.com/kriskowal/asap).  The polyfill service also already has a separate implementation of setImmediate offered as a standalone polyfill for that feature.

Options here include:
1. Continue to support setImmediate with one of the implementations we already have, remove the other (such that the surviving implementation is in the setImmediate polyfill.js rather than embedded in Promise's polyfill.js), and make setImmediate a dependency of Promise.
2. Continue to support setImmediate with a new implementation based on asap (or other), make that a dependency of Promise, and remove the embedded implementation within the Promise polyfill.js
3. Drop support for setImmediate, and use a scoped function within the Promise polyfill to provide the desired async task spawning, possibly by embedding asap.js

jonathantneal:
- Resolves #266
- Requires resolution to #280


triblondon:
Thanks Jonathan!  Looks like the setTimeout variant isn't configured.  We'll probably need some further changes to this after we agree an approach for #280.

jonathantneal:
I was getting the best speed tests with `setInterval`, but I left `setTimeout` in there because it’s closer to how ASAP did it. It has been removed.

Issue #286, opened 2015-02-09
Mutation methods polyfill overwrites native implementations


mislav:
[`Element.prototype.mutation`](https://github.com/Financial-Times/polyfill-service/blob/57cbada7f0f1882eb200b4ff3a02d8e926bd1dba/polyfills/Element.prototype.mutation/polyfill.js) polyfill implements these DOM methods: `append`, `prepend`, `before`, `after`, `replaceWith`, `remove`. It does so without checking whether any individual methods already exists on Element. For example, Chrome, Firefox, Opera, and Safari implement the `element.remove()` method natively.

The detection code for mutation polyfill will let the polyfill happen if any of these methods are missing. Therefore, if the browser implements 5 of these methods but not the 6th, the polyfill would still kick in (as it should) but it will overwrite the 5 native implementations (as it shouldn't).

My proposed solution is to have per-method conditionals within the polyfill to only define the methods which are missing.

/cc @jonathantneal

jonathantneal:
@mislav good catch. The quick fix is to add a check before every method in both variations of the polyfill. However, if any browser implements another mutation method unlike our existing sets, we’ll have even more fragmented polyfills.

Each method (`append`, `prepend`, `before`, `after`, `replaceWith`, `remove`) should be its own dependency, and the mutation macro should be its own private dependency ala #280.

This would also make transitions like #254 all the easier.

mislav:
I think that the complexity of splitting mutation methods each into its own polyfill isn't worth it. All of them share the same internal method that they're implemented in, after all, and their total size is negligible when served over the wire. I would suggest keeping everything as it is now, but avoiding overriding native methods at all costs.

jonathantneal:
@mislav, I didn’t mean to miscommunicate; this is not about total size.

This is about keeping the polyfills organized and easily maintainable. What if Firefox begins supporting the ParentNode methods `prepend` and `append`? What if Chrome finishes supporting the remaining ChildNode methods `after`, `before`, and `replaceWith`? What if `after` is implemented in one browser but then becomes `appendAfter`? We could have 3 to 5 versions of the same polyfill.

All of those scenarios have happened in one way or another already, because we’re dealing with **prollyfills**. It’s an exciting new game to play, and this project has always benefited from keeping its polyfills modular.

triblondon:
Resolved by #288

PR #288, opened 2015-02-09
Separate Element mutations methods

jonathantneal:
Resolves #288 and #286
Requires resolution to #280


jonathantneal:
`Fatal error: Cannot read property 'browserName' of undefined`

Is that due to my commit, or due to Travis?

triblondon:
It's sauce being unreliable.  Just restart the build.  Usually works 2nd or
3rd time.

jonathantneal:
I’m on my fifth or sixth attempt of the day. Even after I removed an unused file in #287, that one has ceased passing as well. Could something else be wrong? It’s always the same error.

triblondon:
I removed the unreliable browser (iOS 7.1) from the browsers tested in CI, and also removed the deploy tool haikro from the package manifest, which was for some reason uninstallable on Travis.  The PR should now be testable.

jonathantneal:
Awesome blossom!

PR #289, opened 2015-02-10
Fix matchMedia queries with multiple dimensions

apaleslimghost:
Regex global flag was missing, so only the first dimension in a media query was being converted.

### DIFF
--- a/polyfills/matchMedia/polyfill.js
+++ b/polyfills/matchMedia/polyfill.js
@@ -13,7 +13,7 @@
 			.replace(/,/g, '||')
 			.replace(/and/g, '&&')
 			.replace(/dpi/g, '')
-			.replace(/(\d+)(cm|em|in|mm|pc|pt|px|rem)/, function ($0, $1, $2) {
+			.replace(/(\d+)(cm|em|in|mm|pc|pt|px|rem)/g, function ($0, $1, $2) {
 				return $1 * (
 					$2 === 'cm' ? 0.3937 * 96 : (
 						$2 === 'em' || $2 === 'rem' ? 16 : (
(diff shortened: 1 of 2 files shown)


apaleslimghost:
I can add tests if needed, but the current `matchMedia` tests are very barebones.

triblondon:
Adding a test for this specific case would be appreciated - it will help us gradually improve test coverage and this was obviously a gotcha during development.

apaleslimghost:
Added a basic test, just checks that the `new Function` doesn't throw a `SyntaxError`.

commuterjoy:
Out of interest, does this get merged then released? Or do we need to work around it locally for a bit.

triblondon:
It doesn't hit the CDN until we release 1.3, but as soon as it's merged you could check out and use master if you wanted the fix right away.

samgiles:
Maybe we could continuously deploy to latest.polyfill.io or something
similar?

On 11 February 2015 at 17:46, Andrew Betts notifications@github.com wrote:

> It doesn't hit the CDN until we release 1.3, but as soon as it's merged
> you could check out and use master if you wanted the fix right away.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/Financial-Times/polyfill-service/pull/289#issuecomment-73927521
> .

## 

_Sam Giles_
Developer
FT Labs

For Origami questions: origami.support@ft.com

Number One Southwark Bridge, London SE1 9HL

## 

---

This email was sent by a company owned by Pearson plc, registered office at 
80 Strand, London WC2R 0RL.  Registered in England and Wales with company 
number 53723.

matthew-andrews:
i think in this case we have a workaround

triblondon:
CI won't work here because you're using a fork.  FT employees do not need to use forks, so it's easier for me if you use a branch in the FT account.  I'll take your word for it on the test!

PR #290, opened 2015-02-10
Fix matchMedia queries with landscape

apaleslimghost:
Regex for `and` is matching `landscape` and converting it to `l&&scape`. This changes the regex to match at word boundaries.


matthew-andrews:
:trollface: test!

Issue #271, opened 2015-02-02
Custom polyfills


monolithed:
Based on https://github.com/Financial-Times/polyfill-service/issues/180 and https://github.com/Financial-Times/polyfill-service/issues/270 I propose to add a special option for custom polyfills

``` js
getPolyfillString({
    customPolyfills: [ 'paths/to/their/location/here' ]
})
```

Something like `customTests` in [`grunt-modernizr`](https://github.com/Modernizr/grunt-modernizr)

It may be useful for such methods like [URL](https://github.com/Financial-Times/polyfill-service/pull/273), `String.prototype.at`, `Array.prototype.includes` and other methods are not present in the repository

jonathantneal:
`Array.prototype.includes` was still using the withdrawn name `Array.prototype.contains`. I’ve filed a pull request #291. Not to interrupt your request, but thanks for the catch.

triblondon:
It would be hard to make this part of the library API, since by the time you call `getPolyfillString`, we have already compiled all the poyfill sources into a single sources.json file.  Adding custom polyfills would need to be factored into the grunt `buildsources` task.  Currently the polyfills location is defined here:

https://github.com/Financial-Times/polyfill-service/blob/master/tasks/buildsources.js#L20

So that could be made into an array of paths, and could be passed as a grunt parameter.  Happy to accept a PR along those lines.

triblondon:
No intent to implement this.

PR #291, opened 2015-02-11
Array.prototype.contains => Array.prototype.includes

jonathantneal:
Like `String.prototype.contains`, `Array.prototype.contains` was found to be [not web-compatible](https://esdiscuss.org/topic/having-a-non-enumerable-array-prototype-contains-may-not-be-web-compatible).

Also touches on a need in #271.


triblondon:
As with `String.prototype.contains`, I suggest we continue to support contains but make it depend on includes and have a polyfill that is simply an alias to includes.   Thankfully contains was never in the default set, though it was in the `es7` and `modernizr:es7array` sets, so anyone using those features would no longer get contains.

triblondon:
I'll re-add contains and fix #294 at the same time

PR #293, opened 2015-02-11
Use graphite for metrics (+better env var docs to support that)

triblondon:
FT is changing the way we collect and monitor service metrics to standardise on graphite.  This adds metrics to replace existing /__metrics HTTP endpoint.


service/index.js
```diff
@@ -171,8 +160,6 @@ app.get(/^\/v1\/polyfill(\.\w+)(\.\w+)?/, function(req, res) {
 	res.set('Content-Type', contentTypes[fileExtension]+';charset=utf-8');
 	res.set('Access-Control-Allow-Origin', '*');
 	res.send(op);
-	metrics.addResponseTime(Date.now() - responseStartTime);
```
samgiles, line 175: I reckon we should still monitor this?

service/index.js
```diff
@@ -17,9 +17,13 @@ var polyfillio = require('../lib'),
 
 var argv = parseArgs(process.argv.slice(2));
 
-var port = argv.port || Number(process.env.PORT) || 3000,
-	metrics = new ServiceMetrics(),
-	contentTypes = {".js": 'application/javascript', ".css": 'text/css'};
+var port = argv.port || Number(process.env.PORT) || 3000;
+var metrics = new StatsD({
+	host: process.env.STATSD_HOST,
```
samgiles: What's the behaviour if this env var isn't set?

samgiles:
We should look at http://github.com/Financial-Times/next-metrics.  StatsD isn't really the one true way.

service/index.js
triblondon, line 175: I think this should be end to end.  Currently there's nothing in between the start and stop points that has any chance of being slow - there's no IO of any kind, other than the final HTTP response.  The current data shows an average time of 2ms, while the total RTT will be more like 20 at a minimum.

service/index.js
triblondon: It just acts as a mock and doesn't do anything.

service/index.js
samgiles, line 175: Where's the current data?  If it's based on the stuff in Heroku it's more likely to be how fast node responds to the `__gtg`, which v8 will have JIT'd for us.  I still think measuring this here makes sense.

triblondon, line 175: With node-measured this turns out to be really easy so I added it back.

triblondon:
@samgiles this is worth another look.  Adding node-measured made for a very elegant solution, though I worry a bit that this isn't necessarily easy for other tech stacks.  Technologies that don't maintain a running process, like PHP, will need more complex solutions to collect metrics if aggregation is to be the responsibility of the service.  That said, if everyone prefers the raw carbon protocol, it is at least pretty easy to use from Node, and we could make a pattern from this very easily especially if my suggested flatten function were to be pushed into the node-measured module.

service/metrics.js
```diff
@@ -1,109 +1,36 @@
...
+var Measured = require('measured');
+
+var data = Measured.createCollection('polyfill.' + (process.env.ENV_NAME || "unknown"));
+var reportInterval = 5000;
+var timer = null;
+var graphite = null;
+
+if (process.env.GRAPHITE_URL) {
+	graphite = Graphite.createClient(process.env.GRAPHITE_URL);
+	timer = setInterval(function() {
```
samgiles, line 15: I'd suggest using `timer.unref()` 

From nodejs.org docs:

> The opaque value returned by setTimeout and setInterval also has the method timer.unref() which will allow you to create a timer that is active but if it is the only item left in the event loop won't keep the program running

service/metrics.js
```diff
@@ -1,109 +1,36 @@
...
-		"val": this.cssResponsesCount,
-		"unit": "items",
-		"since": new Date(this.serviceStartTime).toISOString(),
-		"lastUpdated": new Date(this.cssResponsesLastUpdated).toISOString()
-	};
-};
+// This could be added to node-measured, perhaps as a static 
+// Measured.flattenMetrics(instance.getJSON()), or by adding a 
+// boolean arg to getJSON(). Consider making a PR.
+function flatten(obj, prefix, root) {
```
samgiles: This looks like the [same algorithm](https://github.com/felixge/node-graphite/blob/master/lib/GraphiteClient.js#L15) which is called by [`graphite.write`](https://github.com/felixge/node-graphite/blob/master/lib/GraphiteClient.js#L15), I don't think this is required.

README.md
```diff
@@ -29,12 +27,22 @@ Alternatively, deploy straight to Heroku:
...
 For an HTTP API reference, see the [hosted service documentation](http://polyfill.webservices.ft.com).
 
+### Environment configuration
+
+The service reads a number of environment variables when started as a service:
+
+* `PORT`: The port on which to listen for HTTP requests (default 3000)
+* `FASTLY_SERVICE_ID`, `FASTLY_API_KEY`: Used to fetch and render cache hit stats on the [Usage](https://cdn.polyfill.io/v1/docs/usage) page of the hosted documentation.  If not specified, no stats will be shown.
+* `PINGDOM_CHECK_ID`, `PINGDOM_API_KEY`, `PINGDOM_ACCOUNT`, `PINGDOM_USERNAME`, `PINGDOM_PASSWORD`: Used to fetch and render uptime and response time stats on the [Usage](https://cdn.polyfill.io/v1/docs/usage) page of the hosted documentation.  If not specified, no stats will be shown.
+* `GRAPHITE_URL`: URL to which to send Carbon metrics, including protocol, host and port, eg `plaintext://graphite.example.org:2003/`.  If not set, no metrics will be sent.
```
samgiles: When we require TLS we'll have to update this:

``` JS
var tls = require('tls');
var GraphiteClient = require('graphite');
var CarbonClient = require('graphite/lib/CarbonClient');

var socket = tls.connect(2003, {
    rejectUnauthorized: false
}, function() {
    console.log('client connected: ', socket.authorized ? 'authorized' : 'unauthorized');
});

var graphiteClient = new GraphiteClient({ 
    carbon: new CarbonClient({
        socket: socket
    }) 
});
```

PR #298, opened 2015-02-11
Url polyfill

triblondon:
Superseding https://github.com/Financial-Times/polyfill-service/pull/273

### DIFF
--- a/polyfills/URL/detect.js
+++ b/polyfills/URL/detect.js
@@ -1,0 +1,21 @@
+(function (global) {
+	/*
+	 * Browsers may have:
+	 * No global URL object
+	 * URL with static methods only - may have a dummy constructor
+	 * URL with members except searchParams
+	 * Full URL API support
+	 */
+	'use strict';
+
+	try {
+		if (global.URL) {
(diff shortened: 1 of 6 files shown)


polyfills/URL/detect.js
```diff
@@ -0,0 +1,21 @@
...
+		if (global.URL) {
+			var nativeURL = new global.URL('http://example.com');
+
+			if ('href' in nativeURL && 'searchParams' in nativeURL) {
+				return nativeURL;
+			}
+		}
+	}
+	catch (error) { }
+}(this))
```
triblondon, line 21: This could be a bit more concise:

``` javascript
(function() {
try {
  var test;
  return URL && test = new URL('http://example.com')) && 'href' in test && 'searchParams' in test;
} catch(e) {};
})()
```

triblondon:
CI passed, excellent job @monolithed.  Can you consider the shorter detect?

monolithed:
@triblondon, thanks for the one!

I have no any objections, but I don't understand why do the minifier's job? :)

PR #300, opened 2015-02-12
Fixes #299: typo on line Event/polyfill-ie8.js:99

### LINKED ISSUE #299: IE8 Event polyfill typo
https://github.com/Financial-Times/polyfill-service/blob/de07f9237ec4a73624e8f2479a1b308a7c8d56b0/polyfills/Event/polyfill-ie8.js#L99

The line:

``` .js
listener = arguments[1].
```

Should be:

``` .js
listener = arguments[1],
```

tejacques:
Pretty straightforward change. There was a typo causing listener to be set to the `arguments[1].index` instead of `arugments[1]` and then creating an index variable.

### DIFF
--- a/polyfills/Event/polyfill-ie8.js
+++ b/polyfills/Event/polyfill-ie8.js
@@ -96,7 +96,7 @@
 		var
 		element = this,
 		type = arguments[0],
-		listener = arguments[1].
+		listener = arguments[1],
 		index;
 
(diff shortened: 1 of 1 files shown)


triblondon:
Fixed in 61fc98e7bf0df940833c395ebd59cb58bb7c5b32 (I wanted to have a test for this)

PR #301, opened 2015-02-16
Add IndexedDB polyfill, closes #52

### LINKED ISSUE #52: IndexedDB
Could I request that an IndexedDB polyfill be added (that uses WebSQL as an underlying storage tech if IDB is not available)?  [Perhaps - as it is quite big - one that is only added when it is explicitly requested, ie. http://polyfill.io/maybe/indexeddb ]

(The use case of the optional-ness of it would be just that such a polyfill is likely to be huge and [for most people] unwanted)

http://nparashuram.com/IndexedDBShim/

Also the [IndexedDB wrapper](https://github.com/jensarps/IDBWrapper) could be handy for 'smoothing over the differences' in IDB implementations across browsers.

triblondon:
DON'T MERGE YET

Starting point for IndexedDB.  Using @axemclion's polyfill.  To be resolved:
- Are the browser ranges correct?  Currently I just copied from caniuse, excluding only browsers showing as green, but we should also be excluding browsers that don't have the required prerequisites (websql?)
- Dependencies.  I have not reviewed the polyfill extensively for implicit dependencies or embedded polyfills
- Tests.  Tests in the source repo use qunit, we use mocha. If we don't edit the code, we're probably OK to rely on the source repo CI.

@axemclion are you interested in helping us integrate your polyfill into the polyfill service?

### DIFF
--- a/polyfills/indexedDB/config.json
+++ b/polyfills/indexedDB/config.json
@@ -1,0 +1,11 @@
+{
+	"browsers": {
+		"ie": "*",
+		"firefox": "<10",
+		"safari": "*",
+		"opera": "<15",
+		"ios_saf": "*"
+	},
+	"author": "Parashuram N <github@nparashuram.com>",
+	"licence": "MIT"
+}
--- a/polyfills/indexedDB/detect.js
+++ b/polyfills/indexedDB/detect.js
@@ -1,0 +1,1 @@
+'indexedDB' in window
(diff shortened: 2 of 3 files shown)


triblondon:
@matthew-andrews are we still interested in this at FT?  If so can Next team spare any time on the above?

axemclion:
@triblondon - yes i am :)

triblondon:
Hi @axemclion sorry this has languished. There is no longer internal demand at the FT for IndexedDB to be in the polyfill service so we're not putting time into it but if you want to, you'd be very welcome to push it forward.  Contribution guidelines are at cdn.polyfill.io/v2/docs/contributing.

PR #306, opened 2015-02-24
delete element._events[type] when last listener is removed. Fixes #302

### LINKED ISSUE #302: Event ie8-polyfill.js removeEventListener breaks addEventListener
### Bug

Calling removeEventListener on the last listener on an element detaches the event handler and addEventListener no longer re-adds it because the `_events[type]` object still exists.

The problem is the event is detached when the list is empty, but `element._events[type]` is not removed.
### Fix:

Change: https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Event/polyfill-ie8.js#L108-L112

From:

``` .js
                if (!element._events[type].list.length) {
                    if (element.detachEvent) {
                        element.detachEvent('on' + type, element._events[type]);
                    }
                }
```

To:

``` .js
                if (!element._events[type].list.length) {
                    if (element.detachEvent) {
                        element.detachEvent('on' + type, element._events[type]);
                    }
                    delete element._events[type];
                }
```
### Steps to reproduce

**Note**: You must use a real DOM event and not a custom one, because otherwise dispatchEvent catches the fireEvent error, and manually runs the event listeners.

``` .js
var listener = function() { console.log('should log'); };
document.addEventListener('click', listener);
document.removeEventListener('click', listener);
document.addEventListener('click', listener);
// click the document
document.dispatchEvent(new Event('click'));
```
#### Expected output

```
should log
```
#### Actual output

```
```
#### Mocha Test

``` .js
it('should fire event', function (done) {
    var listener = function() {
        // remove listener
        document.removeEventListener('click', listener);
        done();
    };
    document.addEventListener('click', listener);
    document.removeEventListener('click', listener);
    document.addEventListener('click', listener);
    // click the document
    document.dispatchEvent(new Event('click'));
});
```

tejacques:
Currently when Calling removeEventListener on the last listener on an element detaches the event handler and addEventListener no longer re-adds it because the `_events[type]` object still exists.

The problem is the event is detached when the list is empty, but `element._events[type]` is not removed.

This change fixes the problem by deleting `element._events[type]` when the last listener is removed in `removeEventListener`, and also adds a test to catch this case.

### DIFF
--- a/polyfills/Event/polyfill-ie8.js
+++ b/polyfills/Event/polyfill-ie8.js
@@ -109,6 +109,7 @@
 					if (element.detachEvent) {
 						element.detachEvent('on' + type, element._events[type]);
 					}
+					delete element._events[type];
 				}
 			}
 		}
--- a/polyfills/Event/tests.js
+++ b/polyfills/Event/tests.js
@@ -74,3 +74,22 @@ it('should should not trigger an event handler once removed', function() {
 	testEl.removeEventListener('test', listener);
 	testEl.dispatchEvent(testEvent);
 });
+
+it('should trigger an event handler once added, removed, and added again', function () {
+	// NOTE: The event must be a real DOM event or the
+	// dispatchEvent polyfill will catch the fireEvent
+	// error, simulate firing the event by running the
+	// event listeners.
+	var fired = false;
+	var listener = function() {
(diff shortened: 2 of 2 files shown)


triblondon:
Many thanks for this.

PR #315, opened 2015-02-25
Fix Promise.all, fixes #307

### LINKED ISSUE #307: Bug in Promise polyfill
The array check in the `Promise.all` method (line 296) will throw a TypeError. I think it should use `Object.prototype.toString`.

triblondon:


jonathantneal:
Thanks, @triblondon!

PR #318, opened 2015-02-27
Fix installing collections

juandopazo:
When I tried building I ran into a couple of issues with the `installcollections` task:
- It was using Linux commands (and I'm on Windows :ghost:)
- It was using a symlink (which doesn't work in Windows without admin privileges) and symlinks aren't packaged by npm, which could lead to issues for people using the package as a library
- There was a regexp which wasn't working

So I replaced the `rm -rf` with grunt-contrib-clean, removed the symlink and fixed the tag list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /node_modules
 /npm-debug.log
+/polyfills/__repo
 /polyfills/__versions
 /polyfills/aliases.json
 /polyfills/sources.json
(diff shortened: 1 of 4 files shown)


triblondon:
So on deploy we'll now have two copies of the latest library version, assuming there's a version tag pointing at the commit that's being deployed.  I guess that's OK.

juandopazo:
@triblondon you could remove the source `polyfills/` folder from the package by having npm ignore it, but I guess you use git directly for deploying...

PR #324, opened 2015-03-02
Window.location.origin

Casao:
Adds a window.location.origin polyfill, as this is missing from many older versions but available in all of the current.

### DIFF
--- a/polyfills/Window.location.origin/detect.js
+++ b/polyfills/Window.location.origin/detect.js
@@ -1,10 +1,1 @@
-(function(global) {
-	'use strict';
-
-	try {
-		if (global.window) {
-			return window.location.origin
-		}
-	}
-	catch (error) { }
-}(this))
+'origin' in window.location
--- a/polyfills/Window.location.origin/detect.js
+++ b/polyfills/Window.location.origin/detect.js
@@ -1,0 +1,10 @@
+(function(global) {
+	'use strict';
+
+	try {
+		if (global.window) {
+			return window.location.origin
+		}
+	}
+	catch (error) { }
+}(this))
--- a/polyfills/Window.location.origin/polyfill.js
+++ b/polyfills/Window.location.origin/polyfill.js
@@ -1,0 +1,7 @@
+var origin = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '');
+Object.defineProperty(window.location, 'origin', {
+	enumerable: true,
+	writable: false,
+	value: origin,
+	configurable: false
+} );
--- a/polyfills/Window.location.origin/polyfill.js
+++ b/polyfills/Window.location.origin/polyfill.js
@@ -1,7 +1,6 @@
-var origin = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '');
 Object.defineProperty(window.location, 'origin', {
 	enumerable: true,
 	writable: false,
-	value: origin,
+	value: window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : ''),
 	configurable: false
 } );
--- a/polyfills/Window.location.origin/config.json
+++ b/polyfills/Window.location.origin/config.json
@@ -1,0 +1,17 @@
+{
+    "browsers": {
+        "ie": "6 - 9",
+        "firefox": "<=20",
+        "safari": "<=4",
+        "ios_saf": "<=6"
+    },
+    "dependencies": [
+        "Object.defineProperties"
+    ],
+    "author": {
+        "name" : "Colin Ewen",
+        "email" : "colin@colinewen.com",
+        "url" : "http://github.com/casao"
+    },
+    "license": "MIT"
+}
--- a/polyfills/Window.location.origin/config.json
+++ b/polyfills/Window.location.origin/config.json
@@ -1,9 +1,10 @@
 {
     "browsers": {
-        "ie": "6 - 9",
+        "ie": "6 - 10",
         "firefox": "<=20",
-        "safari": "<=4",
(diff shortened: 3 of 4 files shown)


jonathantneal:
Thanks for pointing out that this was missing.

So, this is probably a deeper discussion for FT, but why are there author fields? Are we encouraging users to license polyfills? I’m not even sure how licenses work when a polyfill fulfills an existing specification.

On the one hand, @Casao is being super helpful and is following the contributors guide as I read it. On the other hand, I have never encountered a scenario where someone MIT’d a well known one liner.

@Casao, @triblondon, thoughts?

Casao:
I agree @jonathantneal. Licensing is generally an exemption from copyright/patent litigation, which would not apply at all here.

I'd be happy to remove these if that's the consensus -- as you said, I just went with the contributing guidelines as much as possible.

jonathantneal:
Yes, you followed the documentation I see; only trading one relatively unused probably-not-well-understood license for a more-common-and-better-understood license. Thanks for being so reasonable about this!

IMO, the spirit of licensing a polyfill in any fashion is to credit original contributions that are not apparent or explicit in the specification. There are a handful of instances — enqueuing a microtask, parsing a CSS file, transforming data into base64  — where I see the steps being taken to accomplish the task are neither intuitive nor explicitly stated.

Perhaps that could be put into writing somewhere more official? I’m also sorry to anyone out there I could be offending, as licenses can be touchy subject.

polyfills/Window.location.origin/polyfill.js
```diff
@@ -0,0 +1,7 @@
+var origin = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '');
```
triblondon: This value will leak into all other polyfills.  We don't put a closure around each polyfill since many have their own anyway.  In this case you don't really need one, you could just move the expression inline on line 5.

polyfills/Window.location.origin/detect.js
```diff
@@ -0,0 +1,10 @@
+(function(global) {
+	'use strict';
+
+	try {
+		if (global.window) {
+			return window.location.origin
+		}
+	}
+	catch (error) { }
+}(this))
```
triblondon: I think a way of expressing this that's more consistent with our existing detects would be:

``` javascript
'origin' in window.location
```

If this isn't good enough, wouldn't lots of our other detects be flawed for the same reason?  We should potentially have something in the contributors' guide about a standard way of testing for the presence of an object property.

triblondon:
Thanks @Casao, this is much appreciated.  I've made a few comments inline.  To summarise those and respond to the larger point above:
- We don't create an individual protected scope for each polyfill, so you need to either create one, or not define any variables at the top level of your polyfill.  You could easily avoid creating the `origin` variable, so this should be an easy change
- The detect could be simpler
- Could you add a test?  It might seem contrived but it helps us map out support automatically using CI
- Regarding licences and attribution, I am talking on and off with the FT's legal team about this.  In order for the FT to dedicate the code in this repo to the public domain (CC0 licence) we need to be as sure as we can that the author (where the code is supplied by a third party) has already done so. It becomes more complex when the person who adds the code to the polyfill service is not the original author.  The problem with CC0 in particular is that since it is the 'free-est' possible licence, we cannot re-release code that is encumbered by any other licence at all (even MIT) as CC0.  Therefore if this code was your own invention and you intended to licence it MIT, I'd suggest you retain the author and licence properties to document that.  If not, I'll take agreement acknowledgement of this comment as your agreement to releasing the code CC0.

Once I've got to the bottom of this issue there'll probably be some sort of contribution terms clause added to the contributors' guide, basically saying it's CC0 if you don't add licensing info.  Something as non-scary and efficient as possible!

jonathantneal:
On closures for each polyfill: what are the guidelines for polyfills now that we maintain a global iife? 

On licensing: Had it not been for the base64 polyfill, I would have drawn a hard-line on not accepting anything non-public-domain. How many polyfills have licenses now (maybe two or three at most; base64, Promise, and the html5shiv)? Licenses should be the exception, not the rule. Most projects do not accept contributions that do not already conform to the project itself. Hopefully all contributors are celebrated, and eternally credited in the history and contributions section of this repository.

triblondon:
There's a closure around the whole bundle, not around each polyfill.  Currently, @Casao's 'origin' variable will exist across all the polyfills but will not leak into global scope.

On licensing, I think we need to chat in real time about this.  I understand the principle, but we can make faster progress with this project if we're willing to use third party code more often.  Your polyfills are excellent, but isn't it better when there's a whole team of people working on them, like there are looking at es6-shim and es5-shim?  Both of those are MIT licensed.  I am happy to lead the project but my JavaScript just isn't good enough to resolve all the issues with individual polyfills, and I see it as a waste of time to recreate (or debug) something that already exists and has been thoroughly tested in earnest by thousands of people.  That's why whenever there's a problem with a polyfill we have, my instinct is to say "is there an es*-shim version of this?  does it have the problem?"  If not, well, it's a lot quicker (for me at least) to just replace our polyfill with the community one than try to solve the issue with ours.

Casao:
@triblondon - I've taken your feedback into place and put something up! I'll work on testing, but I've always found testing window.location to be a bit tricky to do well. I'll see what I can do.

As far as license/copyright/whatever goes, it's such a simple snippet of code that's been arrived at independently a number of times, I'm not sure there is truly an "original author" - I'm definitely open to it being CC0.

triblondon:
Thanks Colin.  Yes, point very much taken on the simplicity, I guess it just prompted the discussion.

Casao:
That's probably the best I can do for tests without a mocking library, but if there's any more feedback I'm happy to hear it :)

triblondon:
Looks good to me.  Thanks!

Issue #313, opened 2015-02-25
Function.prototype.bind on iOS 5 (old iPads & iPhones)


Savageman:
This is a missing polyfill.

I'm not too sure which "ios_safari" version to check against in the **config.json** file...

Another important information is that the [MDN polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Polyfill) doesn't work (triggers some `instanceof` error), I had to use the one from **es5-shim.js**.

triblondon:
I don't have an iOS 5 device to test on.  Does the polyfill we have in the service already work on Safari on iOS 5?

Savageman:
I just tested it and no, it doesn't work. I have the same error as the MDN polyfill: "TypeError : instanceof called on an object with an invalid prototype property".

triblondon:
OK, if the es5-shim polyfill works on our test browsers (IE 6,7,8 and FF3.6 for this one) then I'm happy to switch to that.  Can you do a PR?

Savageman:
You closed this, but the original issue is still not resolved.

The included polyfill should now work on iOS 5, but it's still not included (I can't see the `ios_safari` version in the `config.json` file).

triblondon:
Sorry, I didn't check that the targeting was correct.  I'll push a commit to target iOS < 6.

Issue #261, opened 2015-01-28
Function.prototype.bind error on Safari 5.1


blainekasten:
When loading the polyfill in Safari 5.1 on Windows 7.

The following is spit into the console:

`TypeError: instanceof called on an object with an invalid prototype property.`

It occurs in this function:

```
Function.prototype.bind = function bind(scope) {
    var callback = this,
    prepend = Array.prototype.slice.call(arguments, 1),
    Constructor = function () {},
    bound = function () {
        return callback.apply(
            this instanceof Constructor && scope ? this : scope,
            prepend.concat(Array.prototype.slice.call(arguments, 0))
        );
    };

    Constructor.prototype = bound.prototype = callback.prototype;

    return bound;
};
```

jkimbo:
:+1: I've just come across this problem in OSX Lion with the same version of Safari

triblondon:
Thanks for the report.  We'd be grateful if you could debug it and contribute a fix if you have time.

triblondon:
Note that issue https://github.com/Financial-Times/polyfill-service/issues/313 also has a problem with the `bind` polyfill.  If you can confirm that es5-shim version fixes your problem, we could switch to that.

jkimbo:
@triblondon I've just added in the es5-shim version and it seems to work on Safari.

jonathantneal:
Yes, my polyfill was more like the MDN one. John Dalton even tried to warn me about this long long ago. Unfortunately, I did not yet fully understand prototype inheritance at that time. So, I’ve been reviewing those notes and these and refactoring the polyfill early this morning. I needed to break for my daily work, but I expect to complete the update tonight.

Here’s a snippet of where I left off, and my new thinking for the polyfill:

``` js
function Empty() {}

var splice = Array.prototype.splice;

Function.prototype.bind = Function.prototype.bind || function bind(scope) {
    var
    callback = this,
    args = splice.call(arguments, 1),
    index = -1,
    length = 0,
    buffer = 'return function (',
    binder = function () {
        var
        self = this,
        isNew = self instanceof bound,
        result = callback.apply(isNew ? self : scope, args.concat(splice.call(arguments, 0)));

        return !isNew || Object(result) !== result ? result : self;
    };

    if (!(callback instanceof Function)) {
        throw new TypeError(callback + ' is not a function');
    }

    length = Math.max(length, callback.length - args.length);

    while (++index < length) {
        buffer += '$' + index;
    }

[...]

PR #329, opened 2015-03-03
Update Function.prototype.bind

### LINKED ISSUE #89: Function.prototype.bind
Actually this is not a polyfill. When the returned function by bind is called as: `new (String.bind('test'));` then it's called the [[Construct]] method of the bound function. Since built-in constructors has overloaded behavior depends on its calling, your polyfill is not reliable. If you don't sure, try this: `new (String.bind('test')) === new (String.bind('test'))`

jonathantneal:
Throw on non-functions
Calculate bound function arity
Bind prototype

Resolves #89, #261, #313

### DIFF
--- a/polyfills/Function.prototype.bind/polyfill.js
+++ b/polyfills/Function.prototype.bind/polyfill.js
@@ -1,16 +1,42 @@
-Function.prototype.bind = function bind(scope) {
-	var
-	callback = this,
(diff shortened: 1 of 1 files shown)


jonathantneal:
@triblondon, all is well with Travis. Would you review and merge?

PR #333, opened 2015-03-06
devicePixelRatio broken in IE8

triblondon:
Our devicePixelRatio polyfill uses `Object.defineProperty` on IE8, but the IE8 defineProperty polyfill does not support accessors, so it throws "getters & setters can not be defined on this javascript engine".  Well actually it throws "Exception thrown", because IE8.

The IE7 polyfill is pretty impressive but I propose a much simpler solution for both.  The chances of a device that runs IE < 9 having a DPR of more than 1 are so low that I suggest quite simply hard coding the value to 1, which solves this problem


triblondon:
Need to merge to get CI to pass

PR #335, opened 2015-03-06
document.head dependency in geolocation

alexander-akait:
Ie8 and lte comments and fix create scripts for ie 8
http://stackoverflow.com/questions/17100246/document-head-appendchildelement-ie-ie7-and-ie8


jonathantneal:
Isn't this already handled by the [document.head polyfill](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/document.head/polyfill.js)?

triblondon:
Yes, a better way to do this would be to express a dependency in the geolocation config.json on the `document.head` polyfill (which we do have)

PR #336, opened 2015-03-06
Use es5-shim version of Function.prototype.bind

triblondon:
Currently we are failing a test for `Function.prototype.bind` in IE8 and below:

image

Unfortunately there was a bug in CI that suppressed this error from the testing for PR #329.  I don't know how to solve it, but I've verified that the version of bind in es5-shim does not have the problem.  This PR switches to use that polyfill.

Annoyingly, the ES5-shims make use of a fair number of shorthand variables that make it hard to extract just the polyfill that you want, so this is not a clean extract.  Suggestions welcome, and of course @jonathantneal if you prefer to fix your earlier implementation feel free to do that and propose an alternative PR.

### DIFF
--- a/polyfills/Function.prototype.bind/polyfill.js
+++ b/polyfills/Function.prototype.bind/polyfill.js
@@ -1,42 +1,157 @@
-(function (FUNCTION, splice, Empty) {
-	FUNCTION.bind = FUNCTION.bind || function bind(scope) {
-		var
-		callback = this,
-		args = splice.call(arguments, 1),
-		arity = [],
-		index = 0,
-		bound, length;
+Object.defineProperty(Function.prototype, 'bind', {
+    value: function bind(that) { // .length is 1
+        // 1. Let Target be the this value.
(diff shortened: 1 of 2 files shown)


triblondon:
I need to get the CI passing, so am going to merge this.  Feel free to come back with a fix to the original.

polyfills/Function.prototype.bind/polyfill.js
```diff
@@ -1,42 +1,157 @@
...
+                //   property.
+                // 4. Let args be a new list containing the same values as the
+                //   list boundArgs in the same order followed by the same
+                //   values as the list ExtraArgs in the same order.
+                // 5. Return the result of calling the [[Construct]] internal
+                //   method of target providing args as the arguments.
+
+                var result = target.apply(
+                    this,
+                    args.concat(array_slice.call(arguments))
```
vvo, line 68: array_slice is not defined anywhere thus this polyfill fails

PR #342, opened 2015-03-10
Load webfonts

kaelig:


triblondon:
Is this all there is to it? Don't you also have to assign the font family to something the header will inherit from?

kaelig:
Good catch, my bad! I added the font-family to the html element and that should do it.

PR #349, opened 2015-03-14
fix polyfill String.prototype.endsWith

christophehurpeau:
endsWith doesnt return the right result when the length of string in arguments is equals to the other string's length -1

For exemple, if you have this = 'a' and string = 'ab', 'a'.endsWith('ab') will return true.


triblondon:
Added some tests in 4f0ab0c42c8185c4845d895448955ee7fc86c488 that include a test for the case discussed here. Current lib fails, proposed change passes, so happy to merge.

Thanks for the contribution.

PR #351, opened 2015-03-16
_enqueueMicrotask should target all browser versions

samgiles:
If there is a dependency on the _enqueueMicrotask it should always
be included regardless of the browser version.

This is because it is a non-standard browser API specific to some polyfills
where efficient async task scheduling is required.

### DIFF
--- a/polyfills/_enqueueMicrotask/config.json
+++ b/polyfills/_enqueueMicrotask/config.json
@@ -4,13 +4,13 @@
 		"caniuse:promises"
 	],
 	"browsers": {
-		"android": "* - 4.3",
-		"chrome": "* - 17",
-		"firefox": "* - 13",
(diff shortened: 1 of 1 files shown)


triblondon:
The existing config does cover all browsers, doesn't it, if you take the Union of both variants?

samgiles:
Not really, because MutationObserver covers single browser versions and the default has an upper bound.  There's quite a lot missing (Safari 7 for example), this covers everything that isn't in a variant.  (I've tested the variants still work).

triblondon:
I could have sworn I fixed this, sorry.  You need to tweak your edit a bit, because you now have overlapping ranges between the default and the mutationobserver variant.  I suggest reverting the proposed change and instead extending the mutationobserver ranges to `- *`.

PR #354, opened 2015-03-19
Add (most) missing docs links

callumlocke:
couple of difficult ones:
- `viewport` – can't think of any particular docs page for this. the polyfill just ensures a mixed bag of `window.*` dimension properties are available (`innerHeight innerWidth pageXOffset pageYOffset scrollX scrollY`).
- `~html5-elements` – again not obvious how to document this. maybe the link should be to [html5 shiv on wikipedia](http://en.wikipedia.org/wiki/HTML5_Shiv)? I left it out for now

### DIFF
--- a/polyfills/Document/config.json
+++ b/polyfills/Document/config.json
@@ -12,5 +12,6 @@
 			}
 		}
 	},
-	"spec": "http://dom.spec.whatwg.org/#interface-document"
+	"spec": "http://dom.spec.whatwg.org/#interface-document",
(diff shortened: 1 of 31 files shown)


triblondon:
Thanks so much for doing this, I was expecting to end up doing it myself.

callumlocke:
no problem it had been bugging me :)

I think it would be nicer if the feature name (in the leftmost column of the table) was itself a link to the docs, rather than having Documentation hidden away as an item in the hamburger menu. Do you agree? If so I'll PR that too.

triblondon:
OK, go for it.

Issue #358, opened 2015-03-21
polyfill addEventListener for old ie


alexander-akait:
https://gist.github.com/jonathantneal/3748027

callumlocke:
This already exists, it's part of the [Event](https://cdn.polyfill.io/v1/docs/features/#Event) polyfill.

This could be made clearer in the table by adding a note under Event.

triblondon:
Thanks, merged your suggestion

PR #362, opened 2015-03-27
Add note about EventTarget methods

callumlocke:
(addresses #358)

MDN's [Event](https://developer.mozilla.org/en/docs/Web/API/Event) page makes no mention of `addEventListener`, `removeEventListener` and `dispatchEvent`. These methods are part of a separate interface called [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget).

This PR adds a note to explain that with Event you also get EventTarget methods.

(in fact, as these are separate things in the spec [[Event](https://dom.spec.whatwg.org/#interface-event), [EventTarget](https://dom.spec.whatwg.org/#interface-eventtarget)], maybe we should have them as separate polyfills?)

### DIFF
--- a/polyfills/Event/config.json
+++ b/polyfills/Event/config.json
@@ -23,5 +23,8 @@
 			]
 		}
 	},
-	"docs": "https://developer.mozilla.org/en/docs/Web/API/Event"
+	"docs": "https://developer.mozilla.org/en/docs/Web/API/Event",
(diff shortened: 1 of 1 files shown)


triblondon:
I'm not opposed to these being separate polyfills, but this is probably sufficient.

PR #368, opened 2015-04-02
Update browser spec for Array.prototype.findIndex

andyburke:
Applies the same change as: https://github.com/andyburke/polyfill-service/commit/9f5ea576033a4bf295b41a1027d48f833046d933

### DIFF
--- a/polyfills/Array.prototype.findIndex/config.json
+++ b/polyfills/Array.prototype.findIndex/config.json
@@ -4,11 +4,11 @@
 		"modernizr:es6array"
 	],
 	"browsers": {
-		"chrome": "* - 29",
+		"chrome": "*",
 		"firefox": "3.6 - 24",
 		"ie": "*",
-		"opera": "* - 16",
-		"safari": "* - 7"
+		"opera": "*",
+		"safari": "* - 7.0"
 	},
 	"dependencies": [
 		"Object.defineProperty"


triblondon:
Thanks.  Have you proved this through testing or is this data from another source?

andyburke:
It's the same change that was applied to Array.prototype.find. (the link is to my branch but it's actually your change)

I have not tested it. I just assumed it was an oversight that Array.prototype.find was changed like this and findIndex wasn't.

I arrived here because I am not getting Array.prototype.findIndex w/ chrome 41 via cdn.polyfill.io with the modernizr:es6array feature. I also tried explicitly doing the Array.prototype.findIndex feature and was still not getting it. It was only once I added |always that it worked.

I just thought this might be the fix and figured it would be better to send a PR than an issue.

If this fix is ridiculous, my apologies, I haven't dug in much on how the service itself works.

triblondon:
Ah, OK, makes sense, thanks.

andyburke:
Any idea when this will get deployed? I am still not getting findIndex back when I hit https://cdn.polyfill.io/v1/polyfill.min.js?features=modernizr:es6array

triblondon:
RC of 1.4 will be released today, barring complaints about interop with 1.3, it will go stable in a week.

PR #376, opened 2015-04-07
Remove license metadata for fetch polyfill

### LINKED ISSUE #374: "fetch" polyfill licence info is wrong
site has a warning about `fetch` polyfill having a proprietary licence:

image

..but it doesn't, it just uses the MIT licence:

https://github.com/github/fetch/blob/master/LICENSE

callumlocke:
fixes #374

### DIFF
--- a/polyfills/fetch/config.json
+++ b/polyfills/fetch/config.json
@@ -13,7 +13,7 @@
 		"Promise",
 		"XMLHttpRequest"
 	],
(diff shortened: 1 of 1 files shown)


triblondon:
Change the value to "MIT" so it gets refected correctly in the docs.  Otherwise we assume it's CC0

callumlocke:
done and rebased

triblondon:
Thanks!

PR #381, opened 2015-04-14
Adds test to Object.create for API objects as it fails in IE8

AlbertoElias:
So API objects in IE8 aren't instances of `Object`, so Object.create doesn't work on them. I've created the test that fails for this use case. I'm not really sure what the ideal solution to this problem is though


triblondon:
I'm not sure what you mean by "API object".  Do you mean "DOM object"?!

If so, I think I fixed it, and edited your test to clarify that language.

AlbertoElias:
I meant Native Objects

AlbertoElias:
@triblondon how's that?

triblondon:
I'm not convinced that the polyfill for CustomElements is needed to solve this problem :-)

AlbertoElias:
Oops! Let me remove those files!

AlbertoElias:
Done

PR #382, opened 2015-04-16
Added failing test of matchMedia using dppx

jkimbo:
When using the unit dppx in a media query the `matchMedia` polyfill throws an error. Note this happens on all browsers if you force the polyfill. I've added a failing test for it though I have yet to look into fixing it incase anyone else has a better understanding as to why it is failing.

### DIFF
--- a/polyfills/matchMedia/polyfill.js
+++ b/polyfills/matchMedia/polyfill.js
@@ -13,11 +13,11 @@
 			.replace(/,/g, '||')
 			.replace(/\band\b/g, '&&')
 			.replace(/dpi/g, '')
-			.replace(/(\d+)(cm|em|in|mm|pc|pt|px|rem)/g, function ($0, $1, $2) {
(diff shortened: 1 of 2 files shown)


jkimbo:
Looks like my build failed because of `Error: SAUCE_USER_NAME and SAUCE_API_KEY must be set in the environment` ? Is there something I need to include in the pull request?

triblondon:
The sauce tests only work when they run on the `Financial-Times` owner, not on forks.  Happy to push this to a local branch for CI if you have a fix you can verify locally.

jkimbo:
@triblondon thats fair enough. I've just pushed what I think is a fix for it based on https://developer.mozilla.org/en/docs/Web/CSS/resolution so can you check it?

**N.B.** to get the tests to fail you will need to run it them on a browser that doesn't have `matchMedia` i.e. IE 9 until https://github.com/Financial-Times/polyfill-service/issues/383 has been addressed.

triblondon:
Merged in #393

PR #393, opened 2015-04-28
Failing match media test

triblondon:
#382, but in a local branch


triblondon:
This is failing on the Promise polyfill in FF3.6, which is also failing on master, so I'm going to merge it.  I'm not sure how Promises began failing as it hasn't changed for yonks.

PR #396, opened 2015-05-06
Object.assign avoid arguments object deopt

maxnordlund:
For more information see https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments

### DIFF
--- a/polyfills/Object.assign/polyfill.js
+++ b/polyfills/Object.assign/polyfill.js
@@ -1,6 +1,6 @@
 Object.assign = function assign(target, source) {
-	for (var index = 1, key; index in arguments; ++index) {
-		source = arguments[index];
+	for (var index = 1, key, src; index < arguments.length; ++index) {
+		src = arguments[index];
 
 		for (key in source) {
 			if (Object.prototype.hasOwnProperty.call(source, key)) {
--- a/polyfills/Object.assign/polyfill.js
+++ b/polyfills/Object.assign/polyfill.js
@@ -2,9 +2,9 @@ Object.assign = function assign(target, source) {
 	for (var index = 1, key, src; index < arguments.length; ++index) {
 		src = arguments[index];
 
-		for (key in source) {
-			if (Object.prototype.hasOwnProperty.call(source, key)) {
-				target[key] = source[key];
+		for (key in src) {
+			if (Object.prototype.hasOwnProperty.call(src, key)) {
+				target[key] = src[key];
 			}
 		}
 	}


polyfills/Object.assign/polyfill.js
```diff
@@ -1,6 +1,6 @@
 Object.assign = function assign(target, source) {
-	for (var index = 1, key; index in arguments; ++index) {
-		source = arguments[index];
+	for (var index = 1, key, src; index < arguments.length; ++index) {
+		src = arguments[index];
```
samgiles, line 3: `src` is never used.  Did you intend to change references to `source` to `src`?

samgiles:
Nice spot.  See inline comment.  I think we could do with some tests for this feature :).

polyfills/Object.assign/polyfill.js
maxnordlund, line 3: Yes, I missed it. Will fix it right away.

samgiles:
Thanks!

jonathantneal:
Optimization questions:
- Would reusing the `source` variable declared in the parameters perform better? [cite](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object.assign/polyfill.js#L1)
- Would a while loop perform better? [cite](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object.assign/polyfill.js#L2)
- Would moving the declaration of `key` to within the loop when it is first needed perform better? [cite](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object.assign/polyfill.js#L5)

``` js
Object.assign = function assign(target, source) {
    var index = 0;

    while (++index < arguments.length) {
        source = arguments[index];

        for (var key in source) {
            if (Object.prototype.hasOwnProperty.call(source, key)) {
                target[key] = source[key];
            }
        }
    }

    return target;
};
```
- Would `Object.prototype.hasOwnProperty` perform better if it were abstracted to `_hasOwnProperty`, like `_enqueueMicrotask` and `_mutation`?

``` js
Object.assign = function assign(target, source) {
    var index = 0;

    while (++index < arguments.length) {
        source = arguments[index];

        for (var key in source) {
            if (_hasOwnProperty.call(source, key)) {
                target[key] = source[key];
            }
        }
    }

    return target;
};
```

samgiles:
@jonathantneal re-assigning to arguments is a bad idea if you're using `arguments`. I think it's generally safer to create an alias.

A while loop _may_ perform better in older browsers only because you can remove an operation, otherwise the interpreted code is mostly the same.  To realise this with `while`, count backwards because `0` is falsey - incidentally, when optimising C, this pattern is also suggested as a comparison with 0 is suggested to avoid an extra `lt` or `gt` instruction:

``` JS
var index = arguments.length;

while (index--) {
  // ... access[index]
}
```

Although it's arguably less clear.

On allocating `key` only when necessary, I'm pretty sure that doesn't matter, typically, afaik, stack space will be allocated anyway to avoid having to allocate on an as need basis, allocating space for an extra variable is way cheaper than allocating for each variable.

In engines where there is a JIT, all of this is invalid anyway.  I've confirmed that this function is optimisable by V8 using `optical` https://github.com/samgiles/optical

PR #401, opened 2015-05-21
Added Object.prototype.toString()

pygy:
Hello,

`Object.prototype.toString()` is sometimes used as a saner type checker. For example:

``` JavaScript
type = {}.toString;

type.call("foo") === "[object String]"
type.call(new String("foo")) === "[object String]"

type.call([]) === "[object Array]"
// etc...
```

It has a lot of advantages over `typeof` and `instanceof` for the above types.

In modernish Browsers (ES 5.1) the following also holds:

``` JavaScript
type.call(null) === "[object Null]"
type.call(undefined) === "[object Undefined]"
```

IE < 9, FF <4 and maybe other browsers however return `'[object Object]'` for these types.

This PR attempts to correct this issue.

---

It is not ready for merging:
- At this point, I haven't been able to get the tests to run. `grunt dev` launches the server, but visiting `http://127.0.0.1:3000/test/tests?feature=Object.prototype.toString&mode=all|control|targetted` gives just this:

```
Object.prototype.toString
    has testing suite ‣
    has feature detection
```
- I've copied the `browser` field for the config file from the `Object.keys` entry which was matching the IE and FF versions that I know to be problematic. No idea about Chrome, Safari and Opera.
- I don't have older browsers at hand to test it.

Any help is welcome :-)


polyfills/Object.prototype.toString/test.js
```diff
@@ -0,0 +1,28 @@
+var toString = {}.toString
+
+[
```
samgiles: You might want to do something like `var testData = [...]; testData.forEach(..);`  Because I was seeing `undefined is not a function` errors otherwise.

samgiles:
Just rename the 'test.js' file to 'tests.js' :).

The SauceLabs build we'll do will test on older browsers.

No need to change `compat.json` we automatically generate this from the tests run on SauceLabs.

Thanks for the PR!

pygy:
D'oh! Thanks for the reply (-:

How do I set up the `config.json` so that the detection is performed on as many browsers as possible? Should I just disable anything but IE and FF?

Edit: Still no luck with the tests.

polyfills/Object.prototype.toString/test.js
pygy: It doesn't help :-/

pygy:
The comments of @kitcambridge [here](https://github.com/es-shims/es5-shim/issues/304#issuecomment-104705988) are also relevant...

ljharb:
Why does this need shimming? Which engine that exists has differing behavior for `Object#toString`? (repeated from https://github.com/es-shims/es5-shim/issues/304#issuecomment-104720540)

barneycarroll:
The OP clearly describes the target environments (BTW Kyle Simpson is working on a jsperf/caniuse-type validitor for JS engines which could provide some non-anecdotal data for the support grid — you contributors should totally talk). 

I agree with @ljharb that this feels kind of pyrrhic (but for different reasons) — but it's totally legitimate as a patch. As a procedural, on-demand service, it's not within polyfill-service's remit to say "you shouldn't want this in the first place" / "nobody expects IE8 to do X anyway".

@pygy FWIW, IMO the root problem being addressed here is that `typeof` and `instanceof` are unshimmably broken APIs and that `Object.prototype.toString` is a convoluted hack that kind of stands in for them. I use (npm) `type-of-is` to offer an opinionated and altogether better API.

ljharb:
My bad, I'd misread it. However, `foo === null || foo === void 0` covers `null`/`undefined` in all browsers, and is very fast, so I don't see the advantage to shimming this one.

polyfills/Object.prototype.toString/polyfill.js
```diff
@@ -0,0 +1,6 @@
+var Object_prototype_toString = Object.prototype.toString
+Object.prototype.toString = function() {
+    if (this === null) return "[object Null]";
+    if (this === undefined) return "[object Undefined]";
+    return Object_prototype_toString.call(this);
+}
```
triblondon, line 6: I suggest returning the new function from an IIFE into which you pass the existing native impl to avoid creating a variable that exists across all polyfills.

docs/assets/compat.json
```diff
@@ -801,6 +801,41 @@
...
+      "7": "native"
+    },
+    "android": {
+      "4.4": "native",
+      "4.3": "native",
+      "4.2": "native",
+      "4.1": "native",
+      "4.0": "native"
+    }
+  },
```
triblondon, line 838: Don't edit this file - it's generated by a grunt task.

triblondon:
Other than the comments I've raised I see no reason not to accept this, though I don't think it's worth including this in the default set.

polyfills/Object.prototype.toString/polyfill.js
```diff
@@ -0,0 +1,6 @@
+var Object_prototype_toString = Object.prototype.toString
+Object.prototype.toString = function() {
+    if (this === null) return "[object Null]";
```
JakeChampion, line 3: I don't believe this will work as per 15.3.4.4 from the ES3 specification:

> Function.prototype.call (thisArg [ , arg1 [ , arg2, … ] ] )
> ...
> If thisArg is null or undefined, the called function is passed the global object as the this value. Otherwise, the called function is passed ToObject(thisArg) as the this value.

JakeChampion, line 3: For reference, it seems this was discussed somewhere else and came to the same conclusion - https://github.com/es-shims/es5-shim/issues/304#issuecomment-104705988

JakeChampion:
Closing this as it turns out to [not be possible](https://github.com/Financial-Times/polyfill-service/pull/401#discussion_r34554521).

polyfills/Object.prototype.toString/polyfill.js
ljharb, line 3: It's faster and clearer for a caller to explicitly check null and undefined anyways before using `Object#toString`

PR #407, opened 2015-06-09
add test and fix matchMedia operator "only"

tuukka:
Support for the matchMedia operator "only" was broken (IE9 error "SCRIPT1006"). The polyfill code only matched in the beginning of the string while there was always parts of the template string before the operator.


triblondon:
That makes much more sense :-)

PR #377, opened 2015-04-08
Allow polyfills to be organised into subfolders in the source tree

triblondon:
This PR allows polyfills to be organised into subfolders, but also permits the current structure.  This is intended to make it easier to maintain the project after adding large polyfills such as Intl, as discussed in #108.

Issue #108, opened 2014-10-02
Including the Intl polyfill


juandopazo:
Hi,

I'd love to help get the [Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) polyfill in this service. It's complicated because it includes a lot of extra files besides the main polyfill. See https://github.com/andyearnshaw/Intl.js/tree/master/locale-data/jsonp.

If we can agree on a way to include it, I'll be happy to send a pull request.

triblondon:
@juandopazo where do you feel like we should take this one?

caridy:
More and more people asking us how to load the Intl polyfill, let's make this happen.

slorber:
+1 and it would be nice to be able to specify by query string the languages we want to load data for

johanneslumpe:
Just as @slorber said, being able to only load the locales we want would be a huge plus here. Instead of having to push down all the available locales out there.

But in general: really looking forward of having this available.

sebastienbarre:
@johanneslumpe This doesn't do the trick for you?

```
  require('intl/Intl');
  require('intl/locale-data/jsonp/en.js');
  require('intl/locale-data/jsonp/fr.js');
```

to load en/fr, as opposed to:

```
require('intl'); 
```

Which brings everything.

johanneslumpe:
@sebastienbarre No that actually works :) And I'll stick with that for now. But it might be nice to now having to bundle this for browsers which already support `Intl`

juandopazo:
@triblondon we need to answer two questions:
1. Do we want the Intl source code in this package or to just reuse it from the npm package?
2. Do we treat locales as polyfills or as a special case like `features=Intl&locales=en-US,fr-FR`?

triblondon:
The easiest answer here that requires no changes to the service is:
- Add `Intl` as a polyfill with the core lib, copied from its source repo, but with repo URL listed in config to allow it to be updated easily
- Add additional separate polyfills for each locale under a sensible naming scheme.  I'm not sure whether there's a property available in `Intl` that represents a locale - if there is, then it can be named for that, eg `Intl.localesAvailable.en-US`.  However I suspect this doesn't exist, in which case strictly speaking the locale polyfills should have a `~` prefix, like HTML5shiv does, and perhaps we'd call them eg. `~IntlLocale.en-US`.  This prefix indicates that the name of the polyfill is not a JavaScript global.

[...]

PR #410, opened 2015-06-16
adds grunt task to general intl polyfill structure

caridy:
related to #377 and #108.

``` bash
$ grunt intl
Running "clean:intl" (clean) task
>> 2 paths cleaned.

Running "importintl" task
Importing Intl polyfill from /Users/caridy/repo/polyfill-service/node_modules/intl/dist/Intl.js
Importing Intl.~locale.* polyfill from /Users/caridy/repo/polyfill-service/node_modules/intl/locale-data/jsonp
717 imported locales
Intl polyfill imported successfully
```

TODO: 
- [x] grunt task to copy files from intl npm pkg.
- [x] commit all polyfills (I haven't done this to keep the DIFF manageable)
- [ ] update docs.


caridy:
@triblondon the good news is that it works just fine, e.g.: `http://localhost:3000/v1/polyfill.js?features=Intl,Intl.~locale.en,Intl.~locale.ar-001`

the bad news is that now it takes more than 2 minutes to run `grunt dev`. I suspect those 717 locales (which are treated as polyfills) are responsable for the degradation, but I haven't look into the details yet.

/cc @juandopazo

caridy:
Two more notes:
- sourcemap information is not been removed from the source (there is a grunt task for that).
- the order from the url is not preserved. this is not a big deal because it happens to work just fine, but when I use urls like: `http://localhost:3000/v1/polyfill.js?features=Intl,Intl.~locale.en,Intl.~locale.ar-001`, I see `ar-001` locale data appearing before `en`, but since `Intl` is there first, the order really does not matter much, but it is something to keep in mind.

triblondon:
Thanks @caridy.  Some thoughts:
- The import mechanism seems like something that we might want to run for other polyfills - I know of several that are imported from external repos (manually at the moment) and being able to update them would be good.  Could we maybe consider adding an `update.task.js` file as a generic solution to this problem, as part of the polyfill files?  Then a generic task could 
- Intl doesn't strike me as something that should be a project dependency, unless we plan to import it at build time and not commit it to the library.  But I'm in favour of removing the dependency and handling the downloading of the source as part of the update task (especially when taken with the point above, because other polyfills may need to be downloaded via different mechanisms)
- I don't think the order of locales is important and don't propose we do anything about it. The service sorts the output in dependency graph order, so items that have no dependency relationship with one another may appear in either order.

I haven't tried running it yet - could you push it to a branch in the Financial-Times fork so I have push access to it and we can run the CI process?

caridy:
> The import mechanism seems like something that we might want to run for other polyfills - I know of several that are imported from external repos (manually at the moment) and being able to update them would be good. Could we maybe consider adding an update.task.js file as a generic solution to this problem, as part of the polyfill files? Then a generic task could

I'm not so sure understand what you mean. For my conversation with @juandopazo, it seems that `grunt dev` is also applying some minification process (I want to confirm this), but if that's the case, we should have a way to flag that since all locales data are already minified, and we should serve the same, that might speed up the process. Another idea will be to have a hierarchical configs and detection routine to shrink the number of files to be analyzed since all those locales will have the exact same detection as Intl, and the exact same config. Will that speed up the process and reduce the metadata in memory?

> Intl doesn't strike me as something that should be a project dependency, unless we plan to import it at build time and not commit it to the library. But I'm in favour of removing the dependency and handling the downloading of the source as part of the update task (especially when taken with the point above, because other polyfills may need to be downloaded via different mechanisms)

it is a dev dependency, and it is meant to be used only when running `grunt intl` manually.

> I don't think the order of locales is important and don't propose we do anything about it. The service sorts the output in dependency graph order, so items that have no dependency relationship with one another may appear in either order.

agreed. if we need to, we can always add the dependency relationship between locales and intl.

andyearnshaw:
Nice work, @caridy :smile:.

The only thing I might add is that the order in which locales are loaded is (kind of) important only for the very first one, because it [becomes the default](https://github.com/andyearnshaw/Intl.js/blob/f5ab43f764d6cf1be61731f7c57b325ac4c2b6d6/src/core.js#L2765).  This could probably be changed in the polyfill to set the default to the best match for `navigator.language || navigator.userLanguage` instead, though (after it's loaded, that is).

caridy:
@andyearnshaw, yeah. we should probably change that in the polyfill. Not a big deal since no many people are loading more than one locale AFAIK.

triblondon:
It's mainly an issue of encapsulating polyfill related stuff in the relevant part of the polyfill library structure.  I don't like having Intl as a top level project dep because it relates to only one specific polyfill in the library, it is not a generic behaviour.  Same applies to the changes to the gruntfile and the addition of a polyfill specific task - all these things put Intl-related stuff outside of the Intil directory in polyfills.

What I'm proposing is something like this:

```
polyfills/
    Intl/
        config.json
        update.task.js   <-- Grunt task
        polyfill.js
        locales/
            locale1
            localen
```

Then we change the global Gruntfile to import all the tasks from the polyfills so you could do something like `grunt updatelibrary` to run all the update tasks.

I actually didn't address the speed of `grunt dev`, which is probably where we're misunderstanding one another, sorry.  On that point, the slowness is indeed probably in the minification. The `buildsources` task does two slow things: minification and transpilation of ES6, but it only does the latter for polyfills written using ES6 syntax which declare that in their config.json.  So in practice, it's minification.

The solution is probably to be smarter about minifying only where the source is not already minified.  How can we detect that?  Or should there be an opt-out property explicit in config.json?

caridy:
> The solution is probably to be smarter about minifying only where the source is not already minified. How can we detect that? Or should there be an opt-out property explicit in config.json?

Is either opt-in or opt-out. Imported libraries, like Intl can probably skip that process by opting out (and it will be easier to implement opt-out at this point I guess). I doubt people will debug those external/complex polyfills by using this service.

caridy:
@triblondon, I have updated the PR with a new configuration option called `external`, when set to true (falsy by default), it will skip any validation of minification process. This definitely solves the problem with `grunt dev`, which now works pretty fast.

caridy:
> Then we change the global Gruntfile to import all the tasks from the polyfills so you could do something like `grunt updatelibrary` to run all the update tasks.

I'm ok with this. But, I don't think we should pretend we can solve this without adding extra dev dependencies (in this case, `intl`). My point here is: when allowing custom grunt tasks, most likely, developers will have to rely on extra dev dependencies, whether those are polyfill libraries to be copied over, of just build tools. In other words, isolation of the service can't be happening if there is a command like `grunt updatelibrary` that can be executed from the main app. In any case, those devDependencies will NOT affect the service in any way.

triblondon:
OK, rather than external, I suggest a `build` property in config.json, to allow polyfills to control to some extent how the service should build them.  We can put minify in there as a first property.  I'd move esversion in there too, and other contributors are looking at enabling browserifying of commonJS-packaged polyfills so that would be another build opt-in.

```
{
  browsers: { ... }.
  build: {
    minify: false,    <-- defaults to true
    esversion: 6,    <-- defaults to 5
    module: "commonjs"   <-- defaults to 'global'
  }
}
```

What do you think?

Regarding the devDependencies, I don't strongly object to having Intl as a devDependency provided that everything else is in the intl polyfill directory and the top level grunt tasks are generic.  I see your point that any custom tasks embedded with polyfills are going to want to use npm modules that we don't already have as project dependencies, so no reason not to include the polyfill source itself as well.

I'd be happy to help with this but it needs to be on a branch I can push to.  Can we collaborate in the financial-times org?  You have an outstanding invite to be a collaborator.

caridy:
- about the build: partially agreed, I'm not a fan of true/undefined paradox jejeje. but we can go with that for the sake of this PR, I will update the task to rely on build->minify, and then you can update all others to use `esversion` and `module`
- about the devDep: thanks, I will flush the initial work for the generic task.
- about the invite: accepted

caridy:
- [x] `grunt updatelibrary` now picks any `polyfills/**/update.task.js` and runs them.
- [x] if `build->minify` is set to false, minification will be skipped.
- [x] intl polyfill updated to the latest.

Gruntfile.js
```diff
@@ -83,6 +83,14 @@ module.exports = function(grunt) {
 					failOnError: true
 				}
 			}
+		},
+		"updatelibrary": {
+			options: {
+				expand: true
+			},
+			tasks: {
+				src: ['polyfills/**/update.task.js'],
```
caridy, line 92: in case we want to have multiple different groups of polyfills in the future.

package.json
```diff
@@ -67,6 +67,8 @@
     "grunt-contrib-watch": "^0.6.1",
     "grunt-mocha": "^0.4.11",
     "grunt-simple-mocha": "^0.4.0",
+    "intl": "~1.0.0-rc-4",
+    "rimraf": "^2.4.0",
```
caridy, line 71: rimraf is used to remove folders async from `update.task.js`. this is a side effect of not allowing people to define `clean` configuration in `Gruntfile.js`, but I guess that's fine.

polyfills/Intl/config.json
```diff
@@ -0,0 +1,18 @@
...
+		"ie": "<11",
+		"firefox": "<29",
+		"opera": "<15",
+		"chrome": "<24",
+		"safari": "*",
+		"ios_saf": "*"
+	},
+	"dependencies": [],
+	"build": {
+		"minify": false
```
caridy, line 12: :fallen_leaf:

polyfills/Intl/update.task.js
```diff
@@ -0,0 +1,58 @@
...
+var fs = require('fs');
+var path = require('path');
+var rimraf = require('rimraf');
+var IntlSourcePath = require.resolve('intl/dist/Intl.min.js');
+var LocalesPath = path.dirname(require.resolve('intl/locale-data/jsonp/en.js'));
+var IntlPolyfillOutput = path.resolve('polyfills/Intl');
+var LocalesPolyfillOutput = path.resolve('polyfills/Intl/~locale');
+
+// this is not really a grunt task, but a function that is suppose
+// to be sync, and receive access to grunt instance for convenience.
```
caridy, line 12: This is the part that I don't like, but I could find a reliable way to make update.task.js a real grunt task that can be loaded and executed dynamically.

polyfills/Intl/update.task.js
```diff
@@ -0,0 +1,58 @@
...
+module.exports = function(grunt) {
+	var detectFileSource = fs.readFileSync(path.join(IntlPolyfillOutput, 'detect.js'));
+	var configSource = require(path.join(IntlPolyfillOutput, 'config.json'));
+
+	// deleting existing files
```
caridy, line 17: this is the side effect that I mentioned above.

tasks/updatelibrary.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+var path = require('path');
+
+module.exports = function(grunt) {
+
+	grunt.registerMultiTask('updatelibrary', 'Update Polyfill imported from external libraries', function() {
+		this.files.forEach(function (file) {
+			grunt.log.writeln('Running ' + file.src + ': ');
+			require(path.resolve(file.src.toString()))(grunt);
```
caridy, line 10: dummy task to talk and execute any `update.task.js` files expanded by grunt.

caridy:
let me know what do you think so I can commit the generated files to get the tests to pass.

caridy:
@triblondon any update on this? should I commit all files?

triblondon:
Sorry for delay.  I'm up to my ears prepping for edgeconf.com.  Can you give me a couple of days?

polyfills/Intl/update.task.js
```diff
@@ -0,0 +1,58 @@
...
+	grunt.log.writeln('Importing Intl polyfill from ' + IntlSourcePath);
+
+	var IntlPolyfillSource = fs.readFileSync(IntlSourcePath);
+	grunt.file.write(path.join(IntlPolyfillOutput, 'polyfill.js'), IntlPolyfillSource);
+
+
+	grunt.log.writeln('Importing Intl.~locale.* polyfill from ' + LocalesPath);
+
+	var locales = fs.readdirSync(LocalesPath);
+    locales.forEach(function (file) {
```
triblondon: Possible indentation issues here

triblondon:
This looks really good.  Thanks!

CI will work if you push it to a branch in the Financial-Times org.

caridy:
will get that ready today.

JakeChampion:
Closing as #419 supercedes this PR.

PR #412, opened 2015-06-19
Element.prototype.classlist for svg elements

jkimbo:
Make sure classList works on svg elements

### DIFF
--- a/polyfills/Element.prototype.classList/polyfill.js
+++ b/polyfills/Element.prototype.classList/polyfill.js
@@ -2,7 +2,8 @@
 	Object.defineProperty(Element.prototype, 'classList', {
 		get: function () {
 			function pull() {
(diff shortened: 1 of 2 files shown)


triblondon:
I'm getting 3 test failures, in IE 6, 7 and 8.

triblondon:
If you have a Sauce Labs account and you set the appropriate env vars (`SAUCE_USER_NAME`, `SAUCE_API_KEY`) you can run `grunt ci` to use Sauce to run tests across our key browser targets.  Otherwise just fire up some VMs :-)

jkimbo:
@triblondon found the issue, IE8 and below doesn't support svg at all...

I've added a check to skip the test if there isn't svg support since it's not particularly relevant.

triblondon:
Looks good.  See #423 for local PR.

PR #415, opened 2015-06-29
Do not polyfill fetch for supporting browsers

hannesj:
As per [this](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch/fetch#Browser_compatibility) these browsers now support the fetch API.


samgiles:
Great, thanks!

PR #424, opened 2015-07-06
Symbol

JakeChampion:


polyfills/Symbol/tests.js
```diff
@@ -0,0 +1,165 @@
...
+	expect(numberKeyFor).to.throwError();
+	expect(arrayKeyFor).to.throwError();
+	expect(objectKeyFor).to.throwError();
+	expect(boolKeyFor).to.throwError();
+	expect(undefinedKeyFor).to.throwError();
+	expect(symbolKeyFor).to.not.throwError();
+});
+
+it('Symbol.keyFor should return undefined if can not find symbol in global registry', function() {
+	expect(Symbol.keyFor(Symbol('a'))).to.be.undefined;
```
samgiles: Isn't there a possibility that 'a' could be in the global symbol registry?  Not sure we can guarantee the order of execution: the test at line 85 running before this test might cause this one to fail erroneously.

samgiles:
Missing `Symbol.prototype.toString` (19.4.3.2 ) and `String(_sym_)` (21.1.1.1) tests? Only mention it because it covers the abstract operation: `SymbolDescriptiveString` which is probably useful for debugging.

mehcode:
How is this going? This would be a nice polyfill to have. 

Is there something I can do to help it along?

triblondon:
@mehcode pinging @JakeChampion to see where he got to with this.

JakeChampion:
Work has moved over to -- https://github.com/Financial-Times/polyfill-service/pull/585

PR #425, opened 2015-07-07
Add shrinkwrap file

JakeChampion:


triblondon:
We had a shrinkwrap file back in the day.  Found it really annoying for contributors, so got rid of it.

JakeChampion:
@triblondon If I understand the release process correctly, the shrinkwrap file is required because we don't push the same artifact from qa to live. Since we do not re-use the artifact, during our QA period of 7 days if any of our dependencies pushes an update that is within our specified semver ranges, we would be using an untested dependency in production.

triblondon:
That makes sense. Ok, let's give it another go.

PR #426, opened 2015-07-07
Switch the range for _enqueueMicrotask back to IE 9 and above

JakeChampion:


triblondon:
Looks fine to me

matthew-andrews:
weird error …… suspect there may be a bug in travis at the moment where environment variables are not available to pull request builds………

JakeChampion:
@matthew-andrews @triblondon 
 As the push build has passed, is this okay to merge?

matthew-andrews:
got this back from travis:-

> Hey Matthew,
> Thanks for reporting and sorry about the troubles.
> It looks like a bug on our side, which our devs are already looking into it. We'll write a follow-up e-mail as soon as we know some more.
> Thanks again for you help and if you run into any other issues - please let us know, we'll be happy to help.
> Cheers,
> María

up to @triblondon whether to merge this or not but i would suggest not not merging because of the failed pr build :smile:

JakeChampion:
I have ran the test suite from my machine and all passed. :+1:

PR #428, opened 2015-07-07
Cache node_modules on travis

JakeChampion:
Since we are using [container-based infrastructure](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) on Travis, we can cache assets for free.

Considering the biggest bottleneck in our testing is SauceLabs, don't expect a big decrease in CI executing time from this change.

Are there any other assets we should be caching?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,6 @@ before_script:
   - npm install -g grunt-cli
 script:
   - npm run-script citest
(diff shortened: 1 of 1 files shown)


samgiles:
Looks good.  Feel free to merge.

No other assets for CI I believe.

PR #431, opened 2015-07-08
Switch from .semver to package.json version

JakeChampion:
Since we will be publishing the rc versions to NPM, we might as well avoid duplicating the version number in 3 places (git tag, .semver, package.json) and use the package.json version number as the value used within the application.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 /polyfills/sources.json
 /test/results
 /.env.json
-/.semver
 /.haikro-cache
 /node-*
 /service-process.pid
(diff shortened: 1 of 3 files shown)


matthew-andrews:
you could eliminate the package.json one too :) https://github.com/matthew-andrews/npm-prepublish

JakeChampion:
Once this is merged we will need to bump the rc version up and publish to QA as well as NPM.

@triblondon What do you think about using @matthew-andrews' `npm-prepublish` package? As you are the owner of the NPM package it's really down to if you mind Travis publishing packages under your account.

matthew-andrews:
oh `npm-prepublish` doesn't necesarily force you to publish from Travis.  it just eliminates the need for version in package.json.  publishing from travis is something that doing this enables and you might like to do in addition :)

PR #433, opened 2015-07-15
Update Object.getOwnPropertyNames

jeremykohn:
`for ... in` iterates over all properties, including those from the prototype chain. However, `getOwnPropertyNames` should only return properties for the specified object, not its ancestors. To correct this, the polyfill can use `hasOwnProperty` as a condition to filter the returned properties.


triblondon:
Will test on local branch for CI

triblondon:
This needed tests, and having added them I found that the IE8 variant of the polyfill wasn't working.  Have used the generic impl in all IEs, which is working and passing my rudimentary tests.

See https://github.com/Financial-Times/polyfill-service/pull/439

PR #434, opened 2015-07-16
Change TTL for polyfills to 1 week

richard-still-ft:


service/index.js
```diff
@@ -29,7 +29,7 @@ var contentTypes = {".js": 'application/javascript', ".css": 'text/css'};
 
 // Default cache control policy
 app.use(function(req, res, next) {
-	res.set('Cache-Control', 'public, max-age='+one_day+', stale-while-revalidate='+one_week+', stale-if-error='+one_week);
+	res.set('Cache-Control', 'public, max-age='+one_week+', stale-while-revalidate='+one_week+', stale-if-error='+one_week);
```
matthew-andrews, line 32: the stale bits don't really do anything anymore :relaxed:

triblondon, line 32: Hey, hold on a minute. Read the spec!

image

matthew-andrews, line 32: oh wow ok , sorry for the confusion @richard-still-ft  …

richard-still-ft, line 32: I should have read the spec!

samgiles:
There's no harm in increasing the `stale-while*` values though if we're willing to increase default `max-age` to a week. Previously `stale-while*` values were 7x higher. :information_desk_person:

PR #437, opened 2015-07-21
Added the PictureFill polyfill for the Picture element

jonathan-fielding:
Hi

This is a pull request to add support for the Picture element in unsupported browsers using the Polyfill service

Thanks

Jonathan


docs/assets/compat.json
```diff
@@ -3007,6 +3007,41 @@
...
+      "7": "polyfilled"
+    },
+    "android": {
+      "4.4": "polyfilled",
+      "4.3": "polyfilled",
+      "4.2": "polyfilled",
+      "4.1": "polyfilled",
+      "4.0": "polyfilled"
+    }
+  },
```
triblondon, line 3313: Don't edit this file.  It gets generated automatically by automation.

polyfills/Picture/config.json
```diff
@@ -0,0 +1,19 @@
...
+		"ios_saf": "*"
+	},
+	"dependencies": [
+	],
+	"license": "MIT",
+	"spec": "https://html.spec.whatwg.org/multipage/embedded-content.html#embedded-content",
+	"repo": "https://github.com/scottjehl/picturefill",
+	"notes": [
+		"Standard picturefill polyfill used to enable support in older browsers"
+	],
```
triblondon: This note could be removed - it doesn't signal any important gotchas

triblondon:
Thanks Jonathan.  Aside from the minor issues above, are there any tests that can be added?  We like to have a test for at least the main core use cases, so that our test framework can check browser compatibility.

triblondon:
Also, it would be good to have an update task.  This is a new feature that enables third party polyfills to be kept up to date automatically.  I've just pushed some documentation for how to create an update task in https://github.com/Financial-Times/polyfill-service/commit/d7cc2cd0a2a1d7afbad6b706d77172502080433d.

There's a good example of an existing task in https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Intl/update.task.js

docs/assets/compat.json
jonathan-fielding, line 3313: If this is generated automatically is there any reason why it is not in the .gitignore?

triblondon, line 3313: it needs to be committed - it's used to display the compatibility chart in
the documentation.  I suppose we could add it during deployment but the
full compat test takes about 2 hours, so I'd rather not block deploys on
that.

jonathan-fielding, line 3313: ok removed my changes

polyfills/Picture/config.json
jonathan-fielding: removed note

jonathan-fielding:
Update task has been added

polyfills/Picture/update.task.js
```diff
@@ -0,0 +1,33 @@
...
+	// deleting existing files
+	if (grunt.file.exists(PictureFillPolyfillOutput + 'polyfill.js')) {
+		try {
+			// Actually delete. Or not.
+			rimraf.sync(PictureFillPolyfillOutput + 'polyfill.js');
+		} catch (e) {/* ignore */}
+    }
+
+	// customizing the config to add Picture as a dependency
+	configSource.dependencies.push('Picture');
```
JakeChampion: I think the dependency might be the ~html5-elements/polyfill.js

jonathan-fielding: Just committed the change, is that what you mean? https://github.com/Financial-Times/polyfill-service/commit/34bfe544cf59d825250bccfbc3b4d83bb1ae62b2

triblondon:
Implemented in #448

PR #438, opened 2015-07-22
Stop using document.head as a hook in querySelector

alexander-akait:
remove head, we need search in node dom, not in document, because this code not working if search in document. Nodes not available in document

``` javascript
var progress = document.createElement('div');
var progress.id = 'nprogress';
var progress.innerHTML = '<div class="bar" role="bar"><div class="peg"></div></div><div class="spinner" role="spinner"><div class="spinner-icon"></div></div>';
var bar = progress.querySelector('.bar');

var spinner = progress.querySelector('.spinner');
spinner && removeElement(spinner);

document.body.appendChild(progress);
```

Also we need ";", some early version IE7 bugs without this symbols

### DIFF
--- a/polyfills/document.querySelector/polyfill.js
+++ b/polyfills/document.querySelector/polyfill.js
@@ -1,20 +1,17 @@
 (function (global) {
-	var
-	head = document.getElementsByTagName('head')[0];
-
 	function getElementsByQuery(node, selector, one) {
 		var
 		generator = document.createElement(),
 		id = 'qsa' + String(Math.random()).slice(3),
 		style, elements;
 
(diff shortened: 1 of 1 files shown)


triblondon:
Thanks for this.  Can you add a test that fails without this patch?

alexander-akait:
@triblondon, well, as soon as I get time, I will do tests

triblondon:
Thanks.  Meanwhile I've accepted the `;` fix, committed to master in 41d0227888e8c04f308a9e2b3303dc84fc054892

triblondon:
Ping @evilebottnawi, will you be able to do the test for this?

PR #439, opened 2015-07-22
Object get own property names patch

triblondon:
From https://github.com/Financial-Times/polyfill-service/pull/433


samgiles:
LGTM, maybe we should add a test to cover the first conditional branch, maybe test whether `null` and `undefined` throws the exception?

polyfills/Object.getOwnPropertyNames/polyfill.js
```diff
@@ -1,12 +1,27 @@
...
+		throw new TypeError('Cannot convert undefined or null to object');
 	}
 
-	var buffer = [], key;
+	object = Object(object);
 
+	// Enumerable properties only
 	for (key in object) {
-		buffer.push(key);
+		if (object.hasOwnProperty(key)) {
```
ivoreis, line 17: Have you tested this on IE8?  

The implementation found in this commit (3385d9728b42f3fd9302fa6511634a591babc0ad) works for ie8. Do you need a PR for this?

PR #441, opened 2015-07-24
Fix npm licence SPDX conformity warning by using the CC0-1.0 identifier

marionebl:
This should silence the warning issued when installing polyfill-service with recent npm versions:

```
npm WARN package.json polyfill-service@0.0.0 license should be a valid SPDX license expression
```

See [spdx.org](https://spdx.org/licenses/CC0-1.0.html) for reference.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "test": "grunt test",
     "citest": "grunt ci --stack"
   },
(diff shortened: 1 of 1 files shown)


samgiles:
Thanks!

PR #442, opened 2015-07-24
Provide bin entry, bin for global installation and CLI usage

marionebl:
This
- adds a `bin` entry to package.json, exposing `service/index.js` as `polyfill-service` when installed globally or for npm run-script when installed locally
- adds `bin/polyfill-service` as central entry point for starting the service
- pulls the argument parsing / defaults from `service/index.js` into `bin/polyfill-service`
- makes `service/index.js` export a function that accepts the parameters `port` and `callback`:
  - `port`: Port to run the service on
  - `callback`: callback executed after service has started
- adapts other npm run-scripts to new entry point
- adapts the `grunt:service:polyfillservice` task to work with new entry point

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -74,9 +74,7 @@ module.exports = function(grunt) {
 		},
 		"service": {
 			polyfillservice: {
-				shellCommand: process.argv[0] + ' ' + __dirname + '/service/index.js',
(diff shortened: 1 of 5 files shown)


triblondon:
This seems like a nice improvement to the architecture of the app.  Will review in detail shortly.

PR #445, opened 2015-07-29
polyfill querySelector in ie8

wheresrhys:
According to caniuse the native implementation is incomplete

> Partial support in IE8 is due to being limited to CSS 2.1 selectors and a small subset of CSS 3 selectors (see notes there). Additionally, it will have trouble with selectors including unrecognized tags (for example HTML5 ones).


triblondon:
It would help a lot if you could check the CI passes when submitting a PR.

In this case there's a test case that depends on the previous config of queryselector.  I've updated it.

PR #448, opened 2015-07-30
Picturefill (local for #437)

triblondon:
See #437


triblondon:
OK, currently fails the detect in all browsers :-)

@jonathan-fielding try running the service with `grunt dev` and then load http://localhost:3000/test/tests?feature=Picture&mode=targeted in Safari

jonathan-fielding:
Tests added, also got the detect working BUT am hoping @scottjehl gets back to me about why the picturefill polyfill doesnt expose HTMLPictureElement

triblondon:
@jonathan-fielding one other thing I've just realised - this needs to be called `HTMLPictureElement` not `Picture`.

scottjehl:
@triblondon picturefill polyfills all responsive image markup, not just `picture`. So `img[srcset]` and `img[sizes]` are in there too, independent of picture. Just in case that's helpful to know.

jonathan-fielding:
Updated the name, and the test

jonathan-fielding:
The new test didnt work so reverted

jonathan-fielding:
Ok thats working now and I have updated the unit tests run in polyfill service as they broke with new version of Picturefill

triblondon:
Asked Jonathan to remove the tests because they are testing the polyfill's proprietary API, not the spec.

PR #457, opened 2015-08-06
Add missing mobile browsers

callumlocke:
I think there might be some others with the same problem, so hold off merging this for now.

### DIFF
--- a/polyfills/Array/prototype/includes/config.json
+++ b/polyfills/Array/prototype/includes/config.json
@@ -8,7 +8,12 @@
 		"firefox": "*",
 		"ie": "*",
 		"opera": "*",
-		"safari": "*"
+		"safari": "*",
+		"ios_saf": "*",
+		"ios_chr": "*",
+		"android": "*",
+		"op_mob": "*",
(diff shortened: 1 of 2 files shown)


callumlocke:
OK that's the only two I care about right now, feel free to merge in, although there might be more with the same problem.

Am I right that currently we automatically assume a browser natively supports a feature if the config.json makes no mention of that browser?

Could there be an argument for this: if a particular UA is not mentioned in the config.json at all, the server concludes it might not have the feature, and gives it a gated version of the polyfill?

triblondon:
> Could there be an argument for this: if a particular UA is not mentioned in the config.json at all, the server concludes it might not have the feature, and gives it a gated version of the polyfill?

Arguably, yes.  But we're trying to be fairly conservative.

Thanks for this - looks good to me.

callumlocke:
> Arguably, yes. But we're trying to be fairly conservative.

Isn't what I'm suggesting more conservative – i.e. opting for 'gated polyfill' when support is unknown?

PR #459, opened 2015-08-10
polyfill Promise also for mobile IE11

tuukka:


triblondon:
Suspect this is an evolved mistake - @jonathantneal orignally configured it as just v10, when there probably wasn't a v11. Then it became \* - 10, but actually Promises aren't supported by any version of IE mob, so \* covers it nicely.

PR #463, opened 2015-08-12
Event/polyfill-ie8.js certain event listeners on window in IE8 never fire

tejacques:
Notifies users attempting to add event listeners to unlistenable events in IE8.


triblondon:
See https://github.com/Financial-Times/polyfill-service/pull/475

PR #464, opened 2015-08-13
Include the PID, if available, in the metric key

samgiles:


triblondon:
Nice, thanks!

triblondon:
No impact on polyfill output or API, so I'm going to sneak this into 1.6 without a new RC.

PR #465, opened 2015-08-17
Local UA cache

triblondon:
I spoke to @samgiles about pushing a PR to the useragent module to make its caching behaviour configurable, but decided not to for some reasons:
1. They initialise their LRU cache immediately on require, so getting the opportunity to specify a config is a bit awkward
2. Their cache stores Agent instances, which are larger than we really need them to be, but useful for other module users.  Our use case affords a more efficient approach.
3. We run our own processing to further normalise the output of the useragent module, which would not benefit from our ability to configure their cache.

Instead, I'm proposing a local cache that maps long UAs to an array, and does so after we've done our normalisation.


triblondon:
Updated to latest master to get CI working

PR #472, opened 2015-08-31
Async refactor for individual source caching

triblondon:
A couple of oddities:
- In the node tests, the "should understand the 'all' alias" test regularly takes ~700ms.  Not sure why.  This underlying slowness slows down any test with mode=targeted.
- When running the service with `grunt dev`, console output is delayed.  It appears in real time if running the service with `node bin/polyfill-service`

Any insight into those much appreciated, along with a general code review!


samgiles:
> In the node tests, the "should understand the 'all' alias" test regularly takes ~700ms. Not sure why. This underlying slowness slows down any test with mode=targeted.

This is weird, I can reproduce, but _only_ if the request to the polyfill endpoint is made through a page: 

screen shot 2015-09-01 at 11 12 24

Might be worth trying this: http://www.brendangregg.com/blog/2014-09-17/node-flame-graphs-on-linux.html?  (Only I don't want to spend time setting up a Linux VM or an environment on an existing one (shell01 maybe?)).

> When running the service with grunt dev, console output is delayed. It appears in real time if running the service with node bin/polyfill-service

I don't think this is as weird, most likely a problem with the stdout multiplexing grunt is probably doing - only speculation.  Probably a `grunt` bug.

lib/aliases.js
```diff
@@ -50,29 +48,34 @@ module.exports = AliasResolver;
 function applyResolverToIdentifiers(featureSet, nameResolverFunction) {
 	var _ = require('lodash');
 
-	Object.keys(featureSet).forEach(function(featureName) {
+	return Promise.all(Object.keys(featureSet).map(function(featureName) {
 		var feature = featureSet[featureName];
-		var resultItems = nameResolverFunction(featureName);
-		if (!Array.isArray(resultItems)) return true;
+		return Promise.resolve(nameResolverFunction(featureName)).then(function(resultItems) {
```
samgiles, line 53: ~~I wonder if there is slowness because `nameResolverFunction` is sync.~~  Doubt it

lib/index.js
```diff
@@ -63,62 +54,71 @@ function listPolyfills(options) {
...
+				});
 			}
-		}
-		delete features[featureName];
-	}
 
-	// chooseVariant modifies the features object, so use forEach rather than map and disregard return value
-	Object.keys(features).forEach(function(featureName) {
-		chooseVariant(featureName, features[featureName]);
+			// chooseVariant modifies the features object, so use forEach rather than map and disregard return value
```
samgiles: This comment is out of date.

samgiles:
Looks good.

I'm a little worried some of the code is getting a little dense.

PR #474, opened 2015-09-07
Configure IE polyfills to match IE Mobile

tuukka:
Currently most polyfills are not served to ie_mob although it's basically the same browser as ie > 9. This pull request duplicates the config definitions for ie > 9 to ie_mob.


triblondon:
Do you know that these polyfills work in IE mobile?

Also note that some version constraints are less-than, so in some cases you are, for example, applying a polyfill to ie_mob < 11, which would include versions under 9.

tuukka:
Thanks for your feedback! Indeed, I treated IE Mobile like an "unknown" variant of IE, using the same polyfills and versions by default without testing them. I was able to run `/test/director?mode=targeted` on IE Mobile 10 and got 51/51 all pass (compared to 5/5 all pass on cdn.polyfill.io). Baseline for ie_mob seems to be ">=8" so ie_mob 8 and ie_mob 9 remain untested. I found this document that includes a list of differences between IE9 on desktop and mobile: https://msdn.microsoft.com/en-us/library/dn629259%28v=vs.85%29.aspx

Would you consider a revised patch that set a lower bound 10 since that's the version I've been able to test?

triblondon:
Yes, that would be fine.  It wouldbe useful if you could also drop some screenshots of your test results.

tuukka:
The new commit adds version 10 as the lower bound. Here's a screenshot of `/test/director?mode=targeted` on IE Mobile 10:

wp_ss_20150915_0003

triblondon:
OK, LGTM,

Issue #317, opened 2015-02-26
Event/polyfill-ie8.js certain event listeners on window in IE8 never fire


tejacques:
### Bug

Currently if you use the polyfill service for Events in IE8, and try to add a `click` handler to the `window`, it will never fire, because in IE8 `click` events (and others, listed below) are fired on the `document` but not `window`.
#### Test Case

``` .js
it('should bubble to window when fired from element', function() {
    var fired = false;
    window.addEventListener('click', function() {
        fired = true;
        // cleanup
        window.removeEventListener(type, listener);
        document.body.removeChild(element);
    });
    var element = document.createElement('div');
    document.body.appendChild(element);
    element.dispatchEvent(new Event('click', { bubbles: true }));
    expect(fired).to.be(true);
}
```
#### Fix

Re-target event listeners for events triggered on the `window` in modern browsers, but `document` and not `window` in IE8 to the `document` when added to the `window`. Care needs to be taken to ensure that:
1. The listeners for `window` fire after `document` when events bubble up
2. Stopping propagation on the event from a listener on `document` properly stops the event from bubbling/triggering on the `window`
3. The listeners for `window` only fire if `window` is the target, or the event bubbles up.
##### Discussion

I'm currently working on this in a side-fork because I need it for a project (creating as close to a polyfill as possible for IE8 StorageEvents). I'm happy to upstream the changes if you're interested in a pull request for this issue, but I'm not sure whether the fix is actually worth the added complexity since most of the time developers will add handlers to the `document` instead of `window` for these events anyway:
- `click`
- `dblclick`
- `keypress`
- `keyup`
- `keydown`
- `mousedown`
- `mouseup`
- `mousemove`
- `mouseover`
- `mouseout`

`storage` is broken as-is though as modern browsers only trigger the `storage` event on `window` whereas IE8 triggers it on `document`. Unfortunately, just retargeting the `storage` event isn't sufficient because the `Storage` API on IE8 is busted

triblondon:
Wow, nice diagnosis, thanks.  I think you're right, this is a real edge case.  It wouldn't be hard to add a line to addEventListener to bind the listener to document instead of window, but as you say this potentially has unexpected consequences on stopPropagation.

Potentially the best thing to do here is to throw an error to the console if addEventListener is called on window for an element specific event, with a message that links to this issue.  What do you think?

tejacques:
That sounds like a good middle ground.

triblondon:
[...]

PR #475, opened 2015-09-07
Event/polyfill-ie8.js certain event listeners on window in IE8 never fire (local for #463)

triblondon:


triblondon:
See #317

triblondon:
Erroring in IE6 and 7

image

This is because the XHR polyfill creates an object that doesn't have a prototype, and the new code in Event uses instanceof.  Needs to be slightly more defensive.

PR #482, opened 2015-09-18
IE11 doesn't support Event constructor - adding to required browsers

### LINKED ISSUE #479: Event polyfill should be served to IE 11
The Event polyfill should be served to IE 11. The detect.js returns false in IE 11. Also the polyfill documentation says that IE 11 does received this polyfill.

annez:
This is a fix for `Event` support in IE11. The config was saying that it should polyfill up to IE10, but as can be seen from these tests, IE11 requires the `Event` polyfill as well.

[Throws in IE11](http://annez.github.io/event-test/)
[Works in IE11 with Event polyfill forced](http://annez.github.io/event-test/fix.html)
## Changes
- Updated `config.json` file to reflect this and polyfilling IE11

This should fix #479

### DIFF
--- a/polyfills/Event/config.json
+++ b/polyfills/Event/config.json
@@ -4,7 +4,7 @@
 	],
 	"browsers": {
 		"firefox": "6 - 10",
-		"ie": "9 - 10",
+		"ie": "9 - 11",
 		"ie_mob": "10",
 		"opera": "10 - 11.5",
 		"safari": "4 - 7"


triblondon:
Did you say on twitter, that IE11 is passing the tests without the polyfill loaded?  If so, we need to fix the test so that we can be more confident that the polyfill is (and continues to be) effective in IE/Edge

annez:
Sorry should have clarified - IE11 doesn't pass the tests with or without the polyfill loaded unless you force `Event` as a feature on polyfill.io. The test examples I have shown are with the production V1 polyfill loaded with and without `Event`.

My understanding of your tests under https://cdn.polyfill.io/test/director?feature=Event mean that you force that feature to be loaded anyway, so the tests pass but IE11 is not part of the browsers which the polyfill will load for.

IE Edge supports the `Event` constructor without the polyfill, hence why the fix is just adding `IE11` to the version support. :smile:

triblondon:
Ah, gotcha, yes when test suite is run in targeted mode we skip tests for browsers that we don't think we have a polyfill for. But our creature grid already shows ie11 is polyfillable, meaning the tests fail without the polyfill and pass with it, so you're right we just need to correct the config.

PR #483, opened 2015-09-23
fix: Function.prototype.bind was failing on new BoundFn()

vvo:
- add a new test
- use the standard es5-shim polyfill without any changes BUT include
  the necessary helpers/variables

This error could be spotted by using ESLint for example (undeclared
variable).


vvo:
cc @triblondon let me know what you think, this is keeping me from using polyfill-service for now :)

vvo:
(ran the tests locally and it was fine).

Also I have seen projects like ember making their sauce keys public for ease of collaboration, what do you think?

thanks

vvo:
@triblondon @jonathantneal @samgiles would love to have your input on this one, thanks!

vvo:
I see you released 2.0.0 4 days ago, let me know if you need anything from this PR.

Current bind implementation is **broken**

PR #485, opened 2015-09-29
fix: Function.prototype.bind was failing on new BoundFn() (local for #483)

triblondon:
See #483


vvo:
Awesome

vvo:
Android failed in one check, can you relaunch? I bet saucelabs had issues here

triblondon:
Restarted.

Sauce has been really flakey for ages.  I do have an open source account from Browserstack that they're keen for us to use but I haven't found time yet to port the test runner to work with browserstack.

vvo:
Same issues on my side with saucelabs but wondering if browserstack would really solve them. Maybe remotely starting and controlling browsers over the internet is something that will fail no matter what :D

triblondon:
OK, we're good here.

vvo:
Are you planning to release 2.1 soon? Thanks

PR #486, opened 2015-09-30
Make window.hasOwnProperty work in IE8.

ivoreis:
This fixes hasOwnProperty IE8 error if you try to query window object. window.hasOwnProperty is undefined otherwise.. 

Uses the same fix used in for Object.assign (929ce80270c2ed886b09a1b25242551b3ff2f461)

Relates with: #433

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "polyfill-service",
-  "version": "2.0.0",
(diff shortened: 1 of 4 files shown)


triblondon:
@ivoreis thanks for this - could you add a test for it as well?

ivoreis:
Added tests for window object. Tested locally and on browserstack.

Issue #501, opened 2015-11-01
Chrome 31 needs promise polyfill


jhiswin:
Chrome 31 doesn't support native Promise, and needs to be polyfilled.  It was added in Chrome 32.

*tested on BrowserStack and using 360safe browser in webkit mode (which is a clone of Chrome 31)

jhiswin:
On an additional note, V8's native promise is known to be buggy:
https://code.google.com/p/v8/issues/detail?id=4162
And most recommend to just polyfill it regardless.  core-js, for example, always polyfills on Chrome.

jhiswin:
After some more testing, it seems as if Chrome never gets polyfill Promise.  Really odd.

triblondon:
Yes it does:
http://cdn.polyfill.io/v2/polyfill.js?ua=chrome/31&features=Promise

It's just not in the default set, because it's big.  See the docs to find out which polyfills are served by default.

jhiswin:
@triblondon ah so it was deliberate because of size issues; I just kind of assumed Promise would be included by default

PR #502, opened 2015-11-01
Chrome 31 doesn't have native Promise yet

jhiswin:
Just need to change to Chrome 32 as the endpoint #501

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "polyfill-service",
-  "version": "2.0.0",
+  "version": "2.1.0-beta.1",
   "dependencies": {
     "babel-core": {
       "version": "4.7.16",
--- a/package.json
+++ b/package.json
@@ -74,5 +74,5 @@
     "sauce-tunnel": "^2.2.3",
     "wd": "^0.3.8"
   },
-  "version": "2.0.0"
+  "version": "2.1.0-beta.1"
 }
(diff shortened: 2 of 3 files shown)


triblondon:
I think we're right here already actually.  The config says we serve the polyfill to `* - 31`, and MDN says Promises became native in 32.  The numbers are inclusive - 31 in our config means everything < 32, following standard semver conventions.

jhiswin:
@triblondon the problem is that when testing with Chrome 31, the polyfill service does not serve promises.
BTW you guys shuold apply for a free browserstack account.  It's free for open source project devs.

jhiswin:
This is what I get from the polyfill service using Chrome 31 on Windows 7 (on BrowserStack):

/\* Rerun without minification for verbose metadata */

(function(undefined) {Element.prototype.matches=Element.prototype.matches||Element.prototype.webkitMatchesSelector;Element.prototype.closest=function(t){for(var e=this;e;){if(e.matches(t))return e;e=e.parentElement}return null};Object.assign=function(r,t){for(var n,e,a=1;a<arguments.length;++a){e=arguments[a];for(n in e)Object.prototype.hasOwnProperty.call(e,n)&&(r[n]=e[n])}return r};String.prototype.includes=function(t,e){if("object"==typeof t&&t instanceof RegExp)throw new TypeError("First argument to String.prototype.includes must not be a regular expression");return-1!==this.indexOf(t,e)};}).call('object' === typeof window && window || 'object' === typeof self && self || 'object' === typeof global && global || {});

triblondon:
See #501.  Re browserstack, I have been thinking of moving over from Sauce to Browserstack but I don\t know anything about their platform.  If you'd like to submit a PR to add a `browserstack.js` task to `tasks/` that would be super appreciated.  The current CI task is sauce.js

Issue #498, opened 2015-10-20
Opera Mini support


kdzwinel:
Is Opera Mini supported? We use polyfill.io but I'm getting `localStorage` errors (which this browser does not support).

triblondon:
Yes, we do identify opera mini as `op_mini`.  But I suspect the targeting of our polyfills is very incomplete.  Please do add the additional metadata that is needed to target the right polyfills at Opera Mini.

triblondon:
Sent a quick note to Bruce Lawson.

kdzwinel:
I'm not sure what metadata you are referring to, could you elaborate on that?

This is my minimal test: http://jsbin.com/rufasuc.

It fails (localStorage doesn't work) even thought polyfill.io script is included. Tested on Opera Mini Emulator (v8).

<img width="279" alt="opera mini localStorage" src="https://cloud.githubusercontent.com/assets/985504/10866149/61ce47ee-8023-11e5-9e23-3ccadabfb1b0.png">

triblondon:
We only target localstorage at IE currently:

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/localStorage/config.json

I don't know if our LS polyfill is effective in Opera Mini, or what it would achieve. I would generally advocate not using JS at all in proxy browsers like Opera Mini.  But if the polyfill does work, feel free to do a PR to add opera mini targeting.

Contribution guidelines:
http://cdn.polyfill.io/v2/docs/contributing

brucelawson:
Opera Mini doesn't support localStorage at the moment - because there is no "local"; everything is rendered on the server.

kdzwinel:
Right, the question is if there is a way to pollyfil it on Opera Mini, e.g. using cookies. Or, maybe, load existing polyfill, so that a dummy localStorage object is created. It won't persist state, but it will prevent page from failing when some script tries accessing `localStorage`.

brucelawson:
cc  @mathiasbynens

mathiasbynens:
The polyfill seems to load just fine in Opera Mini and creates the global `localStorage` object correctly.

Let’s enable the polyfill in Opera Mini, as that prevents pages from failing in the case @kdzwinel mentioned. I’ll submit a PR.

triblondon:
Closing this since we do support Opera Mini so that question seems well answered.

PR #503, opened 2015-11-02
Enable `localStorage` polyfill in Opera Mini

mathiasbynens:
#498.

### DIFF
--- a/polyfills/localStorage/config.json
+++ b/polyfills/localStorage/config.json
@@ -4,7 +4,8 @@
 		"modernizr:localstorage"
 	],
 	"browsers": {
-		"ie": "6 - 7"
+		"ie": "6 - 7",
+		"op_mini": "*"
 	},
 	"dependencies": [
 		"Window"


triblondon:
Thanks @mathiasbynens.  Can I clarify what versions mean in relation to mini?  Is everyone on the latest version regardless of the age of their client, or is the client version used to determine which version of the server component to run?

So in practical terms, what does it mean to say 'users of Opera mini 5'?

mathiasbynens:
cc @brucelawson

brucelawson:
finding out definitively from the Mini P.M., will report back.

brucelawson:
Apologies for the delay, but I wanted to be absolutely sure, so asked Opera Mini PM and engineers.

All Opera Mini clients use the same server version and that
also means they all use the same Presto version. JavaScript and CSS
support is the same for all clients, with the exception that user may
enable Small Screen Rendering which removes a lot of CSS rules. They
can also use different viewport settings, but the same holds for
normal browsers. Version numbers effectively refer to different
versions of the client UI and capabilities (whether it supports
'night mode', private tabs, progressive download) etc.

brucelawson:
Added a note to https://dev.opera.com/articles/making-sites-work-opera-mini/ prompted by this Q. Addition reads:

Because Opera Mini renders on our servers, all Opera Mini clients use the same server version, so all use the same Presto version. JavaScript and CSS support is therefore identical on all clients, with the exception that user may enable Small Screen Rendering which removes a lot of CSS rules.

Version numbers refer to different versions of the client UI and capabilities (whether it supports night mode, private tabs, progressive loading, etc.)

Opera Mini 4.x clients load the whole transcoded page at once, as they have no progressive loading. This slightly reduces the time allowed for JavaScript execution (including XHRs). With newer clients, the transcoder is more generous, since it is able to send the compressed page to the client in chunks.

triblondon:
OK, I think I've gotten my head around this:
- When I request a polyfill bundle from the service through Opera Mini on a generic fruit based telecommunications device (model 6), I get identified as `op_mini/12.0.0`
- According to the [Opera Mini request headers doc](https://dev.opera.com/articles/opera-mini-request-headers/), this appears to be parsing correctly
- The last server upgrade was in March 2015, and all Opera Mini users will be on the same version
- This PR targets Opera Mini up to v11, which no-one will be using

So it seems to make sense to me for all our Opera Mini targets to be `*`, unless the polyfill is for something related to the core rendering engine, in which case we can set it to the presto version in which that feature is introduced and then we should automatically stop serving the polyfill once the Big Machine is updated to a version of Presto greater than that.

But for localStorage, it's presumably not in the rendering engine but in your JS engine, which I imagine is very custom for Mini and not comparable with a desktop version series.

So, TLDR: this should probably be `*`, not `5 - 11`.  WDYT?

triblondon:
Paging @brucelawson and @mathiasbynens see update above (soz for not mentioning you in it)

mathiasbynens:
+1 to using `*` instead of `5 - 11`.

triblondon:
@mathiasbynens can you change it then?  This PR in in your fork.

mathiasbynens:
Done.

PR #506, opened 2015-11-05
Add robots.txt to avoid indexing API responses

tom-parker:
@triblondon I'm not sure if the regex is correct for the robots middleware


service/index.js
```diff
@@ -119,7 +124,7 @@ app.get(/^\/v([12])\/polyfill(\.\w+)(\.\w+)?/, function(req, res) {
 	var apiVersion = parseInt(req.params[0], 10);
 	var firstParameter = req.params[1].toLowerCase();
 	var minified =  firstParameter === '.min';
-	var fileExtension = minified ? req.params[2].toLowerCase() : firstParameter;
+	var fileExtension = (minified && req.params[2]) ? req.params[2].toLowerCase() : firstParameter;
```
triblondon: do we even need `minified` here?  The check for `req.params[2]` seems to do the job.

service/index.js
```diff
@@ -20,6 +20,11 @@ var one_week = one_day * 7;
 var one_year = one_day * 365;
 var contentTypes = {".js": 'application/javascript', ".css": 'text/css'};
 
+app.get(/^\/v([12])\/robots\.txt$/, function (req, res) {
+    res.type('text/plain');
+    res.send("User-agent: *\nDisallow: /");
```
triblondon: This will disallow documentation pages, which we want to allow to be scraped.  Also I'm not totally convinced that deploying robots.txt in a subfolder will work for all robots.  I'd suggest publishing it at the root but specifying the specific path patterns that we want to exclude.  Can you do path patterns in robots?

tom-parker:
Yep, robots needs to be at the root. This should allow `/v1/docs/*` and `/v2/docs/*`. I had a look at Jake's suggested module but it doesn't seem to cater for the use of `Allow` which I think is the easiest way to handle all different routes. Regex doesn't seem that widely supported by robots either so unfortunately we can't use a pattern to match the version number rather than hard coding.

PR #510, opened 2015-11-06
Update versions from String.prototype.repeat

hannesj:
As per https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/repeat


triblondon:
Thanks!

triblondon:
Not waiting for CI: None of our CI tests are granular enough with browser versions to be affected by this change.

PR #515, opened 2015-11-11
Corrected indexOf casing in matchMedia polyfill

donatasbacius:


doodlemoonch:
+1

triblondon:
Ooh, thanks.  Can you add a test maybe for both the add and remove listeners?

triblondon:
@donatasbacius doodlemoonch if one of you could sort out a couple of tests here that would be appreciated.

triblondon:
Seems pointless to keep this hanging around.

PR #529, opened 2015-11-30
Add simple memory cache around the latest polyfill collection

triblondon:
Hopefully to reduce the amount of time we spend blocked on IO, though it's hard to tell why async IO functions are causing such long event loop delay.

### DIFF
--- a/lib/sources.js
+++ b/lib/sources.js
@@ -33,6 +33,7 @@ try {
 
 function Collection(version) {
 	this.version = version || 'latest';
+	this.cache = {};
 	if (this.version !== 'latest') {
 		this.version = semver.maxSatisfying(semverVersions, version);
 	}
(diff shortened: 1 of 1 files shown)


lib/sources.js
```diff
@@ -51,9 +52,16 @@ Collection.prototype.getPolyfill = function(featureName) {
...
+		}
 		return readFile(path.join(polyfillsPath, this.version, featureName+'.json'), 'utf-8').then(function(str) {
-			return JSON.parse(str);
-		});
+			var feature = JSON.parse(str);
+			if (this.version === 'latest') {
+				this.cache[featureName] = feature;
+			}
+			return feature;
+		}.bind(this));
```
JakeChampion, line 64: We are using node 5 now, can make use of arrow functions for binding this lexically now

triblondon, line 64: It's just for consistency.  Will change all lexically bound functions to arrows at some point.

PR #531, opened 2015-12-02
Add SVG4Everybody to polyfill service

jonathan-fielding:
Hi

I have added the svg4everybody polyfill for handling external svg sprites in unsupported browsers

Thanks

Jonathan


jonathan-fielding:
Just realised detect wont work when fired after the polyfill is in place as it isn't a feature detection, please advise what I should do

JakeChampion:
Is there a way to detect this via feature detection rather than UA sniffing?

patrickkettner:
@JakeChampion not 100%, no. @jdalton did a good swing at a feature detect [here](https://github.com/jonathantneal/svg4everybody/issues/71), but safari is still an issue

triblondon:
@jonathan-fielding I haven't forgotten about this but if there's no viable feature detect I don't think we can have it in the service.  What do you want to do here?

JakeChampion:
I think we could have it in the service but would have to warn user's that it doesn't come with a detect.
As long as the config.json has a restricted browser set it should be okay.

triblondon:
OK, let me look at this a bit more closely then.
1. The polyfill puts two very generically named functions into the parent scope - `embed` and `loadreadystatechange`.  Other polyfills could easily define similarly named functions.  The polyfill should define these inside of a closure and it actually need not export anything.
2. It uses a very tight loop in browsers that do not support requestAnimationFrame, firing timer events every 17ms.  We would never want a polyfill that does timer stuff that aggressively to be in the default set, and at the very least it needs a warning note in the config.json.
3. The name is wrong, because there's no JavaScript global called `SVGExternalContent`.  You could call it `~svg-external-content` which is the convention for polyfills that target features without an API.  The name does seem descriptive enough.
4. I'm still bothered by the inability to detect.  People who use the `always` flag because they want to bundle the polyfills with their code will end up with a pointless tight raf loop in their code, and maybe it would even disrupt the native implementation, not sure.

On balance I think (2) and (4) mean I'm unhappy to accept this polyfill in its current form. Maybe @jonathantneal would consider revising it a bit.

PR #534, opened 2015-12-03
HTMLPictureElement polyfill in IE8

ironsidevsquincy:


triblondon:
This is covered by CI - need a local test branch.

ironsidevsquincy:
Not sure I understand what that means @triblondon; do you need anything from me?

triblondon:
Nope, just a note-to-self

triblondon:
Ah, sorry Darren, didn't realise you had already done this in a local branch.

There is a test failure in IE8:

```
Failed tests:
 - ie 8 (Sauce results: https://saucelabs.com/tests/7189cee7ebe046ec8e698450ccdc6a0f)
    -> HTMLPictureElement
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=HTMLPictureElement
Browsers tested with no failures: chrome/46, chrome/35, firefox/41, firefox/30, ie/12, ie/11, ie/10, ie/9, ie/7, ie/6, safari/9, safari/8, android/4.4
```

image

triblondon:
Ping @ironsidevsquincy

ironsidevsquincy:
@triblondon, not sure what to do with this. We're forcing the polyfill on IE8 in production and everything's working fine

Is there a readme around what the tests are doing,how they're running, etc?

triblondon:
There's a guide to testing in the contrib guide: http://cdn.polyfill.io/v2/docs/contributing#running-tests

Could you try running the picture tests on IE8?

ironsidevsquincy:
Ok, digging around it looks like it uses the same detect.js for determining whether to polyfill, and whether the test passes.

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/HTMLPictureElement/detect.js

Problem is, this isn't a direct polyfill of a HTMLPictureElement, rather it just emulates it's functionality. Really, the detect should be something like `!!window.HTMLPictureElement` (as used in the [polyfill itself](https://github.com/scottjehl/picturefill/blob/master/dist/picturefill.js#L962)), and the test should test an image is being responsive

Not sure where to go with this. Is this an issue that's arisen elsewhere before?

triblondon:
Here's the issue, we discussed it a bit here:
https://github.com/Financial-Times/polyfill-service/pull/448

And then we quizzed Scott on twitter:
https://twitter.com/JonthanFielding/status/627121322989121536

In summary, the polyfill does not populate `HTMLPictureElement`, but does add `srcset` and `sizes` to the `<source>` element, so that detect should be good for both the native and the polyfilled variants.  The fact that it's failing suggests that maybe the polyfill isn't working completely in IE8.  @ironsidevsquincy why are we trying to use this in IE8 anyway, doesn't Next FT start JS support at IE9?

ironsidevsquincy:
Hmmm, interesting. I wonder if it's because IE8 doesn't natively understand the `source` element. It's definitely working as we have images on IE8 on the front page. 

We pull in the Picturefill for non-ctm browsers - https://github.com/Financial-Times/next-js-setup/blob/master/templates/bootstrapper.html#L24

triblondon:
So it seems our only option is to chuck in the proprietary API and make the detect something like:

```
'HTMLPictureElement' in this || 'picturefill' in this
```

PR #543, opened 2015-12-14
RequestAnimationFrame webkit polyfill ios_saf 6

laurieboyes:
Allow webkit polyfill for requestAnimationFrame on all iOS6 safari browsers, as opposed to just versions greater than 6.1

### DIFF
--- a/polyfills/requestAnimationFrame/config.json
+++ b/polyfills/requestAnimationFrame/config.json
@@ -29,7 +29,7 @@
 			"browsers": {
 				"chrome": "10 - 23",
 				"safari": "6",
-				"ios_saf": "6.1"
+				"ios_saf": "6"
 			},
 			"dependencies": [
 				"performance.now"


triblondon:
Hmm, according to caniuse iOS added support for raf in 6.1, so this should be `<= 6.0`.  What tests have you done?

matthew-andrews:
@triblondon this is the polyfill for the prefixed one — which is only needed in iOS 6.1

matthew-andrews:
screen shot 2015-12-15 at 13 40 42

I believe this only applies to the `webkit` prefixed polyfill — which for Safari applies to 6.0 and 6.1 only

triblondon:
OK, understood.

What testing have you done?

matthew-andrews:
@laurieboyes has been coordinating this and i believe he's off today.  i suspect this can wait until he's back.

laurieboyes:
Cheers for looking at this. I couldn't spot anywhere within the current testing that I could slot in an automated test for this but please let me know if I'm missing anything.

Manual testing so far has involved looking at the polyfill.js file on an iOS6 device and noting that raf was indeed being alias polyfilled. I was putting faith in the fact that the change was so minor but this degree of testing seems fairly pitiful now that I see it written down so will do more thorough testing and report back.

laurieboyes:
**Test results**

/test/director?feature=requestAnimationFrame&mode=targeted

Carried on on iPad2 with Safari iOS 6.0

Before:
image 3

After
image 4

triblondon:
:+1:

PR #544, opened 2015-12-16
Give Array.prototype.find to BB devices

adgad:
/cc @triblondon

### DIFF
--- a/polyfills/Array/prototype/find/config.json
+++ b/polyfills/Array/prototype/find/config.json
@@ -4,6 +4,7 @@
 		"modernizr:es6array"
 	],
 	"browsers": {
+		"bb": "*",
 		"chrome": "*",
 		"firefox": "3.6 - 24",
 		"ie": "*",


adgad:
Tested by:

http://jsbin.com/jevojayapa on a BB10 (Blackberry Passport) device shows Array.prototype.find is undefined

http://qa.polyfill.io/v2/polyfill.js?ua=Mozilla/5.0%20(BB10;%20Touch)%20AppleWebKit/537.35+%20(KHTML,%20like%20Gecko)%20Version/10.3.1.2243%20Mobile%20Safari/537.35+&features=Array.prototype.find doesn't load Array.prototype.find polyfill

Local with changes:

screen shot 2015-12-16 at 16 36 23

Tested Next homepage (previously Markets Data, which uses Array.prototype.find was broken - with local polyfill service running it works).

triblondon:
Thanks

PR #548, opened 2015-12-17
HTMLPictureElement for BB7-10

ironsidevsquincy:
http://caniuse.com/#feat=picture


triblondon:
Does it work in those browsers?

ironsidevsquincy:
Yes, tested the polyfill's examples page (http://scottjehl.github.io/picturefill/#examples) in BB7&10

triblondon:
OK, fine.

PR #555, opened 2015-12-19
Array.of updated browser support for chrome 45

christophehurpeau:
According to [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/of) Array.of is available in chrome since version 45


triblondon:
:+1:

PR #560, opened 2015-12-26
Remove support for v1 API

triblondon:
Removes support for `libVersion`, removes concept of historic polyfill collections.


triblondon:
Full test suite passes.  Still need to sort out Travis.

JakeChampion:
LGTM :+1: 

BTW I've got basic test suite running on browserstack, I'll be carrying that on at some later point

PR #563, opened 2016-01-04
Changed version resolution from lte to lt

seanmtracey:
Seems that <= didn't work, but < does work.

### DIFF
--- a/polyfills/Array/prototype/find/config.json
+++ b/polyfills/Array/prototype/find/config.json
@@ -5,7 +5,7 @@
 	],
 	"browsers": {
 		"bb": "*",
-		"chrome": "<=44",
+		"chrome": "<45",
 		"firefox": "3.6 - 24",
 		"ie": "*",
 		"ie_mob": "10 - *",


triblondon:
I think this bug is papering over a different problem.  The specifier as it exists already should match 44:

image

I agree it doesn't, though:

http://cdn.polyfill.io/v2/polyfill.js?features=Array.prototype.find&ua=chrome/44

```
/* For detailed credits and licence information see http://github.com/financial-times/polyfill-service.
 * 
 * UA detected: chrome/44.0.0
 * Features requested: Array.prototype.find
 *  */
```

So this needs a bit more investigation, probably [starting here](https://github.com/Financial-Times/polyfill-service/blob/master/lib/index.js#L67).

JakeChampion:
Would it be because Chrome has version numbers such as 44.0.2526.106, which is bigger than 44?

triblondon:
Looks like there was a bug in the version of the semver parser that we were using.  I updated it and it looks good now.

JakeChampion:
Should we add this as a test case?

Issue #547, opened 2015-12-17
`grunt build` fails on master branch when using Windows


Ikkyu-os:
I failed to install polyfill-service locally using npm install.

Basically, the script will start grunt build (as the doc' specifies), but it seems that most the files that should be in __dist\latest directory are missing.

Extract of log here: 
Running "buildsources" task
Loading sources...
{ [Error: ENOENT, no such file or directory 'D:\git_clones\openSource\polyfill-service\polyfills__dist\latest\Array\from.json']
  errno: -4058,
  code: 'ENOENT',
  path: 'D:\git_clones\openSource\polyfill-service\polyfills\__dist\latest\Array\from.json',
  syscall: 'open' }

I attached the full result of a grunt build in the issue.

[build.txt](https://github.com/Financial-Times/polyfill-service/files/65252/build.txt)

Could it be due to the fact that master CI build is failing on travis ? 
https://travis-ci.org/Financial-Times/polyfill-service

triblondon:
Travis is broken due to issues tunnelling to Sauce Labs, unfortunately.

triblondon:
I'll look into the build problem that you're having though

triblondon:
Hm, I've never tried building on Windows, so maybe there's a path issue that doesn't work with Windows' filesystem.  I just confirmed that a fresh checkout of master builds fine for me.

triblondon:
Can anyone reproduce this problem?

Ikkyu-os:
Is there any way to add more logs to grunt so I could perhaps investigate which steps is actually failing and fix it ? 

Basically, it seems that \polyfills__dist\latest  is almost empty on my local. where all __dist\v1.0.X\ directories are properly populated.
Which process create those directories ?

triblondon:
Take a look at the gruntfile for the build process.  Your problem is probably somewhere in the installcollections task, or possibly in buildsources.  Adding logs is just a matter of dropping grunt.log.writeln wherever you want to log.

Ikkyu-os:
So i was able to investigate the issue and here's my findings.

When buildsources task is executed, it seems that polyfill subdirectories for "latest" version is not created, generating an error.

Attaching some logs (sorry for the poor debug trace skills)

``` bash
Running "buildsources" task
Loading sources...
debug - version = latest
debug - destFolder to create : D:\git_clones\openSource\polyfill-service\polyfills\__dist\latest
[...]

PR #565, opened 2016-01-05
grunt task buildsources -  fix for missing subdirectories

Ikkyu-os:
Link to issue https://github.com/Financial-Times/polyfill-service/issues/547

Create polyfill subdirectories prior to create polyfill json file.

### DIFF
--- a/tasks/buildsources.js
+++ b/tasks/buildsources.js
@@ -107,8 +107,17 @@ module.exports = function(grunt) {
 				config.rawSource = '\n// '+featureName + '\n' + config.rawSource;
 
 				var featurePath = path.join(destFolder, featureName+'.json');
+        var featureDir = path.dirname(featurePath);
+        mkdir(featureDir, function(err) {
+          if (!err) {
+            fs.writeFileSync(featurePath, JSON.stringify(config));
+          }
+          else {
+            grunt.log.writenln("Can't create directory " + featureDir + "due to the following error: " + err);
+          }    
+        });                    
 				fs.writeFileSync(featurePath, JSON.stringify(config));
-				grunt.log.writeln('+ '+featureName);
+				grunt.log.writeln('+ '+featurePath);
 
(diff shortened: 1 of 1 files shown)


tasks/buildsources.js
```diff
@@ -107,6 +107,15 @@ module.exports = function(grunt) {
 				config.rawSource = '\n// '+featureName + '\n' + config.rawSource;
 
 				var featurePath = path.join(destFolder, featureName+'.json');
+                var featureDir = path.dirname(featurePath);
+                mkdir(featureDir, function(err) {
```
JakeChampion: Could you change this to be a synchronous action instead of asynchronous? The reason being that this introduces a race condition where the directory creation happens after the file creation on line 199 in this file and an error would be thrown.

Ikkyu-os: Hi :)
I actually used the sync version. Check line 17

```
var mkdir = require('mkdirp').sync;
```

Documentation about the package: https://github.com/substack/node-mkdirp

Perhaps I should keep a mkdir variable to get the dependency only, and then explicitly call mkdirp.sync(dir, opts).
What do you think ?

JakeChampion: Ah, it's good as it is then. I assumed async as a callback was being passed in to the function.

tasks/buildsources.js
```diff
@@ -107,6 +107,15 @@ module.exports = function(grunt) {
 				config.rawSource = '\n// '+featureName + '\n' + config.rawSource;
 
 				var featurePath = path.join(destFolder, featureName+'.json');
+                var featureDir = path.dirname(featurePath);
```
JakeChampion: The indentation doesn't seem correct, would you be able o change it from spaces to tabs?

Ikkyu-os:
updates done with comments from @JakeChampion

JakeChampion:
LGTM -- @triblondon happy for this to go in?

PR #568, opened 2016-01-06
[WIP] commonjs build task for polyfills

JakeChampion:


tasks/buildsources.js
```diff
@@ -16,8 +30,8 @@ module.exports = function(grunt) {
 	var babel = require("babel-core");
 	var mkdir = require('mkdirp').sync;
 
-	grunt.registerTask('buildsources', 'Build polyfill sources', function() {
-
+	grunt.registerTask('buildsources', 'Build polyfill sources', function(done) {
+		var done = this.async();
```
maxnordlund, line 34: Why shadow the `done` parameter here? Maybe just rename the parameter to `_done` instead. I also think this causes deoptimization, but is probably negligible in this case.

JakeChampion, line 34: The shadowing was by accident, the code has since been updated :-)

maxnordlund, line 34: No problem, always happy to help.

JakeChampion:
I ran `npm test` locally and it passed.

JakeChampion:
Superseded by https://github.com/Financial-Times/polyfill-service/pull/585

PR #557, opened 2015-12-23
Add polyfill for console

whitneyit:
This PR adds a poly fill for `console` and its various methods. RE: #440 

Whilst putting the polyfill together I was unable to come up with a definite list of methods that each browsers support and in which version they support which method :cry:

Also, I'm not entirely sure how we detect each missing method. Would splitting each method into its own folder similar to [Array](https://github.com/Financial-Times/polyfill-service/tree/master/polyfills/Array) work better?

PR #569, opened 2016-01-07
Console (local for #557)

triblondon:
See #557


triblondon:
Branch passes CI from my machine.  Travis still not working.

whitneyit:
Travis just needs to be updated with the sauce enviornment variables to get it to work. Also, if you have a look at #557 I rebased it to handle more cases. I think you should have a look at it

PR #570, opened 2016-01-07
Add individual console methods

JakeChampion:
This is pulled from https://github.com/Financial-Times/polyfill-service/pull/557


JakeChampion:
It's been tested locally by @whitneyit and [all passed](https://gist.github.com/whitneyit/87931294df90e02c5db3)

Update: I've also ran the suite of tests `grunt ci` and it all passed. Not sure why this works locally  and not on Travis.

triblondon:
This still feels a bit repetitive.  This pattern, for example:

```
this.console.info = function info() {

    var fn = console.log;
    var hasApply = false;

    try {
        hasApply = !!fn.apply;
    } catch (e) { /* do nothing */ }

    if (hasApply) {
        return function () {
            fn.apply(console, arguments);
        };
    }

    return function (arg1, arg2, arg3) {
        fn(arg1, arg2, arg3);
    };
};
```

appears several times.  Could we shorten it to:

```
this.console.info = function info() {

    var fn = console.log;

    try {
        !!fn.apply &&
        return function () {
            fn.apply(console, arguments);
        };
    } catch (e) { /* do nothing */ }

    return function (arg1, arg2, arg3) {
        return fn(arg1, arg2, arg3);
    };
};
```

or even:

```
this.console.info = this.console.log;
```

triblondon:
@JakeChampion are you done with this one?

JakeChampion:
@triblondon Yup, seems to have conflicts now though.

triblondon:
@JakeChampion lots of issues here that needed thinking about:
- Several methods included were not actually part of any console standard.  I removed these
- Some methods require arguments, and (correctly) throw exceptions in implementing browsers if they are not supplied, which was failing tests
- The root console polyfill had a dependency on itself, which created a circular dependency graph.  I removed that and am creating a check for this in the buildsources task.

The inclusion of console polyfills still crashes iOS on Sauce, though it now passes locally on my iPhone.  I'm yet to dig into the Sauce situation because I can't do it offline, but meantime, comments welcome.

JakeChampion:
Good catch on the non standard methods. It's a bit annoying that the spec lives in a markdown document on Github and not within a standards body such as ECMA, W3C or WHATWG.

Most of the methods here are proposed methods, which I think is fine to go into polyfill-service so long as have a plan of action if they are not accepted as part of a standard.

The defacto standard methods are `console.debug`, `console.info`, `console.warn`, `console.error`, `console.log`, `console.dir` and `console.assert`.

I'm in favour of merging this PR.

triblondon:
OK, found the problematic polyfill on Sauce - console.profile crashes the emulator.  Works fine on a physical device. I've excluded those polyfills from CI.

PR #574, opened 2016-01-12
Add picturefill on android

wheresrhys:


wheresrhys:
http://caniuse.com/#feat=picture says it's supported in v46, but given it jumps from v 4.4.4 to 46, I don't really trust that table so going with *

triblondon:
That's the real version sequence.  After 4.4.4, they started using chrome version numbers.  So you do need <46.

PR #582, opened 2016-01-26
Polyfill findIndex for BB

ironsidevsquincy:


ironsidevsquincy:
https://github.com/Financial-Times/polyfill-service/pull/583 won

PR #584, opened 2016-01-27
Element.prototype.placeholder not displaying - fixes #566

### LINKED ISSUE #566: IE9 Placeholder Support: Larger Question around getting computed styles
Hi all,
I've been looking into what's stopping the placeholder polyfill from working in IE9 https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Element/prototype/placeholder/polyfill.js and all I've been able to find is the oddities around getCurrentStyle, getComputedStyle, and whether the browser supports camel-case props. Before looking into a fix for this specific polyfill, I wanted to reach out to see if there's a generic utility around this at the moment in the lib. Thanks!

[This](https://groups.google.com/d/msg/comp.lang.javascript/71ug4NBYJog/xXwUC9sgpP8J) seems to be a decent workaround, but think we could automate the camel-casing if need-be.

deddu:
- document.createElement was failing due to the '-' at the start of the identifier.
- missing getter.

**todo:** the styles aren't applying as they should.

### DIFF
--- a/polyfills/Element/prototype/placeholder/polyfill.js
+++ b/polyfills/Element/prototype/placeholder/polyfill.js
@@ -1,6 +1,7 @@
 Object.defineProperty(Element.prototype, 'placeholder', {
-	get:function(){
-		return this.getAttribute('placeholder') ;},
+	get: function() {
+		return this.getAttribute('placeholder');
+	},
 
 	set: function (value) {
 		if (!value || !/^(input|textarea)$/i.test(this.nodeName) || !/^(email|number|password|search|tel|text|url|)$/i.test(this.getAttribute('type'))) {
--- a/polyfills/Element/prototype/placeholder/polyfill.js
+++ b/polyfills/Element/prototype/placeholder/polyfill.js
@@ -1,4 +1,7 @@
 Object.defineProperty(Element.prototype, 'placeholder', {
+	get:function(){
+		return this.getAttribute('placeholder') ;},
+
 	set: function (value) {
 		if (!value || !/^(input|textarea)$/i.test(this.nodeName) || !/^(email|number|password|search|tel|text|url|)$/i.test(this.getAttribute('type'))) {
 			return;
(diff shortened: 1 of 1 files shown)


deddu:
ehrm.. placeholder not displaying. sorry.

triblondon:
Any way we can test this in mocha?  I'm struggling to see how, the result being primarily visual...

deddu:
@triblondon I honestly don't have any clue how. I am afraid that it will need a regression/integration test instead... At the end the possible problems will arise when dealing with IE's quirks. Do you have anything set up for that, that I can work on?

polyfills/Element/prototype/placeholder/polyfill.js
```diff
@@ -1,4 +1,7 @@
 Object.defineProperty(Element.prototype, 'placeholder', {
+	get:function(){
+		return this.getAttribute('placeholder') ;},
```
triblondon: Line break before `}`, whitespace after `function()`.

We have linting rules in the root of the project.  Can you run them?

triblondon:
OK, fair enough, I can't think of a good test mechanism that doesn't require quite a substantial investment of time.  Could you just post a JSBin/JSfiddle/Codepen with a test case and I'll verify it on IE8 and 9 VMs I have here?

Also a couple of linting failures to address.

Thanks for the contribution - much appreciated!

deddu:
Thank you for the awesome service, @triblondon! 
[here's a quick and dirty fiddle](https:
//jsfiddle.net/j185kx2m/1/) from the html I used for testing. I'll attach the full file it for reference. Sadly ie9 doesn't play very well with fiddle,or codepen but either the attachment or the link should work.
[wutwut.html.zip](https://github.com/Financial-Times/polyfill-service/files/110311/wutwut.html.zip)
I will work on the lint as soon as I get a minute. Silly question from a lazy dev : there's a specific grunt task to run the linter? I was searching for a `grunt lint` but couldn't find it.

triblondon:
Linters are not loaded by default, but we do provide the .jshintrc file so if you have jshint in your editor it should pick up the settings and lint automatically

triblondon:
OK, test case works fine.  If you could fix the linting I'm happy to merge this.

deddu:
I know it will sound ridiculous, but I am having problems with the linting. Using the jshintrc file I have only errors due to undefined global variables, (i.e.: document, element). I don't see the need to fix those, they are part of the design. But I don't have any other warning or error. That said, I manually fixed the spacings and indentations, and updated the PR.

triblondon:
Hmm, that's odd.  I'll put that on the list to look into :-)

Meanwhile this is great, thanks!

PR #585, opened 2016-01-31
Symbol polyfill

JakeChampion:
| Polyfill | unminified size | minified size |
| --- | --- | --- |
| CoreJS | 18.6 KB | 12.6 KB |


JakeChampion:
@mehcode @triblondon - currently failing tests in ie 8/11/12 and android 4.4. 

```
 - ie 12 (Sauce results: https://saucelabs.com/tests/c281d470d7d04f7783b833d52c2f9b72)
    -> Symbol
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Symbol
 - ie 11 (Sauce results: https://saucelabs.com/tests/d4755eeddd1548b28294f3822050a246)
    -> Symbol
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Symbol
 - ie 8 (Sauce results: https://saucelabs.com/tests/6ddc723dbbe14429a6d2a34fb8891084)
    -> Symbol (timeout after 10000ms, 2 retries)
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Symbol (timeout after 10000ms, 2 retries)
 - android 4.4 (Sauce results: https://saucelabs.com/tests/9984af1685ac453da1b0aa913343a34b)
    -> Symbol
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Symbol
```

tasks/buildsources.js
```diff
@@ -40,50 +53,66 @@ module.exports = function(grunt) {
...
-				config.hasTests = fs.existsSync(path.join(polyfillPath, 'tests.js'));
+			if (config.licence) {
+				throw 'Incorrect spelling of license property in '+featureName;
+			}
 
-				if (config.licence) {
-					throw 'Incorrect spelling of license property in '+featureName;
-				}
+			if (config.build && config.build.commonjs === true) {
+				config.rawSource = streamToString(browserify().add(polyfillSourcePath).bundle());
```
zloirock: I do not exactly understand `polyfill-service` build process. Do you wanna bundle commonjs for each polyfill individually? In this case, you will have a problem, mentioned in your PR to `core-js`. Better run browserify for complete polyfills bundle if at least one polyfill requires commonjs.

JakeChampion: Each polyfill is stored as a separate json structure on the filesystem, when a browser requests a set of polyfills the service will combine the necessary files together and return them as the response. Because of this approach, we can only run computations such as browserify for each individual polyfill.

zloirock: CommonJS required for external modules, so filesystem path is not important.

JakeChampion: If you run `grunt buildsources` you will see the `/polyfills/__dist` directory fill up with json files corresponding to the polyfills themselves.

zloirock: I mean `browserify` should be called on the next step, for a bundle for a browser, not for files from this directory. Your approach will work for 1 polyfill from `core-js`, but if you wanna use no only one - you will have a bundle of size in megabytes because of duplication internal modules.

JakeChampion: Good point, I'll look into this, thanks for the feedback Denis :-) This will also help with ES Shim compliant modules which make use of helper modules.

triblondon:
@JakeChampion is this ready to review? If so can you talk me through it?  And do you want to target 3.4?

lib/index.js
```diff
@@ -159,7 +175,11 @@ function getPolyfillString(options) {
 					}
 				});
 				if (builtPolyfillString) {
-
+					return streamToString(browserify(intoStream(builtPolyfillString)).bundle());
```
JakeChampion: @zloirock Happier with this approach?

JakeChampion:
@triblondon This is ready for review

lib/index.js
```diff
@@ -159,7 +175,14 @@ function getPolyfillString(options) {
 					}
 				});
 				if (builtPolyfillString) {
-
+					return streamToString(
+						browserify(intoStream(builtPolyfillString))
```
JakeChampion: Browserify is done on the server for each request, the reason it is not done during build for each polyfill is to enable deduplication of common required modules.

zloirock: :+1:

service/index.js
```diff
@@ -168,7 +168,7 @@ app.get(/^\/v2\/polyfill(\.\w+)(\.\w+)?/, function(req, res) {
 	metrics.counter('hits').inc();
 	var respTimeTimer = metrics.timer('respTime').start();
 	var firstParameter = req.params[0].toLowerCase();
-	var minified =  firstParameter === '.min';
+	var minified = firstParameter === '.min';
```
JakeChampion: Removing an extra space

tasks/buildsources.js
```diff
@@ -161,4 +161,4 @@ module.exports = function(grunt) {
 		grunt.log.oklns('Sources built successfully');
 
 	});
-};
+};
```
JakeChampion, line 162: Adding a newline to end of file

polyfills/Symbol/tests.js
```diff
@@ -0,0 +1,222 @@
...
+	expect(Symbol.prototype.toString.call(Symbol('10'))).to.be('Symbol(10)');
+});
+
+
+it('should not allow implicit string coercion', function() {
+	var implicitStringCoercion = function() {
+		return Symbol('10') + '';
+	}
+	expect(implicitStringCoercion).to.throwError();
+})
```
JakeChampion: Missing a semi-colon.

polyfills/Symbol/detect.js
```diff
@@ -0,0 +1 @@
+'Symbol' in this && 'hasInstance' in this.Symbol
```
JakeChampion: Missing a semicolon

lib/index.js
```diff
@@ -159,7 +175,14 @@ function getPolyfillString(options) {
...
+						.transform({global: true}, 'uglifyify')
```
JakeChampion: `uglifyify` should probably only be used when requesting `.min.js`

lib/index.js
JakeChampion: I should probably not have this run on every requested bundle, instead only running it on ones where polyfills are written in commonjs or ES6 modules.

lib/index.js
JakeChampion: We could look to using [rollup](https://github.com/rollup/rollup) when requesting `.min.js`

lib/index.js
```diff
@@ -11,6 +14,19 @@ require('useragent/features');
 
 "use strict";
 
+function streamToString(stream) {
```
JakeChampion: A helper function, this is probably already out in npmjs-land somewhere.

lib/index.js
triblondon: Agree, we should only browserify if at least one polyfill has `build.commonjs` = true

JakeChampion:
> safari/8:  results received (163 passed, 12 failed), closing remote browser

I'll look into this ^

triblondon:
Where did you get to with this?

JakeChampion:
Not yet looked into the Safari issue, that will be this weekend

gheoan:
So this does not use `core-js` anymore?

polyfills/Object/getOwnPropertyDescriptor/detect.js
```diff
@@ -0,0 +1,11 @@
+'getOwnPropertyDescriptor' in Object && typeof Object.getOwnPropertyDescriptor === 'function' && (function doesGetOwnPropertyDescriptorWork(object) {
```
triblondon: Leaking a global here

triblondon:
Can't really see any benefit in the well-known symbols being defined separately to the Symbol polyfill.

JakeChampion:
Some Chrome versions ship Symbol but not the well-known Symbols

triblondon:
The polyfill itself is beyond me, but I've reviewed what I can understand and :+1:

PR #586, opened 2016-02-01
UserTiming polyfill

nicjansma:
@triblondon better late than never :)

This is my attempt at integrating my [usertiming.js polyfill](https://github.com/nicjansma/usertiming.js/) for [UserTiming](http://www.w3.org/TR/user-timing/).

Tests pass locally, but I'm not setup with SauceLabs to see if that passes as well.

The Contribution docs were slightly confusing, so if I've missed something, please let me know!

### DIFF
--- a/polyfills/UserTiming/config.json
+++ b/polyfills/UserTiming/config.json
@@ -14,5 +14,6 @@
 	},
 	"spec": "http://www.w3.org/TR/user-timing/",
 	"license": "MIT",
-	"repo": "https://github.com/nicjansma/usertiming.js"
+	"repo": "https://github.com/nicjansma/usertiming.js",
+	"docs": "https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API"
 }
--- a/polyfills/UserTiming/config.json
+++ b/polyfills/UserTiming/config.json
@@ -1,0 +1,18 @@
+{
+	"browsers": {
+		"android": "<=4",
+		"chrome": "<=24",
+		"firefox": "<=37",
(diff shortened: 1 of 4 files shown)


polyfills/UserTiming/detect.js
```diff
@@ -0,0 +1 @@
+'performance' in window && typeof window.performance.getEntriesByType === 'function' && typeof window.performance.mark === 'function'
```
triblondon: Prefer `this` to `window`, since we do not guarantee that `window` will not have been overridden.

polyfills/UserTiming/config.json
```diff
@@ -0,0 +1,18 @@
...
+		"ie_mob": "<=9",
+		"ios_chr": "*",
+		"ios_saf": "*",
+		"op_mob": "*",
+		"opera": "<=12",
+		"safari": "*"
+	},
+	"spec": "http://www.w3.org/TR/user-timing/",
+	"license": "MIT",
+	"repo": "https://github.com/nicjansma/usertiming.js"
```
triblondon: Can you add an appropriate `docs` link, eg https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API?

triblondon:
This is a very high quality PR. Thanks for taking so much care over our contrib guidelines, it makes my life a lot easier.

I've tested on Sauce, and locally on an older Opera version, and I'm happy.  Could you make the two small changes I've suggested?

nicjansma:
All set!  Thanks Andrew.

PR #588, opened 2016-02-01
Restrict setImmediate config

JakeChampion:
This patch will restrict setImmediate to the range of browser versions which don't have Promise natively implemented. This feature requires setImmediate in order to work.


triblondon:
Hmm, tricky. The browser targeting you've got here doesn't reflect support for setImmediate, it reflects the support for Promises which is currently the only polyfill in which we use setImmediate. This feels like a bad precedent to set, but if we're going to fix it by not including dependencies if the browser doesn't require the parent, then this could be OK as a temporary solution. But if it's fairly easy to fix the dep loading issue I'd rather jump directly to that.

tracker1:
I'm with @triblondon on this one... `setImmediate` may well be a desired polyfill outside of the use with Promise's polyfill...

http://caniuse.com/#feat=setimmediate

JakeChampion:
@triblondon I agree with finding a solution to the actual problem (loading polyfills which are not required) however I do believe the config for `setImmediate` is too open. For instance, Chrome Stable has `setImmediate`.

triblondon:
Sure, absolutely remove targeting from browsers that actually have the
feature natively.  But I don't think that's what the current PR does, is
it? Sorry if I've jumped to the wrong conclusion.

JakeChampion:
Updated the config to reflect the data in caniuse

PR #592, opened 2016-02-05
Return before trying to cast property to string

craigbilner:
I'm using the polyfill service with Babel on a Samsung Tablet rocking Chrome 38.

Babel is providing a `Symbol` which doesn't like being cast to a string, however, the browser does support `defineProperty` but is being pulled in as a dependency of `Promise`.

The actual fix for me is to remove `Promise|always`, but polyfills should always exit with the native option first anyway which would avoid the casting issue?

### DIFF
--- a/polyfills/Object/defineProperty/polyfill.js
+++ b/polyfills/Object/defineProperty/polyfill.js
@@ -6,6 +6,11 @@
 
 	Object.defineProperty = function defineProperty(object, property, descriptor) {
 
+		// Where native support exists, assume it
+		if (nativeDefineProperty && (object === window || object === document || object === Element.prototype || object instanceof Element)) {
+			return nativeDefineProperty(object, property, descriptor);
+		}
+
 		var propertyString = String(property);
 		var hasValueOrWritable = 'value' in descriptor || 'writable' in descriptor;
 		var getterType = 'get' in descriptor && typeof descriptor.get;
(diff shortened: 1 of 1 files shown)


JakeChampion:
Polyfills included from the polyfill-service can be requested with a gated option which will ensure they are only loaded if they do not already exist. E.G. https://cdn.polyfill.io/v2/polyfill.js?flags=gated

Here are the docs which describe the features available for the service -- https://cdn.polyfill.io/v2/docs/api

craigbilner:
thanks @JakeChampion, I'm aware of this but my issue was with `defineProperty`, which was pulled in as a dependency of `Promise`, so should `gated` gate all dependencies too? Also is it not best practice to just return at the earliest opportunity in a polyfill?

JakeChampion:
If a polyfill is incorrectly included it is a sign that the `config.json` file for the particular polyfill has an incorrect range set for the requesting browser. [I can see that Chrome is not included in the list of browsers which require the `defineProperty` polyfill](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object/defineProperty/config.json#L8).

The issue you're experiencing is becaed of the `|always` flag, that will pull down every polyfill that is required for `Promise` to load.

We don't deliver gated polyfills by default as the main aim of the polyfill-service is to servce each browser the polyfills they specifically need, which means that a gate isn't required.

craigbilner:
Thanks for the clarification @JakeChampion 

If the core philosophy of the polyfill service is to polyfill regardless, because it only provides the required polyfills, the `always` option contradicts this philosophy. Therefore, if the `always` option is used and pulls in dependencies, the only safe thing to do would be to gate these dependencies?

JakeChampion:
Yep, `&flags=always,gated` would enable that.

To bring the conversation back to the changes you've made to the code, the changes look good to me.

@triblondon Happy for this to be merged into master?

triblondon:
Yes, it's a good change, imho.

PR #601, opened 2016-02-08
Add Array.prototype.fill polyfill

### LINKED ISSUE #600: Support Array.prototype.fill
Would be great if support for Array.prototype.fill was included.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/fill
https://www.npmjs.com/package/array.prototype.fill

jonathantneal:
This would resolve #600. @maximilianschmitt, would you be willing to find or write a few tests for `fill`? If so, either of us can add them to `tests.js` and this can be merged by another admin upon their review.


JakeChampion:
Added tests and they pass on CI. Ready for review/merging @jonathantneal @triblondon @rowanmanning

polyfills/Array/prototype/fill/polyfill.js
```diff
@@ -0,0 +1,29 @@
+Object.defineProperty(Array.prototype, 'fill', {
+	configurable: true,
+	value: function fill(value) {
+		if (this === undefined || this === null) {
+			throw new TypeError(this + ' is not an object');
+		}
+
+		var arrayLike = this instanceof String ? this.split('') : Object(this);
+
+		var length = Math.max(Math.min(arrayLike.length, 9007199254740991), 0) || 0;
```
rowanmanning, line 10: Maybe this should rely on `Number.MAX_SAFE_INTEGER` (or a polyfilled version of it) for clarity? Can polyfills rely on other polyfills?

polyfills/Array/prototype/fill/polyfill.js
```diff
@@ -0,0 +1,29 @@
+Object.defineProperty(Array.prototype, 'fill', {
+	configurable: true,
+	value: function fill(value) {
+		if (this === undefined || this === null) {
+			throw new TypeError(this + ' is not an object');
+		}
+
+		var arrayLike = this instanceof String ? this.split('') : Object(this);
```
rowanmanning: I might be wrong but I don't think this is covered by the test suite

rowanmanning:
Few comments, otherwise looking good :+1:

polyfills/Array/prototype/fill/polyfill.js
JakeChampion, line 10: They can declare other features in the dependencies array in `config.json`

polyfills/Array/prototype/fill/polyfill.js
JakeChampion: It is not spec compliant, removed.

polyfills/Array/prototype/fill/polyfill.js
JakeChampion, line 10: Reading the spec for ToLength shows it is this exact number --

> 7.1.15 ToLength ( argument )
> 
> The abstract operation ToLength converts argument to an integer suitable for use as the length of an array-like object. It performs the following steps:
> ReturnIfAbrupt(argument).
> Let len be ToInteger(argument).
> ReturnIfAbrupt(len).
> If len ≤ +0, return +0.
> If len is +∞, return 253-1.
> Return min(len, 253-1).

PR #602, opened 2016-02-09
Polyfill fetch for BlackBerry

ironsidevsquincy:


triblondon:
Which BB versions have you tested?

ironsidevsquincy:
10

triblondon:
In that case you can either change the config to target 10+, or test in earlier versions :-)

ironsidevsquincy:
Done

Issue #522, opened 2015-11-23
Number.isNaN support


ebourg:
Hi,

Could you consider adding a polyfill for the ES6 function `Number.isNaN()` please? This function is supported by all major browsers but IE, this is documented in MDN:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

```
Number.isNaN = Number.isNaN || function(value) {
    return typeof value === "number" && isNaN(value);
}
```

Thank you

triblondon:
We'd certainly welcome a submission that adds this polyfill.

Ikkyu-os:
If it's not urgent, i'd be willing to implement this one, to learn a bit how polyfill-service is working.

triblondon:
@Ikkyu-os well volunteered :-)  We really appreciate your help.

Contributor guide is here: http://cdn.polyfill.io/v2/docs/contributing

Ikkyu-os:
FYI, I started working on this issue (at least doing some research pending https://github.com/Financial-Times/polyfill-service/issues/547).

Worst case, I'll find my old linux box to move on.

triblondon:
@Ikkyu-os are you still working on this?

Ikkyu-os:
Hi.

Sorry, was pretty busy with work. I had a first version locally, but then I found out that test suite is also not working well on Windows OS :) 
So I put it on hold, but I'm still willing to fix it. I would actually need to fix tests execution on windows first.

Ikkyu-os:
Hi @triblondon 

I implemented a first version of the polyfill, code is available on my fork: https://github.com/Ikkyu-os/polyfill-service/tree/numberIsNaN/

However, I have difficulties testing it with my test suite (not yet final) with the mocha test suite, using same url pattern as proposed in the doc
http://127.0.0.1:3000/test/tests?feature=Number.isNaN&mode=targeted (or mode=all)

Whatever the browser (last FF, FF 11.0, IE 11.0, IE 8.0 on VM), the test always load empty with the following status in the page:

```
passes: 0 failures: 0 duration: 0s
```

Indeed, the polyfill is not returned by http://127.0.0.1:3000/v2/polyfill.js?flags=gated,always&features=   HTTP request.
I totally understand I'm missing something, which could likely be that my polyfill is not properly integrated in my local build
My build steps:
- grunt build  (here I can see my Number/isNaN properly listed in the build)
[...]

PR #608, opened 2016-02-14
Number isNaN polyfill

Ikkyu-os:
PR for https://github.com/Financial-Times/polyfill-service/issues/522

Tested on several IE & FF versions.


polyfills/Number/isNaN/tests.js
```diff
@@ -0,0 +1,43 @@
...
+	it('numbers', function () {
+		expect(Number.isNaN("blabla")).to.be(false);
+	});
+	it('numbers', function () {
+		expect(Number.isNaN(true)).to.be(false);
+	});
+	it('numbers', function () {
+		expect(Number.isNaN(37)).to.be(false);
+	});
+	it('numbers', function () {
```
triblondon: looks like these tests have some copy/paste remnants that need sorting out.

polyfills/Number/isNaN/config.json
```diff
@@ -0,0 +1,12 @@
...
+	"aliases": [
+		"default"
+	],
+	"browsers": {
+		"ie": "*",
+		"firefox": "<16",
+		"chrome":"<19",
+		"safari": "<16"
+	},
+	"docs": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN"
```
triblondon, line 21: Also add a `spec` link if possible.

triblondon:
Thanks for this.  A few comments for you to deal with.  I am working on a proper contributor agreement, but in the meantime would you please confirm your agreement to the following:
- This code is all your own work
- You are publishing it under an MIT licence

Thanks.

Ikkyu-os:
Hello

To answer your questions:
- The code is my own, although the polyfill itself comes from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN . Not sure if its an issue, as there is not 100 way to implement it anyway :)
- I publish it under the MIT licence, I updated the config file to reflect this.

I updated the test to remove useless code as suggested and add all browsers according to current support.

polyfills/Number/isNaN/tests.js
```diff
@@ -0,0 +1,27 @@
...
+describe('returns true with', function () {
+	it('numbers', function () {
+		expect(Number.isNaN(NaN)).to.be(true);
+		expect(Number.isNaN(Number.NaN)).to.be(true);
+		expect(Number.isNaN(0/0)).to.be(true);
+	});
+});
+
+describe('returns false with', function () {
+	it('numbers', function () {
```
triblondon: incorrectly described tests.  I suggest either drop the `describe`s and call the nested test cases "returns true for NaNs" and "returns false for valid numbers and non-number data types"; or create an `it` for each value, eg returns false with... string, undefined, empty object.

Sorry for the hassle. Hopefully this will help others understand the tests.

Ikkyu-os:
Ok. I didn't understood how they were wrongly described, thanks for the explanation. 

I'll update the tests likely today.

triblondon:
There's a dodgy indent now, but I'll let you off that :-).  Thanks so much for the contribution!

PR #612, opened 2016-02-16
updated semver ranges for safari mobile array properties

wheresrhys:
@JakeChampion


wheresrhys:
According to mdn https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find, safari desktop was out by a decimal point too

JakeChampion:
LGTM :+1:

triblondon:
Tested, happy with this.

Issue #509, opened 2015-11-06
Intl polyfill not provided for SamsungBrowser/2.1


hannesj:
I've noticed that for both user agents below, we don't supply the Intl polyfill, even tough it would be needed, as we see `Intl is not defined` errors from them.

`Mozilla/5.0 (Linux; Android 5.0.2; SAMSUNG SM-G850F Build/LRX22G) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/2.1 Chrome/34.0.1847.76 Mobile Safari/537.36`

`Mozilla/5.0 (Linux; Android 5.0; SAMSUNG SM-G900F Build/LRX21T) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/2.1 Chrome/34.0.1847.76 Mobile Safari/537.36`

triblondon:
We identify both of those UA identifiers as chrome/34, and do not serve the Intl polyfill.  Since we use the useragent npm module for UA identification, we'd need to either recognise it as a separate browser, or alias it to a browser that matches its behaviour more closely.

Yuripetusko:
Seeing same issue:

on Samsung galaxy s6 Android 5.1 using Samsung's "Internet" browser OR Samsung galaxy s3 Android 4.1 using Android's stock "Browser" browser. Intl polyfill doesn't work giving me following error `ReferenceError: Intl is not defined`

antoinerousseau:
Same problem here with a LG G3: `ReferenceError: Intl is not defined`

UA: `Mozilla/5.0 (Linux; U; Android 5.0; es-sa; LG-D855 Build/LRX21R.A1445306351) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/34.0.1847.118 Mobile Safari/537.36`

vibegui:
We had the exact same issue. Samsung Galaxy S6 "Internet" app. UA: `Mozilla/5.0 (Linux; Android 5.1.1; SAMSUNG SM-G920F Build/LMY47X) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/3.2 Chrome/38.0.2125.102 Mobile Safari/537.36`

vibegui:
Apparently `useragent@2.1.8` reports this as "Chrome mobile":

```
[...]

PR #613, opened 2016-02-16
fix: Need Intl polypill for native android browser

Yuripetusko:
This only fixes half of the problem from issue #509. Where `Intl` is not being polyfilled to Android stock browser on devices before android 4.3. This issue is still present on Samsung stock browsers

### DIFF
--- a/polyfills/Intl/config.json
+++ b/polyfills/Intl/config.json
@@ -1,5 +1,6 @@
 {
 	"browsers": {
+		"android": "<=4.3",
 		"ie": "<11",
 		"ie_mob": "10",
 		"firefox": "<29",


JakeChampion:
@Yuripetusko May I enquire as to why you closed this PR?

Yuripetusko:
hey @JakeChampion . 

I couldn't find a way to remotely debug stock browser on android below 4.3 because it doesn't use chromium based browser so I couldn't  reliably tell if it had Intl support or not or maybe I had some other issues. I can definitely see issue with Intl support on Samsung stock browser tho, but to this PR wouldn't fix it

JakeChampion:
According to [caniuse](http://caniuse.com/#search=intl) your change is correct. I'll run the suite of tests on this which will answer the question for us :-)

Yuripetusko:
Thanks @JakeChampion . got a little bit confused there for a moment :)

Turbo87:
I'm not sure what's going on exactly, but this doesn't seem to work properly. I'm trying this in the GenyMotion Android emulator with Android 4.1.1 on a Nexus 4. I get the following output from http://www.whatsmyua.com:

``` json
{
  "ua": "Mozilla/5.0 (Linux; U; Android 4.1.1; en-us; Google Nexus 4 - 4.1.1 - API 16 - 768x1280 Build/JRO03S) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Mobile Safari/534.30",
  "browser": {
    "name": "Android Browser",
    "version": "4.0",
    "major": "4"
  },
  "engine": {
    "version": "534.30",
    "name": "WebKit"
  },
  "os": {
    "name": "Android",
    "version": "4.1.1"
  },
  "device": {
    "model": "Nexus 4",
    "vendor": "LG",
    "type": "mobile"
  },
  "cpu": {}
}
```

when I use that `ua` string to request the polyfill I only see a comment that confirms the detected browser to be `android/4.1.0`, but the polyfill itself is not delivered:

``` js
/* Polyfill service v3.7.1
 * 
 * UA detected: android/4.1.0
 * Features requested: Intl.~locale.en
 *  */
```

(see https://cdn.polyfill.io/v2/polyfill.js?features=Intl.~locale.en&ua=Mozilla/5.0%20(Linux;%20U;%20Android%204.1.1;%20en-us;%20Google%20Nexus%204%20-%204.1.1%20-%20API%2016%20-%20768x1280%20Build/JRO03S)%20AppleWebKit/534.30%20(KHTML,%20like%20Gecko)%20Version/4.0%20Mobile%20Safari/534.30)

Since this PR add `"android": "<=4.3"` and `4.1.0 <= 4.3` I should see the polyfill or am I doing anything wrong?

I'd be happy to debug this if you point me in the right direction on how to get started.

Turbo87:
I think I may have found the issue in https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Intl/~locale/en/config.json. It seems that the `"android": "<=4.3"` config was only added to the `Intl/config.json` but not to the individual locale config files.

This seems somewhat related to https://github.com/Financial-Times/polyfill-service/pull/698

Issue #614, opened 2016-02-17
Include Firefox on iOS in accommodated useragents


andygout:
Firefox on iOS (released Sep 2015; currently ~528 users) is causing JS issues on the front page (videos and markets data don't display) and possibly other apps too.

Seems useragent [FxiOS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Gecko_user_agent_string_reference#Firefox_for_iOS) (which differs from existing useragent `firefox_mob` as has to adhere to iOS) needs adding to the service (poss. `polyfill-service/lib/UA.js`) as requests from Firefox on iOS do not return any polyfill methods (`fetch`, and possibly others, are required).

triblondon:
Hi Andy, thanks for this.  Could you consider making a PR for this? You can find pretty detailed contrib guidelines on the hosted docs at https://cdn.polyfill.io.  You probably need to alias the fxios browser to `firefox` or `firefox_mob`.

JakeChampion:
@triblondon As it is on ios, should we alias it to ios_saf as I believe ios devices don't allow other engines to be installed.

triblondon:
Good point. Andy, the first step is to check what normalised name we give
to the browser: load https://cdn.polyfill.io/v2/polyfill.js and look at the
comment at the top. Then remap that browser to `ios_saf` in the lib/UA.js
module.

ironsidevsquincy:
Think it might be worth distinguishing between `ios_saf` and `ios_firefox`, as safari uses a different engine to other non-safari ios browsers

https://developer.chrome.com/multidevice/ios/overview

Definitely seen issues that occur on Chrome but not on Safari on IOS. Maybe `ios_saf` and `ios_other`?

wheresrhys:
Could `ios_firefox`/`ios_chrome` fallback to `ios_saf` if not defined for a polyfill? More likely to hit the mark than falling back to no polyfills

JakeChampion:
A pull request for this issue is open at https://github.com/Financial-Times/polyfill-service/pull/615

andygout:
Thanks all; have been in UX training all morning (about to go back in). Will read over everything to make sure I understand.

Is it worth my checking if FxiOS should be processed (along with `CriOS` and `OPiOS`) through the `stripiOSWebViewBrowsers` function (have yet to see if Firefox on iOS fits the criteria)?

JakeChampion:
[...]

PR #615, opened 2016-02-17
add ios firefox to uas

JakeChampion:
Relates to #614


triblondon:
Don't want to add more distinct UAs.

PR #617, opened 2016-02-17
Strip Firefox from iOS UAs

JakeChampion:
An alternative approach to #614 #615


triblondon:
Yes, this idea works for me. If it parses the FFiOS UA correctly as `ios_saf` then I'm happy.

JakeChampion:
I've done a test using the UA provided by their documentation -- https://developer.mozilla.org/en-US/docs/Web/HTTP/Gecko_user_agent_string_reference#Firefox_for_iOS

I can check on a real device as well if wanted

triblondon:
I think this is fine.

PR #622, opened 2016-02-22
Update browser compatibility for Array.from

johnmichel:
There were some gaps in browser support coverage for `Array.from`.

### DIFF
--- a/polyfills/Array/from/config.json
+++ b/polyfills/Array/from/config.json
@@ -7,11 +7,12 @@
 	"browsers": {
 		"chrome": "<45",
 		"firefox": "4 - 31",
-		"ie": "6 - *",
+		"ie": "6 - 11",
 		"ie_mob": "10 - *",
 		"opera": "*",
-		"safari": "7 - *",
-		"firefox_mob": "4 - *"
+		"safari": "<9",
+		"firefox_mob": "4 - *",
+		"ios_saf": "3 - 8"
 	},
 	"dependencies": [
 		"Object.defineProperty"


davidchase:
yeah noticing the same issue :+1:

polyfills/Array/from/config.json
```diff
@@ -7,11 +7,12 @@
 	"browsers": {
 		"chrome": "<45",
 		"firefox": "4 - 31",
-		"ie": "6 - *",
+		"ie": "6 - 11",
```
JakeChampion: We are currently using an older version of the [useragent](https://github.com/3rd-Eden/useragent) module which identifies Edge as IE12. We also alias Edge to IE in our [ua.js](https://github.com/Financial-Times/polyfill-service/blob/db69e86e91dc7e12df4e09e20919a9799c93ae92/lib/UA.js#L70) module.

I don't believe IE development to be carrying on, I think it may be fine to leave this as `6 - 11`.

polyfills/Array/from/config.json
```diff
@@ -7,11 +7,12 @@
...
 		"ie_mob": "10 - *",
 		"opera": "*",
-		"safari": "7 - *",
-		"firefox_mob": "4 - *"
+		"safari": "<9",
```
JakeChampion, line 13: :+1:

polyfills/Array/from/config.json
```diff
@@ -7,11 +7,12 @@
...
+		"firefox_mob": "4 - *",
+		"ios_saf": "3 - 8"
```
JakeChampion, line 15: According to [Kangax's compatibility table](https://kangax.github.io/compat-table/es6/#test-Array_static_methods)iOS 9 has partial support for Array.from - it is missing:
Array[Symbol.species]
Array.from, generator instances
Array.from map function, generator instances

Our polyfill does not support these features either IIRC.

triblondon:
Thanks for this.

PR #626, opened 2016-02-25
Update files from build service to use the v2 endpoint

onishiweb:
Updates 2 files to use the v2 API endpoint of the build service.


AlbertoElias:
o-header is now on v4, would it be worth updating it?

onishiweb:
Probably worth upgrading that too while I'm at it I guess.

Unsure why the tests are failing?

PR #630, opened 2016-02-26
fix config for URL

gheoan:
[MDN for `URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility)
[caniuse for `URL.createObjectURL()`](http://caniuse.com/#feat=bloburls)

Are screenshots needed for this? I can test for everything except `safari`.

### DIFF
--- a/polyfills/URL/config.json
+++ b/polyfills/URL/config.json
@@ -1,19 +1,24 @@
 {
 	"aliases": [
+	  	"modernizr:urlparser",
 		"blissfuljs"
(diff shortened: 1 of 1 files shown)


gheoan:
Also [`MDN for searchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Browser_compatibility) as iterator.

triblondon:
Looks good to me!

PR #632, opened 2016-02-29
Add: Browser support links to contributing docs

andygout:
cc @triblondon @JakeChampion 

Seems like the best place to put these links; let me know if there's somewhere preferential.

### DIFF
--- a/docs/features.html
+++ b/docs/features.html
@@ -74,6 +74,11 @@
 			<p>Currently the <a href='https://github.com/3rd-Eden/useragent'>useragent</a> module identifies Edge as IE12, so we follow that, although Edge's official version number is actually in a different range starting at 20.</p>
 		</aside>
 
+		<h2 id="browser-support-links">Browser support links</h3>
+		<li><a href="http://http://caniuse.com">Can I use... (Support tables for HTML5, CSS3, etc)</a></li>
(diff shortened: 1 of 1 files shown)


triblondon:
I think this needs a bit more context. If it stays in the features page, we're saying "for more information about native browser support for the features above, consult one of these resources". But I wonder how common that use case is relative to a contributor wanting to check browser support when writing or updating a config file. If you wanted to improve that area then it would be the authoring polyfills doc (though I think we do cover that quite well in the refactored docs). What are you mainly trying to achieve here?

andygout:
I see you point and think I'm making this suggestion as a first-time polyfill-service user and that, as you suggest, it may well not be a common enough case for contributors wanting/needing to check browser support at time of contributing to warrant them, so probably best to leave them out.

triblondon:
Didn't mean to put you off this change, Andy, sorry. It's probably a good idea, I'm just looking to understand the perspective better, since I'm obviously a bit deeper into it and can't see the wood for the trees.  If you're still keen on it do persevere and we'll find a good place for it.

JakeChampion:
This came about through discussing how to find out which browsers support a feature prior to running tests on all the browsers. I think it would make sense in a section to do with creating or updating polyfills.

andygout:
I'll familiarise myself with the repo a bit more to come up with some suggestions of where they could be located.

PR #633, opened 2016-03-01
Make Element.prototype.classList polyfill configurable

krawaller:
Just like its native counterpart.


triblondon:
This code is syntactically invalid so would have thrown immediately if you'd tried to run it.  I am happy to accept this change in pinciple.

krawaller:
@triblondon Oh dear, what on earth happened there? Fixed now! :smile:

Issue #610, opened 2016-02-14
Polyfill build doesn't work correctly on windows 7


Ikkyu-os:
As discussed in Issue https://github.com/Financial-Times/polyfill-service/issues/522

The directory __dist is not built correctly during grunt build when used in Windows 7.
Instead of generating files from polyfill like this:

```
Object.method.json
```

It generates a directory with files

```
Object/method.json
```

This cause the polyfill-service to not detect those polyfills, and it doesn't serve them.
It is similar to other issues I faced on windows where directory were not created correctly, whereas it was working fine on linux.

netsgnut:
Hi @Ikkyu-os, I have written a PR (#638) on that. Would you mind having a look and checking if that works for you? Thanks!

triblondon:
Fixed in #610, thanks again to @netsgnut and @Ikkyu-os

PR #638, opened 2016-03-08
Fixing Windows build (#610)

netsgnut:
This PR deals with issue #610, which is related to making a Windows build. Currently when built under Windows 7, the `polyfills/__dist` folder would appear as a tree instead of being a flat directory. The commit 209041e fixes that by using both `/` (POSIX platforms) and `\` (Windows platforms) in generating the files.

Bundled together is another commit 24922ea, which adds `polyfill-service.cmd` to `bin/`. This mimics the behavior that Node/NPM on Windows would make a *.cmd to each executable installed by `npm i`.

I have tested on my Windows box (Windows 7 x64), but frankly Windows is not my prior development platform, so I really love to know if there are anything that I should note as well. Thanks!

### DIFF
--- a/bin/polyfill-service.cmd
+++ b/bin/polyfill-service.cmd
@@ -1,0 +1,7 @@
+@IF EXIST "%~dp0\node.exe" (
+  "%~dp0\node.exe"  "%~dp0\polyfill-service" %*
+) ELSE (
+  @SETLOCAL
+  @SET PATHEXT=%PATHEXT:;.JS;=;%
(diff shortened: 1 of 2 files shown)


Ikkyu-os:
Hello.

I just tested the fix on my Windows 7 and it works perfectly :+1: 

Thank you so much for this contribution !

I checked the commits, and they seem fine to me, but I let the repo owner finalize the review . (@triblondon up to you)

netsgnut:
Great to hear it :)

PR #641, opened 2016-03-11
Add acid test mode to features page, closes #611

### LINKED ISSUE #611: 'Acid test' page
It would be useful for the polyfill service to have something similar to the old 'Acid test' pages, which attempted to run some js features and reported the results visually. See http://bit.ly/next-acid for one I've cobbled together for the uses of next.ft.com. It's already proved useful (preventing a regression and aiding with isolating the cause of a bug to the polyfill service)

I suggest something like `cdn.polyfill.io/acid-test?features=featureA&...`

Where the query string follows the same structure as that for the main polyfill service endpoint, and is used to configure the polyfill bundle included in the page and the acid-tests that are included (if absent it shoudl default to all polyfills and all tests). The acid tests can for most polyfills just be that polyfill's `detect` wrapped in a try/catch and some UI code, with a few exceptions requiring custom js/templates.

triblondon:
image

@wheresrhys. Rhys would you mind doing some testing on this while you're reviewing? I've only run it in Chrome and FF.


test/browser/director.html.handlebars
```diff
@@ -87,7 +96,12 @@
 						}
 					}
 				}
-
+				if (results.testedSuites) {
+                    for (var t in results.testedSuites) {
```
JakeChampion, line 100: Mixing tabs and spaces here

triblondon:
Ping @wheresrhys

triblondon:
Poke @wheresrhys

PR #643, opened 2016-03-13
Update syntax to ES6

triblondon:
This PR updates the syntax used across the polyfill service to introduce arrow functions, replace `var` with `const` or `let`, replace the old jshintrc with an eslintrc, and some minor tidying up that resulted from getting the code to pass linting.

It would be good to merge this quickly as it is going to make subsequent PRs hard to merge cleanly.


triblondon:
@rowanmanning maybe you could review this?  Or someone else that wants to know more about the polyfill service?

lib/UA.js
```diff
@@ -132,14 +133,14 @@ var UA = function(uaString) {
...
-				a = aliases[this.ua.family];
+				const a = aliases[this.ua.family];
 				this.ua = new useragent.Agent(a[0], a[1], (a[2] || 0), (a[3] || 0));
 
 			// Map to different family with per-version mapping
 			} else if (typeof aliases[this.ua.family] === 'object') {
-				for (semver in aliases[this.ua.family]) {
-					if (this.ua.satisfies(semver) && Array.isArray(aliases[this.ua.family][semver])) {
-						a = aliases[this.ua.family][semver];
+				for (const semverExpr in aliases[this.ua.family]) {
```
rowanmanning: I'm not sure if this will work – ESlint incorrectly reports that this isn't changed and so should be a constant. The following prints "foo" three times anyway:

``` js
const obj = { foo: 'bar', bar: 'baz', baz: 'qux' };
for (const prop in obj) { console.log(prop); }
```

Worth investigating?

lib/UA.js
```diff
@@ -188,15 +189,14 @@ UA.getBaselines = function() {
...
+	if (typeof name === 'string') return name;
 	if (Array.isArray(name)) return name[0];
 	if (typeof name === 'object' && version) {
-		var ver = version.split('.');
-		var ua = new useragent.Agent(uaName, ver[0], (ver[1] || 0), (ver[2] || 0));
-		for (semver in name) {
-			if (ua.satisfies(semver)) return name[semver][0];
+		const ver = version.split('.');
+		const ua = new useragent.Agent(uaName, ver[0], (ver[1] || 0), (ver[2] || 0));
+		for (const semverExpr in name) {
```
rowanmanning: Possibly the same issue as mentioned above.

lib/index.js
```diff
@@ -1,17 +1,16 @@
-var tsort = require('tsort');
-var _  = require('lodash');
-var AliasResolver = require('./aliases');
-var UA = require('./UA');
-var sourceslib = require('./sources').getCollection();
-var path = require('path');
-var appVersion = require(path.join(__dirname,'../package.json')).version;
+"use strict";
```
rowanmanning: Single quoted for consistency?

tasks/service.js
```diff
@@ -1,196 +1,175 @@
...
+			let hasStarted = false;
+			let pid;
+			let proc;
+			let command;
+			let args;
 
-		    var buffer = "";
-		    var hasStarted = false;
+			const spawned = function() { return data.blocking ? undefined : callback(); };
+			const closed = function() { return data.blocking ? callback() : undefined; };
```
rowanmanning: These could be arrow functions

tasks/service.js
```diff
@@ -1,196 +1,175 @@
...
+				return killByPid(function() {
+					return start(function() {
+						return done();
+					});
+				});
+			case "start":
+				return start(function() {
+					return done();
+				});
+		}
```
rowanmanning, line 166: All of these could be arrow functions.

rowanmanning:
A few comments, but nothing major I don't think. Also build is failing but that might be related to comments.

JakeChampion:
It passes on CI :tada:

tasks/service.js
DavidBruant, line 166: Further, `function(){return done()}` can even be reduced to `done`

lib/UA.js
JakeChampion: This works, which version of Node are you using?

tasks/service.js
JakeChampion: I don't think we get any benefit from them being arrow functions as the binding is not used inside the functions. Perhaps named functions?

tasks/service.js
```diff
@@ -1,196 +1,169 @@
...
-    try {
-      process.kill(pid, 0);
-    } catch (_error) {
-      err = _error;
-      return false;
-    }
-    return true;
-  };
+module.exports = function(grunt) {
+	const log = grunt.log;
```
JakeChampion: Should this bind `grunt` as the context?

lib/UA.js
triblondon: Actually on 5.2.0 I get the same result as @rowanmanning so we should probably stick to `let` here for the moment. We'll need to ignore the eslint warning then

tasks/service.js
triblondon: I am starting to see arrow functions as a way of declaring that a function cannot change the context in which it is run.  So it's actually a stricter form of declaration than `function` and should probably be preferred in this case.

tasks/service.js
triblondon: It probably can just use `grunt.log` like everything else does.  It's just because the code came from a third party with a different style.

tasks/service.js
JakeChampion: And the `const` stops the function being overridden :-)

PR #649, opened 2016-03-14
Add SVG4Everybody to polyfill service

JakeChampion:
In-house copy of https://github.com/Financial-Times/polyfill-service/pull/531 for running against CI


triblondon:
See my comments on #531

PR #652, opened 2016-03-17
Polyfill custom event for all android

wheresrhys:
I've just found that CustomEvent is missing in samsung stock browser on android 4.4.2 so, to paraphrase Donald Trump, I think we should lock this down until we can figure out what the hell is going on


JakeChampion:
We should probably push for Samsung Android detection in the useragent module instead of this blanket change for all Android devices - https://github.com/ua-parser/uap-core/pull/123

triblondon:
What's the UA?

Issue #655, opened 2016-03-17
Proposal: Remove `all` flag for features


JakeChampion:
We strongly recommend not using the kitchen-sink alias `all` as the bundle will be huge, including massive polyfills like Intl with every single language pack, and will also pick up all polyfills we add in future, even those for experimental features where standards are not yet stable.

The kitchen sink polyfill request is rather large to download, it took my machine on a fast internet connection 12.9 minutes to download and a time-to-first-byte of 4.58 seconds.

I propose we remove the `all` alias in favour of users explicitly declaring their dependencies with individual polyfill names or defined sets of polyfills such as `es5` and `blissfuljs`.

triblondon:
I agree with this in principle, and worth saying that we added `all` before we took in all the Intl craziness. It would be hard to argue that this is removable without moving to a new API version, unfortunately. Also there are a fair number of people using it so we also ought to try and address whatever is driving that behaviour.

Assorted ideas:
1. Add `console.log("Requesting all polyfills from polyfill.io is a really bad idea");` to the `all` response.
2. Operate `all` as a blacklist alias rather than a whitelist one, and mark some polyfills as excluded from it.  Then the Intl locales could be excluded, but new polyfills would automatically be included. It should probably be renamed to `almost-all` but would need to continue to support `all` for compat.
3. Reassess which polyfills should be in the default set, to try and address the demand that leads to use of 'all'.  A good way to start this might be to look at what blissfulljs added their alias to, because they had previously advised their users to use `all`.

JakeChampion:
I like ideas 1 & 3 a lot. We should definitely reassess the default set, if we changed it would that require a major version bump? (specifically, if we removed items).

Do we have graphs/data on the usage of each individual polyfill?

triblondon:
I opened https://github.com/Financial-Times/polyfill-service/pull/659 for adding the warning to `all`.

I've also added more polyfills to the default set in https://github.com/Financial-Times/polyfill-service/pull/660

PR #659, opened 2016-03-28
Show console warning when using the 'all' alias

triblondon:


lib/index.js
```diff
@@ -183,6 +183,12 @@ function getPolyfillString(options) {
 					);
 				}
 
+				if ('all' in options.features) {
+					let warnText = 'Using the `all` alias with polyfill.io is a very bad idea. In a future version of the service, `all` will deliver the same behaviour as `default`, so we recommend using default instead.';
```
JakeChampion, line 187: We should create a page (on polyfill.io or Github wiki/issue) which explains why it is a _very bad idea_ to use the `all` alias. Thinking about it some more, maybe it should link to the [Github issue which proposed removing the feature](https://github.com/Financial-Times/polyfill-service/issues/655), that way users can comment on if they would like us to keep the feature around and what their use-cases are. An RFC from our users.

Issue #420, opened 2015-07-02
Array.from doesn't work on Sets in chrome.


AdaRoseCannon:
```
Array.from(new Set([1,2]));
[]
```

triblondon:
The MDN polyfill also fails on this case, but the native implementation in Firefox works.  It's tough to make this work without other parts of ES6.  Eg you could do:

```
var array = [...set];
```

triblondon:
@JakeChampion I was thinking about this one this morning.  Our Array.from implementation assumes array-like things are ES5 constructs, and normalises them to an object with a length and elements as properties.  Whereas the [core-js version](https://github.com/zloirock/core-js/blob/master/modules/es6.array.from.js) also checks whether the object is iterable.  Can we do that or would we need to polyfill iterators to make that work?  What do you think?

JakeChampion:
Maybe this?

``` js
if (typeof Set === "function" && source instanceof Set) {
    var a = source.values();
    var i;
    var array = [];
    var length;
    for (i = 0; ; ++i) {
        var next = a.next();
        if (next.done) {
            break;
        }
        var value = next.value;
        array[i] = value;
        length = i;
    }
}
```

JakeChampion:
We can avoid the dependency of an iterator polyfill if we only care about the standard iterable objects. If we want `Array.from` to work with custom iterables then we would need `Array.from` to be aware of iterable objects.

zloirock:
@JakeChampion too many special cases. For `Set`, `Map`, generators - ok, but strings - should iterate over code points, generic iterables...

JakeChampion:
Note - [es6-shim](https://github.com/paulmillr/es6-shim/blob/b81c0880dad3bcd7d313a40049fe3009babc3fb9/es6-shim.js#L702) supports iterables.

JakeChampion:
@zloirock You raise a good point about iterating over strings. Normal strings would be easy enough. Strings containing surrogate pairs would require some extra effort, [mathiasbynens/String.prototype.at](https://github.com/mathiasbynens/String.prototype.at) should resolve that issue.

Do you think catering to the standard iterable objects would require more code in the shim than making `Symbol` a dependency? _I'm thinking it would_.

zloirock:
@JakeChampion I use [this common code](https://github.com/zloirock/core-js/blob/v2.0.3/modules/_string-at.js) for `String#at`, `String#codePointAt` and `String#@@iterator`.

[...]

PR #665, opened 2016-04-01
Handle the case where an iterable, or a Set or Map is used in array.from

AdaRoseCannon:
This is my first PS PR so I hope it's okay. I've tried to avoid touching anything beyond the most basic syntax.

### DIFF
--- a/polyfills/Array/from/polyfill.js
+++ b/polyfills/Array/from/polyfill.js
@@ -4,27 +4,32 @@
 		var iterableResponse;
 		var tempArray;
 
-		tempArray = [];
-		while (!done) {
-			iterableResponse = arraylike.next();
-			if (
-				iterableResponse.hasOwnProperty('value') &&
-				iterableResponse.hasOwnProperty('done')
-			) {
-				if (iterableResponse.done === true) {
-					done = true;
-					break;
+		// if the iterable doesn't have next;
+		// it is an iterable if 'next' is a function but it has not been defined on
+		// the object itself.
+		if (typeof arraylike.next === 'function' && arraylike.hasOwnProperty('next') === false) {
+			tempArray = [];
+			while (!done) {
+				iterableResponse = arraylike.next();
+				if (
+					iterableResponse.hasOwnProperty('value') &&
+					iterableResponse.hasOwnProperty('done')
+				) {
+					if (iterableResponse.done === true) {
+						done = true;
+						break;
 
-				// handle the case where the done value is not Boolean
-				} else if (iterableResponse.done !== false) {
+					// handle the case where the done value is not Boolean
+					} else if (iterableResponse.done !== false) {
+						break;
+					}
+
+					tempArray.push(iterableResponse.value);
+				} else {
+
+					// it does not conform to the iterable pattern
 					break;
 				}
-
-				tempArray.push(iterableResponse.value);
-			} else {
-
-				// it does not conform to the iterable pattern
-				break;
 			}
 		}
 
@@ -63,11 +68,7 @@
 			var arrayFromIterable;
 
 			// if it is an iterable treat like one
-			// it is an iterable if 'next' is a function but it has not been defined on
-			// the object itself.
-			if (typeof arraylike.next === 'function' && arraylike.hasOwnProperty('next') === false) {
-				arrayFromIterable = parseIterable(arraylike);
-			}
+			arrayFromIterable = parseIterable(arraylike);
 
 			//if it is a Map or a Set then handle them appropriately
 			if (
--- a/polyfills/Array/from/polyfill.js
+++ b/polyfills/Array/from/polyfill.js
@@ -1,4 +1,6 @@
-(function arrayFromPolyfill() {
+
+// Wrapped in IIFE to prevent leaking to global scope.
+(function () {
 	function parseIterable (arraylike) {
 		var done = false;
 		var iterableResponse;
--- a/polyfills/Array/from/polyfill.js
+++ b/polyfills/Array/from/polyfill.js
@@ -75,10 +75,10 @@
 				typeof arraylike.entries === 'function' &&
 				typeof arraylike.values === 'function'
 			) {
-				if (arraylike.constructor.name === 'Set') {
+				if (arraylike.constructor.name === 'Set' && 'values' in Set.prototype) {
 					arrayFromIterable = parseIterable(arraylike.values());
 				}
-				if (arraylike.constructor.name === 'Map') {
+				if (arraylike.constructor.name === 'Map' && 'entries' in Map.prototype) {
 					arrayFromIterable = parseIterable(arraylike.entries());
 				}
 			}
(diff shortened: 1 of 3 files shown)


polyfills/Array/from/polyfill.js
```diff
@@ -20,6 +61,34 @@ Object.defineProperty(Array, 'from', {
...
+			if (arraylike.constructor.name === 'Set') {
+				arrayFromIterable = parseIterable(arraylike.values());
+			}
+			if (arraylike.constructor.name === 'Map') {
+				arrayFromIterable = parseIterable(arraylike.entries());
+			}
+		}
+
+		if (arrayFromIterable) {
+			arraylike = arrayFromIterable;
```
JakeChampion: This looks to me like it is iterating through the arrayLike object once if it does not have an iterable interface and twice if it does have one.

polyfills/Array/from/polyfill.js
```diff
@@ -1,3 +1,44 @@
+
+function parseIterable(arraylike) {
```
JakeChampion: I think this is better off in a closure as to avoid other polyfills getting access to the function

triblondon:
Fixes https://github.com/Financial-Times/polyfill-service/issues/420 ?

polyfills/Array/from/polyfill.js
```diff
@@ -1,3 +1,44 @@
...
+	var
+	done = false,
+	iterableResponse,
+	tempArray;
```
triblondon: If we're importing from a third party repo, this would be fine, but since the polyfill service is to be the canonical home for this polyfill, the syntax should ahere to the jshint config provided, which includes no multi var declarations.

JakeChampion:
Could you add some tests for the feature?

I'd recommend looking at other Array.from polyfills which support iterators such as https://github.com/studio-b12/array-from

triblondon:
I don't know enough about the mechanics of array-like objects in ES6 but as long as the concerns raised in #420 are addressed and we add some tests for the cases that this polyfill seeks to cover, that's all fine with me.

Also - a lingering concern: are there any browsers that ship support for `Set` but not `Array.from`?  If not, this polyfill is pointless.

JakeChampion:
> Also - a lingering concern: are there any browsers that ship support for Set but not Array.from? If not, this polyfill is pointless.

Mathias Bynens has explained a need for this in the past -- https://github.com/mathiasbynens/Array.from/issues/4#issuecomment-51221192

polyfills/Array/from/polyfill.js
AdaRoseCannon: Ah I was following the style already existing in the polyfill I will update that accordingly.

AdaRoseCannon:
Chrome 44 has this issue, and it is the version of Chrome about to be shipped with Samsung's browser for GS6 and GS7.

polyfills/Array/from/polyfill.js
AdaRoseCannon: That is indeed the case, the second time handles the map function, I didn't want to duplicate that logic for a very mild perf issue.

AdaRoseCannon:
What is the environment for tests can I use native Sets and Maps to test it?

JakeChampion:
The environment for tests is every browser that polyfill service supports.

JakeChampion:
The test authoring guidelines should hopefully explain everything -- https://cdn.polyfill.io/v2/docs/contributing#test-authoring-guidelines

AdaRoseCannon:
Ahh I see, i'll look at 'Map' and 'Set' in this. Thanks Jake.

AdaRoseCannon:
I am getting an error when I run the tests locally so I am relying on circle CI.
screenshot 2016-04-02 at 15 07 47

AdaRoseCannon:
Got the tests running by running the service directly. It seems to me by reading the spec I shouldn't use write tests for Map and Set since that particular part is not testable in all browsers.

AdaRoseCannon:
It's now checking for Map and Set in window as it seems some of the other tests are doing that too.

AdaRoseCannon:
Tests are now passing, btw the tests on 'android 4.4.4' seem flakey.

I have seen tests for Promise and URL spontaneously fail.

AdaRoseCannon:
One thing it does is to check that Iterators have a .next function before trying to b=polyfill because Safari 7/8 does not have this.

I tried to make it fall back to `for (i of iterator)` for that specific case, but that won't compile under uglify and doesn't work when transpiled.

JakeChampion:
That also wouldn't work for an endless iterator such as an iterator for even numbers.

AdaRoseCannon:
Array.from isn't supposed to work on an endless iterable. It is expected to try to go forever.

polyfills/Array/from/polyfill.js
```diff
@@ -1,36 +1,104 @@
-Object.defineProperty(Array, 'from', {
-	configurable: true,
-	value: function from(source) {
-		// handle non-objects
-		if (source === undefined || source === null) {
-			throw new TypeError(source + ' is not an object');
+(function arrayFromPolyfill() {
```
triblondon: Leaking a global here.  Use anonymous fn, or for slightly simpler architecture, bring the parseIterable fn inside the Object.defineProperty getter fn

AdaRoseCannon: A named IIFE isn't available from outside it's own scope. This is good practice because it improves debugging by allowing for the name to be available in stack traces.

screenshot from 2016-04-17 09-50-29

triblondon: Ah, I'd forgotten about that, thanks. Is that true in all the browsers we support? (IE6+, FF3.6, Saf4, Chrome1)?

AdaRoseCannon: After reading http://benalman.com/news/2010/11/immediately-invoked-function-expression/ which seems to be the definitive word on iife they make it clear that an iife can have a name but made no mention of it polluting scope. So I am pretty sure it remains true.

https://kangax.github.io/nfe/#jscript-bugs -> Makes me suspect that IE8 and below could potentially have an issue.

I'll look into it. (Or will just remove the name)

Need to get an IE6 VM. I have an IE11 vm in the office, i'll try compatibility mode.

AdaRoseCannon: Good catch @triblondon it leaks in IE6-8. removing the name, leaving wrapped in anonymous. IIFE

triblondon:
Ok, this is good. With the tests, it's sensible to test for features that might not be supported, if those are not features we're trying to polyfill. The alternative is to make this polyfill depend on Map and Set but if they are only used in tests, that's hugely wasteful. So this approach is fine.

erykpiast:
@AdaRoseEdwards, can you explain, what you `if` on `arraylike.hasOwnProperty('next') === false` inside `parseIterable`? It prevents working "fake" iterators (like ones defined inside [es6-collections package](https://www.npmjs.com/package/es6-collections)). I'm not sure if was strict requirement or "just-in-case" condition. As far as I see, there is no test case defined to check exactly this.

PR #670, opened 2016-04-18
Add Map polyfill

triblondon:
This PR adds a polyfill for Map.  Notable:
- I'm using private pseudo-symbol-objects as reliable markers for undefined and NaN since those values are problematic.  Open to better ways of doing this.
- Not using `Object.defineProperty` for all these - could do but I think the current arch is clearer
- The rules on how the iterators and forEach method behave when mutations occur during enumeration are a bit weird to me, but I think I implemented them correctly.  Both my polyfill and the native impl pass my tests.

### DIFF
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -1,64 +1,122 @@
 (function(global) {
 
+
+	// Deleted map items mess with iterator pointers, so rather than removing them mark them as deleted. Can't use undefined or null since those both valid keys so use a private symbol.
+	var undefMarker = {};
+
+	// NaN cannot be found in an array using indexOf, so we encode NaNs using a private symbol.
+	var NaNMarker = {};
+
+	function encodeKey(key) {
+		return Number.isNaN(key) ? NaNMarker : key;
+	}
+	function decodeKey(encodedKey) {
+		return (encodedKey === NaNMarker) ? NaN : encodedKey;
+	}
+
+	function makeIterator(mapInst, getter) {
+		var nextIdx = 0;
+		var done = false;
+		return {
+			next: function() {
+				if (nextIdx === mapInst._keys.length) done = true;
+				if (!done) {
+					while (mapInst._keys[nextIdx] === undefMarker) nextIdx++;
+					return {value: getter.call(mapInst, nextIdx++), done: false};
+				} else {
+					return {done:true};
+				}
+			}
+		}
+	}
+
+	function calcSize(mapInst) {
+		var size = 0;
+		for (var i=0, s=mapInst._keys.length; i<s; i++) {
+			if (mapInst._keys[i] !== undefMarker) size++;
+		}
+		return size;
+	}
+
+	var ACCESSOR_SUPPORT = true;
+
 	var Map = function(data) {
-		var keys, values;
+		this._keys = [];
+		this._values = [];
 
 		// If `data` is iterable (indicated by presence of a forEach method), pre-populate the map
 		data && data.forEach && data.forEach(function (item) {
 			this.set.apply(this, item);
 		}, this);
+
+		if (!ACCESSOR_SUPPORT) this.size = calcSize(this);
 	};
+	Map.prototype = {};
 
-	Object.assign(Map.prototype, {
-		"get": function(key) {
-			var idx = this.keys.indexOf(key);
-			return (idx !== -1) ? this.values[idx] : undefined;
-		},
-		"set": function(key, value) {
-			if (typeof key !== 'object' && typeof key !== 'function')
-				throw new TypeError('Invalid value used as weak map key');
+	// Some old engines do not support ES5 getters/setters.  Since Map only requires these for the size property, we can fall back to setting the size property statically easch time the size of the map changes.
+	try {
+		Object.defineProperty(Map.prototype, 'size', {
+			get: function() {
+				return calcSize(this);
+			}
+		});
+	} catch(e) {
+		ACCESSOR_SUPPORT = false;
+	}
 
-			var entry = key[this.name];
-			if (entry && entry[0] === key)
-				entry[1] = value;
-			else
-				defineProperty(key, this.name, {value: [key, value], writable: true});
-			return this;
-		},
-		"has": function(key) {
-			return (this.keys.indexOf(key) !== -1);
-		},
-		"delete": function(key) {
-			var idx = this.keys.indexOf(key);
-			if (idx === -1) return false;
-			this.keys.splice(idx);
-			this.values.splice(idx);
-			return true;
-		},
-		"clear": function() {
-			this.keys = this.values = [];
-		},
-		get "size": function() {
-			return this.keys.length;
-		},
-		"values": function() {
-
-		},
-		"keys": function() {
-
-		},
-		"entries": function() {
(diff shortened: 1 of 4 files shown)


triblondon:
Still working on test failures in IE6-8.

polyfills/Map/polyfill.js
```diff
@@ -0,0 +1,124 @@
+(function(global) {
+
+
+	// Deleted map items mess with iterator pointers, so rather than removing them mark them as deleted. Can't use undefined or null since those both valid keys so use a private symbol.
+	var undefMarker = {};
```
JakeChampion: We could create a symbol using `Symbol()` because we have Symbol declared as a dependency for this polyfill.

polyfills/Map/polyfill.js
```diff
@@ -0,0 +1,124 @@
...
+
+	// NaN cannot be found in an array using indexOf, so we encode NaNs using a private symbol.
+	var NaNMarker = {};
```
JakeChampion: Same as above, we can use `Symbol()` here.

polyfills/Map/polyfill.js
triblondon: I didn't use Symbol because private symbols are not really private, so a basic empty object seems better to me.

https://mail.mozilla.org/pipermail/es-discuss/2014-January/035604.html

triblondon:
OK, testing old IE:
- Native impl isn't quite there in IE11, so have extended the detect to fail in IE11 and targeted 11 in config.  We could polyfill IE11 selectively, but easier just to replace the dodgy native impl with the polyfill.
- The polyfill already worked in old IE (as far back as 5!) but the tests required some more modern things like use of reserved words as object properties (so `mapinst.delete` => `mapinst['delete']`) and JSON encoding/decoding (so dropped use of JSON.stringify and replaced with string concat)

polyfills/Map/polyfill.js
JakeChampion: As these are in a closure and not added to the returned object, the symbols wouldn't be surfaced via `getOwnPropertySymbols()`.

polyfills/Map/polyfill.js
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	var ACCESSOR_SUPPORT = true;
+
+	var Map = function(data) {
+		this._keys = [];
+		this._values = [];
+
+		// If `data` is iterable (indicated by presence of a forEach method), pre-populate the map
+		data && data.forEach && data.forEach(function (item) {
```
JakeChampion: Could we use the `Symbol.iterator` here prior to using this iterable detection technique? We should also detect the type of `forEach` is a function.

`(data && data[Symbol.iterator]) || (data && typeof data.forEach === 'function)`

polyfills/Map/polyfill.js
```diff
@@ -0,0 +1,124 @@
...
+			this.set.apply(this, item);
+		}, this);
+
+		if (!ACCESSOR_SUPPORT) this.size = calcSize(this);
+	};
+	Map.prototype = {};
+
+	// Some old engines do not support ES5 getters/setters.  Since Map only requires these for the size property, we can fall back to setting the size property statically easch time the size of the map changes.
```
JakeChampion: typo: `easch` -> `each`

polyfills/Map/polyfill.js
```diff
@@ -0,0 +1,124 @@
...
+		Object.defineProperty(Map.prototype, 'size', {
+			get: function() {
+				return calcSize(this);
+			}
+		});
+	} catch(e) {
+		ACCESSOR_SUPPORT = false;
+	}
+
+	Map.prototype['get'] = function(key) {
```
JakeChampion, line 67: Is the array notation being used for consistency? IIRC `get` is not a reserved word and would be safe to use in dot-notation.

polyfills/Map/tests.js
```diff
@@ -0,0 +1,183 @@
...
+	expect(m.has(b)).to.be(true);
+	expect(m.has(c)).to.be(true);
+	expect(m.size).to.equal(3);
+});
+
+it("implements .size()", function () {
+	expect(o.size).to.be(0);
+	o.set("a", "a");
+	expect(o.size).to.be(1);
+	o["delete"]("a"); // Use square-bracket syntax to avoid a reserved word in old browsers
```
JakeChampion, line 33: Should this be a note in the polyfills config.json? I imagine many developers will forget about `delete` being a reserved word even if it is a method on the object being called upon.

polyfills/Map/tests.js
```diff
@@ -0,0 +1,183 @@
...
+	// new element shows up in iterators that didn't yet finish
+	expect(entriesagain.next().value[0]).to.be("4");
+	expect(entriesagain.next().done).to.be(true);
+});
+
+it("implements .forEach()", function () {
+	var o = new Map(), i;
+	o.set("key 0", 0);
+	o.set("key 1", 1);
+	if ("forEach" in o) {
```
JakeChampion: why does this guard exist?

polyfills/Map/tests.js
```diff
@@ -0,0 +1,183 @@
...
+			expect(key).to.be("key " + value);
+			expect(obj).to.be(o);
+			// even if dropped, keeps looping
+			o["delete"](key);
+		});
+		expect(o.size).to.be(0);
+	}
+});
+
+it("supports mutations during forEach loops", function () {
```
JakeChampion, line 156: 👍

polyfills/Map/polyfill.js
triblondon: OK, so what are the benefits of that?

polyfills/Map/polyfill.js
triblondon: I'm not sure there's a benefit here in introducing a Symbol dependency, but very happy to type check the forEach prop.

polyfills/Map/polyfill.js
triblondon, line 67: Yes, it's consistency.

polyfills/Map/tests.js
triblondon, line 33: Added to notes

polyfills/Map/tests.js
triblondon, line 156: Yeah, this was a pain.

polyfills/Map/tests.js
triblondon: Not sure. Although I wrote the polyfill from scratch I actually largely stole the test cases from https://github.com/WebReflection/es6-collections (which I've now credited in the notes), and he had that guard in there.  I've removed it.

polyfills/Map/polyfill.js
JakeChampion: We already depend upon Symbol directly by depending on `Symbol.iterator` and `Symbol.species`. The benefit would be that it would work with custom iterators which don't have a `forEach` method.

polyfills/Map/polyfill.js
JakeChampion: It's more declarative about it's purpose. No other benefit.

triblondon:
OK, I think I've addressed all these comments.

JakeChampion:
LGTM 👍

PR #671, opened 2016-04-18
Samsung browser detection

### LINKED ISSUE #509: Intl polyfill not provided for SamsungBrowser/2.1

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-service/issues/509 and paves the way to resolve https://github.com/Financial-Times/polyfill-service/pull/652

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -31,7 +31,8 @@ const baseLineVersions = {
 	"opera": ">=11",
 	"op_mob": ">=10",
 	"op_mini": ">=5",
-	"bb": ">=6"
+	"bb": ">=6",
+	"samsung_mob": "*"
 };
 
 /* Aliases may be expressed in three formats:
@@ -65,6 +66,8 @@ const aliases = {
 	"chrome mobile": "chrome",
 	"chrome frame": "chrome",
 	"chromium": "chrome",
+
+	"samsung internet": "samsung_mob",
 
 	"ie mobile": "ie_mob",
 
--- a/polyfills/Intl/config.json
+++ b/polyfills/Intl/config.json
@@ -8,7 +8,8 @@
 		"chrome": "<24",
 		"safari": "*",
 		"ios_saf": "*",
-		"firefox_mob": "*"
+		"firefox_mob": "*",
+		"samsung_mob": "*"
 	},
 	"dependencies": [],
 	"build": {
--- a/polyfills/Intl/~locale/af-NA/config.json
+++ b/polyfills/Intl/~locale/af-NA/config.json
@@ -1,28 +1,30 @@
 {
-	"browsers": {
-		"ie": "<11",
-		"ie_mob": "10",
-		"firefox": "<29",
-		"opera": "<15",
-		"chrome": "<24",
-		"safari": "*",
-		"ios_saf": "*",
-		"firefox_mob": "*"
-	},
-	"dependencies": [
-		"Intl"
-	],
-	"build": {
-		"minify": false
-	},
-	"license": "MIT",
-	"spec": "http://www.ecma-international.org/ecma-402/1.0/",
-	"repo": "https://github.com/andyearnshaw/Intl.js",
-	"docs": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl",
-	"notes": [
-		"Locales must be specified separately by prefixing the locale name with `Intl.~locale`, eg `Intl.~locale.en-GB`."
-	],
-	"test": {
-		"ci": false
-	}
+    "browsers": {
+        "android": "<=4.3",
+        "ie": "<11",
+        "ie_mob": "10",
+        "firefox": "<29",
+        "opera": "<15",
+        "chrome": "<24",
+        "safari": "*",
+        "ios_saf": "*",
+        "firefox_mob": "*",
+        "samsung_mob": "*"
+    },
+    "dependencies": [
+        "Intl"
+    ],
+    "build": {
+        "minify": false
+    },
+    "license": "MIT",
+    "spec": "http://www.ecma-international.org/ecma-402/1.0/",
+    "repo": "https://github.com/andyearnshaw/Intl.js",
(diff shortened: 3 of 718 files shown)


hannesj:
Probably we also want to polyfill everything that is polyfilled to the default android browser and not only Intl.

triblondon:
Adding a family is a major maintenance overhead. I think we can only do this, particularly for a niche browser not available on sauce, if someone takes responsibility for maintaining the polyfill targets.

I reached out to Samsung about this but haven't made much progress yet.

loris:
Any update on this one? I do confirm that right now any website depending on `Intl` is broken on Samsung phones (they do have quite a nice market share, by the way). Is there a workaround?

JakeChampion:
There is two workarounds that I can think of:
- You can request the polyfill to always be returned using `|always` after the specific polyfill - `https://cdn.polyfill.io/v2/polyfill.js?features=Intl.~locale.en|always`

Or
- Request the polyfill-service polyfills dynamically using Javascript. Before making the request, sniff the browser's useragent, if it is a Samsung device, make the request using `|always` otherwise make the request as normal. This will help ensure browser which have the locale/s you are wanting to use don't download the locale pack/s from the polyfill-service. - 

``` js
var script = document.createElement('script');
var alwaysPolyfillUrl = 'https://cdn.polyfill.io/v2/polyfill.js?features=Intl.~locale.en|always';
var polyfillUrl = 'https://cdn.polyfill.io/v2/polyfill.js?features=Intl.~locale.en';
var url = navigator.userAgent.includes('SAMSUNG') ? alwaysPolyfillUrl : polyfillUrl;
script.setAttribute('src', url);
document.head.appendChild(script)
```

loris:
Thanks @JakeChampion, looks fine!

loris:
@JakeChampion Actually I spoke too quickly, it does not work with a Samsung Device I could get my hands on (`SamsungBrowser/3.3 Chrome/38.0.2125.102`). It looks the issue is not at the UA sniffing level but in the polyfill:

```
    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Intl.~locale.en|always"></script>
    <script>
      window.alert(window.Intl);
    </script>
```

It does output `undefined`

BTW, it works as expected when using the following polyfill instead: `http://polyfills.yahooapis.com/polyfill.min.js?version=2.1.9&features=Intl%2Clocale-data-en-US`

JakeChampion:
The polyfill itself is maintained by a different team and in a different repository on Github. If the polyfill has an issue, could you open an issue at https://github.com/andyearnshaw/Intl.js describing the issue and with an small reproducible example? Thanks for taking the time to look into this with us

triblondon:
I haven't had a response from Samsung on this.  In the absence of either well maintained feature support tables (eg caniuse, MDN or kangax) or commitment from the vendor to maintain the targeting for their family, I don't want to add a new family for this browser - it would reflect badly on the service to suggest that we have assessed polyfill compatibility for all features for all versions of a browser when in fact we've just looked at a single feature and haven't tested it (in CI) in any versions of the browser.

So I propose to close and revisit if/when Samsung provides support or publishes feature compat data.

torgo:
Hi @triblondon – sorry for the late response. With my Samsung Browser team hat on, I am working to get support from Samsung on this. Also CCing @jungkees, one of the core browser team members (and co-editor of Service Workers). We are still not sure exactly who will be contributing but there intention is there to contribute. Also note I am reaching out to caniuse to get them to recognize Samsung Browser differently. As you can see from the [stat counter data](http://gs.statcounter.com/#mobile_browser-ww-monthly-201504-201604) Samsung Browser is starting to pick up a pretty significant share of mobile browsing. So the answer is that Samsung are tentatively committed to maintain and we are putting together the team to back up that commitment now.

caridy:
intl@1.2.3 solves this issue. https://github.com/Financial-Times/polyfill-service/pull/698 can be revamped, and eventually merged.

triblondon:
@caridy I think this issue is about the need for Intl polyfilling in Samsung browser, not whether the particular Intl polyfill that we carry works in that browser or not.  Currently we don't serve it to Samsung browser so even if it did work it wouldn't be applied.

triblondon:
Having discussed this on the call with @JakeChampion today, we're going to go ahead and close this.  Adding a browser family is a big deal, as we covered earlier, and we want to do it properly.  So this would involve:
- Having more than one polyfill that we can target at the browser
- Access to a good, reliable source of support data (eg caniuse, MDN or the vendor maintaining the targeting directly)
- Access to instances of the browser in the cloud for testing

ncri:
Is there a polyfill that can be used for that browser? We see every week thousands of users with samsung devices running into this issue, basically finding a broken site.

triblondon:
Hi @ncri, see https://github.com/Financial-Times/polyfill-service/pull/671#issuecomment-218274792 for a possible solution, and bear in mind that we are in the process of updating the version of the Intl polyfill that we host, so any issues with the polyfill itself are separate to the ability to load it in Samsung browsers.

ncri:
Thanks, however 2 comments down someone says it doesn't work. And yes, I understand that. It's okay for us to work around the issue by detecting the browser ourselves. But that only helps if there is a working polyfill.

triblondon:
The existence of an up to date polyfill is not related to this issue.  We are working on that here: https://github.com/Financial-Times/polyfill-service/pull/698 and will ship it when it passes the tests.

ncri:
Thanks.

PR #672, opened 2016-04-18
Add Promise to  es6 set

### LINKED ISSUE #667: Promise polyfill should be a part of es6 set
https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Promise/config.json#L2

It looks like at the moment it isn't a part of `es6` set. Is there a reason for that?

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-service/issues/667


triblondon:
LGTM

Issue #646, opened 2016-03-14
Use HSTS


DavidBruant:
[https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security](https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security) upgrades an HTTP connection to an HTTPS one and tells the browser to remember that this domain should only be accessed via HTTPS.
This increases security and only costs something at the very first request to the domain.

HSTS is easy to implement on an express website using [helmet](https://www.npmjs.com/package/helmet).

Happy to make the contrib if there is an interest.

triblondon:
Go ahead, thanks :-)

Just be aware that we don't (yet) support HTTPS on the apex domain, only on cdn.polyfill.io subdomain, for Fastly-related reasons.

DavidBruant:
> Just be aware that we don't (yet) support HTTPS on the apex domain

I'm not sure I understand this part (and the consequences for adding HSTS support). What does "the apex domain" mean? (or what is it?)

triblondon:
ah, sorry: apex domain means polyfill.io, vs cdn.polyfill.io which is a
subdomain. If you load http://polyfill.io, you get redirected to
https://cdn.polyfill.io. I'm happy to have HSTS apply to the latter.

DavidBruant:
To answer your concern, I need an answer. I'll try to phrase the question as accurately as possible:
Is code from the Financial-Times/polyfill-service repo deployed under the polyfill.io domain? (is it this code that triggers the redirect). If yes, I'm not sure it's a good idea to add HSTS in the current circumstances. If both domains are even on the same physical machine, but handled by different code, then I feel confident HSTS won't affect polyfill.io.

Asked differently, is the code that redirects polyfill.io to cdn.polyfill.io under the current repository?

triblondon:
OK, I understand. polyfill.io's redirect is not handled by the code in the repo - the response comes directly from our DNS provider.  This repo's code is used only to serve cdn.polyfill.io.  Basically, all I'm saying is that `includeSubDomains` should not be set in the HSTS header.

DavidBruant:
> This repo's code is used only to serve cdn.polyfill.io

ok

>  Basically, all I'm saying is that includeSubDomains should not be set in the HSTS header.

[...]

PR #673, opened 2016-04-18
Add hsts header

JakeChampion:
Resolves -- https://github.com/Financial-Times/polyfill-service/issues/646


triblondon:
Do we really need an additional dependency to write a single line, static HTTP header?

JakeChampion:
@triblondon I removed the module and added the header. Ready for review.

triblondon:
Ok, this looks fine. No need for the moment upgrade but no harm I guess. Not sure it's obvious to me why you chose 18 weeks, so if there was a reason, a comment in the code would be useful.

JakeChampion:
@triblondon Switched to one_year

PR #682, opened 2016-05-02
Force TLS, and update VCL on deploy, closes #677

### LINKED ISSUE #677: Redirect HTTP to HTTPS
We now have HSTS #673 - however this is is only considered when a user visits HTTPS which means HTTP can still be used if the user never visits HTTPS. I imagine we can do this in a variety of ways such as adding [express-enforces-ssl](https://github.com/aredo/express-enforces-ssl) to our application server or possibly by adding a redirect inside the VCL on Fastly.

triblondon:
I'm not sure about embedding the service IDs in the gruntfile.  What do you think?

I've test-deployed this to the Fastly QA app and it works fine, including erroring appropriately when the VCL has syntax errors.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -81,6 +81,10 @@ module.exports = function(grunt) {
 			tasks: {
 				src: ['polyfills/**/update.task.js'],
 			}
+		},
+		"deployvcl": {
+			qa: { options: {serviceId: "2l7QhcR5rkdg691tcdRTD7"} },
+			prod: { options: {serviceId: "4E1GeTez3EFH3cnwfyMAog"} }
 		}
 	});
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -83,6 +83,7 @@ module.exports = function(grunt) {
 			}
 		},
 		"deployvcl": {
+			dryrun: { options: {serviceId: "2l7QhcR5rkdg691tcdRTD7", dryRun: true} },
 			qa: { options: {serviceId: "2l7QhcR5rkdg691tcdRTD7"} },
 			prod: { options: {serviceId: "4E1GeTez3EFH3cnwfyMAog"} }
 		}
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -62,6 +62,7 @@
     "babel-core": "^6.2.1",
     "babel-preset-es2015": "^6.1.18",
     "batch": "^0.5.1",
+    "fastly": "Financial-Times/fastly#v2.5.0",
```
matthew-andrews, line 65: i recommend this for fastly deployments:- https://github.com/Financial-Times/fastly-tools

we don't really maintain this other one anymore i think.

triblondon, line 65: Ok, thanks.  I'm linking to a fixed tag so we should be OK with this one.  I'm not sure whether we are moving towards or away from CLI tools at the moment, with haikro on it's way out once we get a docker architecture we like, so for the moment i'd be pretty happy using a grunt task that fits in with our existing task architecture.

Also the fastly-tools thing appears not to support dry running a VCL without activating it.

matthew-andrews, line 65: hm, not really sure i follow what you mean about grunt and cli tools.  fastly-tools performs an identical role in an equivalent way to the financial-times/fastly module — only it's a bit more supported+tested.

> dry running a VCL without activating it

that sounds like a useful thing we should re-add.

Gruntfile.js
```diff
@@ -81,6 +81,11 @@ module.exports = function(grunt) {
 			tasks: {
 				src: ['polyfills/**/update.task.js'],
 			}
+		},
+		"deployvcl": {
+			dryrun: { options: {serviceId: "2l7QhcR5rkdg691tcdRTD7", dryRun: true} },
```
JakeChampion: These could be env vars

triblondon: Separate vars for each target, or do we expect the env to manage the right token?

So ie. `FASTLY_SERVICE_ID` or `FASTLY_SERVICE_ID_[QA|PROD]`?

JakeChampion: I'd go with separate.

PR #683, opened 2016-05-02
Exclude option, closes #435

### LINKED ISSUE #435: Add "exclude" config option
There is an `|always` flag to make sure the bundle always contains a given feature regardless of the UA string. I think there should also be a `|never` flag for preventing a given feature being polyfilled when it usually would be.

Use case:
- I've already polyfilled `window.Promise` using Bluebird (because I'm already using it in my application code, so why not use it as my global Promise polyfill).
- I want to use the polyfill service to load the `default` set and `fetch`
- ...but loading `fetch` will cause the polyfill service to include `Promise` in the bundle - overwriting my own copy, or at least wasting bandwidth.
- so it would be nice if I could do `?features=default,fetch,Promise|never`

triblondon:
Adds the option to use `excludes` as both a library argument and an API param, so that automatically included dependencies can be selectively excluded.

@callumlocke


JakeChampion:
LGTM 👍

PR #685, opened 2016-05-03
Daily stats

triblondon:
Improve the display of our traffic growth by:
- using a line chart rather than columns
- grouping by day to avoid daily cycle effects (still get weekly cycle effects)
- increase time range from 7 to 180 days

image


triblondon:
Note that https://github.com/Financial-Times/polyfill-service/pull/685/commits/e27b4083d390454a98fd3738e8e00aa4133e6435 is already in master.

PR #686, opened 2016-05-03
Use Node 6 to handle HTTP clientError scenarios correctly, fixes #567

### LINKED ISSUE #567: Connection closed without response on Heroku
Sometimes Heroku logs their H13 platform error "Connection closed without response".  This is reproducible by sending malformed HTTP requests to the service, and results from an interaction between Heroku router (which insists that the backend produce a response) and Node (which closes connections that send malformed requests without a response).  Neither of these behaviours are currently modifiable by us.

Node issue raised: https://github.com/nodejs/node/issues/4543
Heroku issue raised: https://help.heroku.com/tickets/320113

It seems like Heroku will be making a change to their router that will make the issue disappear, and Node contributors seem amenable to giving developers control over HTTP responses to malformed requests.  We should keep this issue open while the H13 errors are still occurring.

triblondon:
Before:

image

After:

image

### DIFF
--- a/bin/polyfill-service
+++ b/bin/polyfill-service
@@ -18,6 +18,10 @@ var startService = require('../service');
 var argv = minimist(process.argv.slice(2));
 var port = argv.port || Number(process.env.PORT) || 3000;
 
+process.on('uncaughtException', (err) => {
+  console.log('uncaught exception', err.stack || err);
+});
+
 startService(port, function(err, app){
 	if (err) {
 		console.error('Error while starting service:', err);
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Requirements
 
 Running Origami Build Service requires a few tools:
   * [Git]: For downloading the source code
-  * [Node.js] 5.x and [npm] 3.x: For installing the dependencies and running the application (npm is installed with Node.js)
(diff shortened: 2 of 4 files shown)


bin/polyfill-service
```diff
@@ -18,6 +18,10 @@ var startService = require('../service');
 var argv = minimist(process.argv.slice(2));
 var port = argv.port || Number(process.env.PORT) || 3000;
 
+process.on('uncaughtException', (err) => {
```
JakeChampion, line 21: What prompted the adding of this event handler?

triblondon, line 21: Initially I made a mistake in index.js and referred to the `sock` var as `socket`.  Until I added the code above, all I saw on the console was this:

```
3 May 12:38:09 - uncaughtException: db42aa429377452ebe9a3a5ad36fee73
```

JakeChampion, line 21: Should we send the error to Sentry as well?

triblondon, line 21: Seems sentry actually already logs it

https://app.getsentry.com/nextftcom/polyfill-service-dev/issues/126700489/

JakeChampion:
LGTM 👍

PR #687, opened 2016-05-03
Raise baseline to exclude IE6, Safari 3, Android 2.  Fixes #456

### LINKED ISSUE #456: Raise baseline to exclude older IEs, iOS, Safari, Android
Quite a number of our polyfills are intended for IE6 and 7, and they are generating issues like https://github.com/Financial-Times/polyfill-service/issues/452.  The FT doesn't support these browsers for its own products, but does not want to be irresponsible in not dealing with issues with polyfills intended for older browsers that may be causing significant problems like the one above.

Given the usage of these older browsers (tiny), and difficulties in running tests on them, I propose we increase the support baseline of the polyfill service to exclude the following browsers that we currently support:
- IE: 6, 7
- Safari: 3, 4, 5
- iOS Safari: 3, 4, 5, 6
- Android: 2, 3

Caniuse reports that global usage of the above browsers combined is less than 0.5% of total traffic.

The impact of this change would be that we would serve empty files to these browsers, rather than the requested polyfills, and we would also remove any polyfills intended only for those browsers.  By doing this we are not suggesting that users on these browsers should be unsupported, but that a technique like "Cuts the mustard" should be used so that more advanced JavaScript features are only accessed in newer user agents.

triblondon:
I think the majority of polyfill.io users are not testing IE6 and have no significant audience using it.  It would be better not to try and serve polyfills to it anymore.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -152,7 +152,6 @@ const browsers = {
 		'ie/9',
 		'ie/8',
 		'ie/7',
-		'ie/6',
 		'safari/9',
 		'safari/8',
 		'android/4.4'
@@ -177,7 +176,6 @@ const browsers = {
 		'ie/9',
 		'ie/8',
 		'ie/7',
-		'ie/6',
 		'safari/9',
 		'safari/8',
 		'safari/5.1',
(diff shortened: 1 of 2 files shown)


JakeChampion:
I thought FT Chinese was still had an audience on IE6?

JakeChampion:
Looks like FT has dropped IE6 support. 🎉

PR #689, opened 2016-05-03
Add navigator.sendBeacon, closes #539

### LINKED ISSUE #539: navigator.sendBeacon
I hope requests are acceptable here. I would like to request a navigator.sendBeacon polyfill. This is mostly useful for sending analytics data, and is currently supported natively in most browsers (except, y'know, IE).

[Here is more, from Google,](https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits#specifying_different_transport_mechanisms) on how it works, and here is [current browser stats](http://caniuse.com/#search=sendBeacon). There seems to be quite a few polyfills available - I have not researched any of them, and hence am not making any recommendations.

triblondon:
Adds a sendBeacon polyfill.  Some caveats:
- I can't find any tests for sendBeacon.  There aren't any in the polyfill repo, and sendBeacon is only mentioned in web platform tests as part of CSP test suites.  Creating tests would be hard since the success criteria is that the request made it to the server.  A serviceworker could be used, I suppose, but then the tests will only run in browsers that support SW.  A detect will need to suffice for now.
- The polyfill source includes a UMD, which is fine, but it also references `navigator` without pulling it off the `root` object.  This is probably an oversight on the part of the author, so we should probably push that change upstream.

### DIFF
--- a/polyfills/navigator/sendBeacon/polyfill.js
+++ b/polyfills/navigator/sendBeacon/polyfill.js
@@ -1,33 +1,15 @@
-(function(root) {
-
-	function sendBeacon(url, data) {
-		var xhr = ('XMLHttpRequest' in window) ? new XMLHttpRequest() : new ActiveXObject('Microsoft.XMLHTTP');
-		xhr.open('POST', url, false);
-		xhr.setRequestHeader('Accept', '*/*');
-		if (typeof data === 'string') {
-			xhr.setRequestHeader('Content-Type', 'text/plain;charset=UTF-8');
-			xhr.responseType = 'text/plain';
-		} else if (Object.prototype.toString.call(data) === '[object Blob]') {
-			if (data.type) {
-				xhr.setRequestHeader('Content-Type', data.type);
-			}
+this.navigator && this.navigator.sendBeacon = function sendBeacon(url, data) {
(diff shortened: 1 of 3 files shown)


polyfills/navigator/sendBeacon/polyfill.js
```diff
@@ -0,0 +1,33 @@
...
+			if (data.type) {
+				xhr.setRequestHeader('Content-Type', data.type);
+			}
+		}
+		xhr.send(data);
+		return true;
+	}
+
+	if (!('sendBeacon' in navigator)) {
+		navigator.sendBeacon = sendBeacon;
```
triblondon: @JakeChampion what do you think about the global ref to `navigator` here?  I think it should be:

```
root.navigator.sendBeacon = sendBeacon;
```

And it should probably be in a final `else` on the UMD below, so it does not expose the global if in commonJS or AMD environments (though that's irrelevant to us)

JakeChampion: Agree. Could we strip out the UMD wrapper?

triblondon: It's not doing any harm, and I'd rather keep the polyfill aligned with the third party source.  I'll push the navigator change upstream.

triblondon:
Proposed https://github.com/miguelmota/Navigator.sendBeacon/pull/4, and will commit the equivalent change directly to our copy.

PR #708, opened 2016-05-25
Fixed Object.create polyfill issue when called through Babel's Object's create wrapper

AlbertoElias:
Opening this in favor of #706 to run CI

Original comment by @roland-vachter:

"arguments can be 2 even if descriptors is explicitly undefined, so it's good to check if it's not undefined. Tested, solves the issue."


AlbertoElias:
@JakeChampion is on holiday but he gave the ok before leaving

triblondon:
Looks OK to me.

PR #713, opened 2016-05-31
Reinstate the existing Event prototype, fixes #705

### LINKED ISSUE #705: Inheritance of events is broken on IE11
The following test was passed before using your polyfill.

Note: `ev` is a click event object.

```
IE 11.0.0 (Windows 7 0.0.0) Unit: lib/dom bind click FAILED
        AssertionError:   #

          assert(ev instanceof Event)
                 |  |          |
                 |  false      #function#
                 PointerEvent{}
```

triblondon:
My knowledge of JS prototypes is pretty sketchy, but I think this may fix the problem described in #705.  The issue is that redefinition of `Event` (in order to make it constructable in IE9-11) detaches it from the prototype chain for event instances.  This appears to be what's happening:

image

### DIFF
--- a/polyfills/Event/config.json
+++ b/polyfills/Event/config.json
@@ -13,7 +13,8 @@
 	"dependencies": [
 		"Window",
 		"Document",
-		"Element"
+		"Element",
+		"Object.defineProperty"
 	],
 	"docs": "https://developer.mozilla.org/en/docs/Web/API/Event",
 	"notes": [
--- a/polyfills/Event/detect.js
+++ b/polyfills/Event/detect.js
@@ -5,7 +5,7 @@
 
 	try {
 
-		// In IE 9 and 10, the Event object exists but cannot be instantiated
+		// In IE 9-11, the Event object exists but cannot be instantiated
 		new Event('click');
 		return true;
 	} catch(e) {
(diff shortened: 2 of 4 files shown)


JakeChampion:
@triblondon Could you add a test for this bug-fix?

triblondon:
Sure.  I just reached out to @getify to see if this solution is non-bonkers enough to fly, and if so tests will ensue.

triblondon:
Quick [response from @getify on twitter](https://twitter.com/getify/status/737824639813046273):

>  first glance, i think there may be an issue where you need to use defineProperty instead, but will play w/ it ASAP

triblondon:
Can you define the `prototype` property using `defineProperty`?  I guess so!  Maybe what I've done makes it enumerable?

getify:
The other thing I'm partially recalling besides enumerability is that IIRC some prototypes of built-ins are read only, and failed settting, like maybe in older IE or something.

triblondon:
We no longer support polyfilling IE6, and a native Event.prototype only exists in IE10+ anyway, so we'll only be attempting to run this code path in 10+ of IE.  It passes our test suite in IE 6-13 (where 12 and 13 are Edge!).  I've never seen anyone define a prototype with defineProperty before, which is why this form seemed likely to be 'more right'.  I'll try using defineProperty.

PR #719, opened 2016-06-01
Add commonJS support for external modules

JakeChampion:
Problem: Consuming third-party modules is currently a bit of a complex issue which involves creating an `update.task.js` file and running a grunt task `grunt updatelibrary`.

Solution: Add a new `build` config to classify a module as being `commonjs`, run the polyfill through browserify to create a browser bundle.


JakeChampion:
I've added a commit which shows one polyfill migration using this approach.
@wheresrhys, @callumlocke, @triblondon - What are your opinions on this?

wheresrhys:
Does this mean
- each polyfill is built separately and has its own copy of the shared code
- all polyfills in a bundle share a single copy of the shared components

For the example you provide both woudl give the same result but e.g. for polyfilling all the different iterator types the former would mean lots of duplication (though with a much improved workflow 👍 )

If the latter is hard to achieve, could the following be achieved in stead: 

For cases where
- a number of polyfills share some underlying functionality
- when a user wants one they probably want them all
- these polyfills are aliased to some word e.g. iterables = `Arry.prototype.@@iterator` + `NodeList.prototype.@@iterator` + ....  

then a bundle for that alias will be constructed that _does_ share code between polyfills?

JakeChampion:
This PR will allow consuming commonjs modules which are located in node_modules, it doesn't yet do relative requires. After reading through the Browserify docs it seems we can have relative requires by passing a `basedir` configuration.

There would be no duplication of code in the bundle of polyfills as the whole bundle is ran through browserify, not each individual polyfill. 

This means:

> all polyfills in a bundle share a single copy of the shared components

wheresrhys:
sounds like an excellent move in general then

JakeChampion:
@wheresrhys I've added the ability to have relative requires however they all are relative from the same location which is the `polyfills` directory.

wheresrhys:
Is there anything blocking this getting merged. Once it has been I'll get to work polyfilling all those iterators

PR #722, opened 2016-06-03
nodelist and other dom iterators

wheresrhys:
So this should work. I need to research version numbers for the config

What are the test requirements for imported code? Simple test for existence of the prop on the prototype? @JakeChampion @triblondon 

I'm happy to take on array iterator too @callumlocke

### DIFF
--- a/polyfills/NodeList/prototype/@@iterator/config.json
+++ b/polyfills/NodeList/prototype/@@iterator/config.json
@@ -11,5 +11,9 @@
 		"Symbol.iterator"
 	],
 	"license": "MIT",
-	"spec": "https://dom.spec.whatwg.org/#nodelist"
+	"spec": "https://dom.spec.whatwg.org/#nodelist",
+	"repo": "https://github.com/medikoo/es6-iterator",
+	"build": {
+		"commonjs": true
+	}
 }
--- a/polyfills/NodeList/prototype/@@iterator/config.json
+++ b/polyfills/NodeList/prototype/@@iterator/config.json
@@ -1,0 +1,15 @@
+{
+	"browsers": {
+		"ie": "*",
+		"safari": "<9",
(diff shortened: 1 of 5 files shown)


polyfills/NodeList/prototype/@@iterator/config.json
```diff
@@ -0,0 +1,15 @@
...
+		"chrome": "< 50",
+		"firefox": "< 40",
+		"ios_saf": "*",
+		"opera": "< 50"
+	},
+	"dependencies": [
+		"Symbol.iterator"
+	],
+	"license": "MIT",
+	"spec": "https://dom.spec.whatwg.org/#nodelist"
```
JakeChampion: need to add this:

```
"repo": "https://github.com/medikoo/es6-iterator",
"build": {
    "commonjs": true
},
```

polyfills/NodeList/prototype/@@iterator/config.json
```diff
@@ -0,0 +1,19 @@
...
+	},
+	"dependencies": [
+		"Symbol.iterator"
+	],
+	"license": "MIT",
+	"spec": "https://dom.spec.whatwg.org/#nodelist",
+	"repo": "https://github.com/medikoo/es6-iterator",
+	"build": {
+		"commonjs": true
+	}
```
JakeChampion, line 18: You'll also need to declare the ECMAScript version of the polyfill as you are using `const`.
If you add `"esversion": 6` to this file then the service will run the polyfill through babel before creating the bundle for the browser.

triblondon:
Source code currently fails to parse in CI:

```
Warning: Error parsing source code for NodeList.prototype.@@iterator feature detect from /home/ubuntu/polyfill-service/polyfills/NodeList/prototype/@@iterator/detect.js Use --force to continue.
```

PR #723, opened 2016-06-05
Implement document.registerElement

### LINKED ISSUE #429: Add document.registerElement() polyfill
Andrea Giammarchi's excellent [document-register-element](https://github.com/WebReflection/document-register-element) polyfill works as advertised, though integrating it might be a bit more complicated because it has some [serious IE8 caveats](https://github.com/WebReflection/document-register-element#about-ie8).

alexwilson:
Closes #429 - though without IE8 support (yet).

While researching this, I found that actually `document.registerElement()` was removed from the spec in favour of `window.customElements.define`, which could have bigger implications elsewhere (ref. https://github.com/w3c/webcomponents/commit/d95392f734895b2c72e1822f464dd70a22b322a4 and later https://github.com/w3c/webcomponents/commit/815910d197077523fb449dfb5a1e6fceab728019).

Does implementing this still make sense, since it is already in-use in some Origami components, or should we skip and go straight for `window.customElements.define`?  (As far as I can tell nothing has implemented the changes yet).

Lemme know if this is a question to raise elsewhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "babel-core": "^6.2.1",
     "babel-preset-es2015": "^6.1.18",
     "batch": "^0.5.1",
+    "document-register-element": "^0.5.4",
     "fastly": "Financial-Times/fastly#v2.5.0",
     "grunt": "~0.4.0",
     "grunt-cli": "~0.1.6",
--- a/polyfills/document.registerElement/config.json
+++ b/polyfills/document.registerElement/config.json
@@ -1,0 +1,22 @@
+{
+	"browsers": {
+		"ie": "9 - *",
+		"ie_mob": "10 - *",
+		"firefox": "14 - 31",
+		"opera": "15 - 25",
+		"safari": "*",
+		"chrome": "* - 35",
+		"ios_saf": "*",
+		"firefox_mob": "4 - 37",
+		"bb": "*",
+		"android": "* - 4.4.4"
+	},
+	"build": {
+		"commonjs": true
+	},
+	"dependencies": ["html5shiv"],
+	"license": "MIT",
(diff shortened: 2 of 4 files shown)


triblondon:
Wow, we're using registerElement in origami?  Where?!

If this is no longer on standards track I would be against putting it in the polyfill service unless to not do so would cause issues for FT

triblondon:
@antoligy did you get anywhere looking into this?  I'm tempted to close if registerElement is dead.  From the diffs it looks like `defineElement` still lives, but I haven't read all the changes in detail so I don't know where `customElements.define` sits in relation to `defineElement`.  Maybe this is something to defer for now - I've no objection to hosting polyfills for features that are not yet stable, but it's a lot of work to keep up.

alexwilson:
Yep, there's an [o-share web-component](https://github.com/Financial-Times/o-share/blob/master/src/js/share.js#L244-L253) which prompted me to start looking for this polyfill!

For now it seems as though the usage of registerElement in Origami comes with feature detection, and I doubt that even with that it's seeing widespread use (though this is a massive assumption, to confirm I'll ask around on Slack tomorrow).

As discussed in https://github.com/WebReflection/document-register-element/issues/58, `customElements.define` is the final (stable?) version of this API and this has already been upstreamed into HTML, so we should be safe to use that going forward.

I'm happy to close this PR for now, and will raise another at a later date with something more concrete.

triblondon:
Ah yes, o-share is a testing ground for web components, but I don't think we intended the WC version to actually be used in production.  I think we're a fair way off that.  If you want to contribute a `customElements.define` method it would be gratefully received!

PR #724, opened 2016-06-05
Import Github's whatwg-fetch via npm, and update to v1.0.0.

### LINKED ISSUE #710: Upgrade Fetch polyfill
Please upgrade Fetch polyfill. I have issue in IE11.

https://github.com/github/fetch

alexwilson:
Closes #710.


triblondon:
Hi Alex, thanks for this!

You need to turn on the commonjs build option in config.json (build.commonjs = true) to make commonjs requires work correctly.

triblondon:
Incidentally, I thought you had push rights on FT repos. If you do, can you create a branch local to this fork so that the CI can run?  

(It would also be very helpful if you could check and update the browser support!)

alexwilson:
Hey - Just updated with commonjs set to true.

Sure, I'll take a look at browser support - Watch this space and all that!

No sorry, I don't have push access to FT Github, only BitBucket - I'll ask around to see if that's something that can be changed.

triblondon:
I opened a local branch for this in #728 so we can use CI.  There's a problem with IE7 and 8, but it seems to be originating in our `Document` polyfill, not the fetch one.

alexwilson:
🎉

PR #726, opened 2016-06-06
Rhys/array iterators

wheresrhys:
A few things arising from this PR
- it seems to require node 6 to run locally now, but not documented (it's in package.json, but node doesn't necessarily give the best feedbak for this - I suggest a 'requirements' section on the contributing page)
- npm shrinkwrap file messed up install for me so I deleted it. I can't recreate it as there are now lots of extraneous package errors.
- the ‘has feature detect’ test is a bit confusing as what it actually tests for is ‘has feature detect _and_ the feature has been polyfilled’. Should be 2 [albeit very similar] tests I think
- even without installing 'es6-iterator' the build completed ok, but I got a runtime error so maybe the commonjs implementation needs tweaking?
- ~~I _think_ array.prototype.values is bringing in the array.prototype.@@iterator polyfill in chrome, which it shouldn’t - it’s declared a s a dependency in config.json, but is available natively. This may just be my misinterpretation of how the test harness works though~~ `mode = targeted` 

@JakeChampion @triblondon


wheresrhys:
It fails in IE because of lack of support for Function.prototype.name. It will always fail in ie 6 & 7 for this reason as not polyfillable here.

So I propose setting the name property in the polyfill, adding a dependency on a  Function.prototype.name polyfill to ensure spec compliace as far back as possible, and either
a) removing the tests for iterator names 
b) removing the ability to serve these polyfills to ie6 - 8

triblondon:
@wheresrhys regarding your [comment](https://github.com/Financial-Times/polyfill-service/pull/728#issuecomment-225525572), the issue i'm having with commonjs builds is that something in the browserify wrapper code is interfering with some of our very-legacy polyfills, potentially because it is already polyfilling those things.  I didn't get any further than confirming that reverting to copy-pasting the source code into the polyfill.js file in the case of `fetch` fixes the problem.

But it seems worth being aware of the fact that browserifying a bundle might cause issues in polyfills that don't themselves require browserifying.  I'm a bit worried about this, and how we aren't really testing those scenarios, so I'm leaning towards backing out the commonjs support unless we can control these risks more effectively.

wheresrhys:
When it comes to polyfilling iterators, unless we have some way of sharing
code between polyfills there'll be a big hit to the bundle size because,
aside from about 3-4 lines of code, each polyfill is identical and not
tiny. So investigating and solving this problem is well worth it. I might
spend some time on it later this week.

Aside from that, is it worth marking polyfills as beta for the first month
or two of their existence?
On Wed, 15 Jun 2016 at 2:45 am, Andrew Betts notifications@github.com
wrote:

> @wheresrhys https://github.com/wheresrhys regarding your comment
> https://github.com/Financial-Times/polyfill-service/pull/728#issuecomment-225525572,
> the issue i'm having with commonjs builds is that something in the
> browserify wrapper code is interfering with some of our very-legacy
> polyfills, potentially because it is already polyfilling those things. I
> didn't get any further than confirming that reverting to copy-pasting the
> source code into the polyfill.js file in the case of fetch fixes the
> problem.
> 
> But it seems worth being aware of the fact that browserifying a bundle
> might cause issues in polyfills that don't themselves require
> browserifying. I'm a bit worried about this, and how we aren't really
> testing those scenarios, so I'm leaning towards backing out the commonjs
> support unless we can control these risks more effectively.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> https://github.com/Financial-Times/polyfill-service/pull/726#issuecomment-226066461,
> or mute the thread
> https://github.com/notifications/unsubscribe/AAbUR-PNtkMty4TU1Q1thJleQ_xT1Badks5qL1lTgaJpZM4Iuws9
> .

## 

---

_This email was sent by a company owned by Financial Times Group Limited 
("FT Group http://aboutus.ft.com/corporate-information/#axzz3rajCSIAt"), 
registered office at Number One Southwark Bridge, London SE1 9HL.  
Registered in England and Wales with company number 879531. This e-mail may 
contain confidential information. If you are not the intended recipient, 
please notify the sender immediately, delete all copies and do not 
distribute it further.  It could also contain personal views which are not 
necessarily those of the FT Group.  We may monitor outgoing or 
incoming emails as permitted by law._

triblondon:
We do have a solution to that - private polyfills, essentially shared dependency functions.  See https://cdn.polyfill.io/v2/docs/contributing/authoring-polyfills#update-tasks, specifically the bullet point "private dependencies"

wheresrhys:
woo! cheers for finishing this off. I may tackle DOM iterators next week

hannesj:
I'd like to add targeting for ie_mob for these before the 3.9 is cut out. I'll try to test and do a PR tomorrow.

JakeChampion:
@hannesj How are you getting on with adding ie_mob to these?

hannesj:
@JakeChampion Sorry, I won't manage it today. Would it be ok to add the targeting after the rc has been cut? Otherwise we can wait for 3.10 if needed.

PR #728, opened 2016-06-08
Update fetch, closes #724, closes #710

### LINKED ISSUE #710: Upgrade Fetch polyfill

triblondon:
Local for #724.

Note that this update removes the copyright statement from the top of the polyfill source file. This is not intentional on our part - that text has also been removed from the upstream repo's copy.


wheresrhys:
Now that jake has added commonjs support shouldn't we just require whatwg-fetch?

triblondon:
@JakeChampion potentially bad news about commonjs builds.  In debugging this issue, I found that something in browserify's boilerplate is interfering with our `Document` constructor polyfill in IE8.

It's possible that when browserifying the bundle we need to make changes to some of the fundamental constructor polyfills like `Document`.  But in the meantime I copied the latest tagged source from the fetch repo and it now parses in IE8 without errors (our test suite is just a feature detect, but no errors are thrown)

Interestingly, the only other commonjs-build polyfill we have currently in master is HTMLPictureElement, so I ran that test in IE8 and it passes - possibly because it has no dependency on Document, but possiblly because browserify seems to be outputting far less boilerplate.  For the fetch polyfill it appears to need to add polyfills for process.nextTick, for example.

Seems like we need to be quite careful with commonjs builds, because they raise the possibility that polyfills that work on their own might not work when in combination with another polyfill that causes the whole bundle to be browserified.

wheresrhys:
This is also failing in ie8 and is a commonjs polyfill https://github.com/Financial-Times/polyfill-service/pull/726. Could you provide me with steps to easily diagnose if it's suffering from the same issue

PR #737, opened 2016-06-21
Update Promise/Symbol/fetch polyfills for Edge

### LINKED ISSUE #736: Extra polyfills being served to Microsoft Edge
Howdy! 👋

I noticed that some of the supported features of Edge 13 (Promise, Symbol) and Edge 14 (fetch) are getting polyfilled when querying the API. E.g.:
- https://cdn.polyfill.io/v2/polyfill.js?features=Symbol,Promise,fetch

In Edge 14, that returns all three polyfills, whereas in Firefox 49 and Chrome 53 it returns nothing. In Edge 13, we get all three as well, even though Symbol and Promise are supported. 

This seems like it might be an issue with the IE/Edge divide, because features that debuted in IE11 are correctly being omitted in all versions of Edge (e.g. Intl).

If there's anything we on the Edge team could do to help out, please let me know. 🙂 I took a look at the code, but wasn't sure of the best place to start tackling it.

(This is polyfill-service v3.6.1 and it says: `UA detected: ie/14.14366.0`.)

nolanlawson:
Probably need to update some other APIs as well, but this is at least a start. Fixes #736.


polyfills/Promise/config.json
```diff
@@ -14,7 +14,7 @@
 		"bb": "* - 10",
 		"chrome": "* - 31",
 		"firefox": "6 - 28",
-		"ie": "8 - *",
+		"ie": "< 12",
```
triblondon: You're extending this polyfill to IE7 here.  I don't believe the polyfill works in IE7.

triblondon:
Since Edge 13 is a browser we already test in CI, support is actually already indicated on our feature matrix.  Looks like it works in 13, but not in IE7:

image

If you want to make the polyfill work in IE7, great, otherwise you probably want to start targeting at IE8.

To merge a change like this we normally ask for you to follow the simple steps in 
https://cdn.polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers. But in this case both browsers concerned are actively tested by our CI, so the testing steps are not required.

nolanlawson:
Whoops, my bad. I guess I hadn't thought about IE < 8.

Question, though: if the polyfills don't work at all in IE7, then why even specify it? Are people supposed to BYOP (bring your own polyfill :wink:) for IE <8?

triblondon:
Many of the polyfills do work in IE7, but this one doesn't.

polyfills/Promise/config.json
nolanlawson: D'oh, I guess I misunderstood what you're asking for. I had it as `< 12` before, but if the polyfill just doesn't work in IE7 (and Promises certainly don't exist in IE7), then what's the proper solution?

triblondon: At the lower end, you want to set the version constraint to the first version of the browser that is capable of applying the polyfill. This is currently set to 8 for Promise in IE, which seems to be right, since our CI shows it doesn't pass tests in IE7.  At the higher end, you want to set it to the last version of the browser that doesn't have the feature you're polyfilling, and that's where I think you want to make the change, right?

So you probably want `8 - 12` (which will include everything <13.0.0)

nolanlawson: Right, isn't that what I'm doing in 5e0fbcd6dbb38586a9689c65c8566763cb2fe82d?

JakeChampion:
Tests for this are running at - https://circleci.com/gh/Financial-Times/polyfill-service/312

JakeChampion:
It passed! 🎉

polyfills/Promise/config.json
triblondon: Oh, sorry, I didn't spot that commit

triblondon:
Great.  Thanks for this.  It will be released in 3.8.

nolanlawson:
Awesome, thanks for the help and for the great service! 💃

Issue #107, opened 2014-09-30
Self-learning configs


triblondon:
@jonathantneal suggested that our CI test suite on Sauce Labs could be used to determine which browsers lack support for features, and in which subset of those the polyfill works, and then update our polyfill configs automatically.

maxnordlund:
This could (should?) be combined with live data from users of the service. This means it would adapt as new browser versions come out.

triblondon:
The idea of beaconing RUM results back to the service was raised as part of https://github.com/Financial-Times/polyfill-service/issues/366 and concerns were raised about security.  I think initially we could simply add to the compatgen grunt task to get it to output a list of inconsistencies between the compat test results and the current targeting.

triblondon:
Prompted by a discussion on #731 and the increasing prevalence of "X isn't served on Y" type issues, and the general reluctance of those reporting those issues to write the one line of JSON required to fix them (whoa, passive aggressive mode enabled by accident, sorry. OK, carry on), I think it is worth resurrecting the idea of self-learning target config via RUM.  Here's an idea:
1. New query param `beacon=1` will activate beacon functionality
2. When loading a polyfill from disk, we initialise an empty `beaconData` property, which is a data tree organised as `.beaconData.<uaFamily>.<uaVersion>.passCount` and `.beaconData.<uaFamily>.<uaVersion>.failCount`
3. When serving a request that includes `beacon=1`, we choose up to `n` polyfills at random from those in the produced bundle, and _prefix_ the bundle with all the detects for those polyfills.  This creates a pass/fail for each polyfill against the browser's native capabilities.  Schedule a beacon using a legacy-compatible `new Image` technique
4. New API endpoint `/v2/reportBeaconData` which takes a query param for each feature.  It then updates the pass and fail counts for that feature in the browser that made the request, and responds with a lovely 1x1px gif.

Risks and complications:
- **Idiots**: We might need to do something to prevent people from just firing millions of fake beacon requests at the service to intentionally pollute the data.  The only thing I can think of is recording the IP address of the beacon source, and requiring a decent variety of IP sources (maybe 1000 unique IPs) before we trust the data.
[...]

PR #738, opened 2016-06-21
Routes

triblondon:
As a precursor to implementing #107 (self-learning configs) which will require new route endpoints, I wanted to refactor to separate the various types of endpoint we already have into modules since the main index.js is now getting quite large.

This update also includes a few miscellaneous fixes:
- Linting
- Fix navigation links to subsections of the contributors guide
- Spelling mistake in docs
- Clearer badging of dev mode output so that previous release number does not continue to be shown after subsequent commits
- Remove some debug
- Add a handler for uncaughtExceptions which otherwise print unhelpful output in more recent versions of Node

### DIFF
--- a/service/routes/docs.js
+++ b/service/routes/docs.js
@@ -5,13 +5,13 @@ const path = require('path');
 const request = require('request-promise');
 const Handlebars = require('handlebars');
 const moment = require('moment');
-const sources = require('../lib/sources');
+const sources = require('../../lib/sources');
 const marked = require('marked');
 const zlib = require('zlib');
-const PolyfillSet = require('./PolyfillSet');
-const polyfillservice = require('../lib');
-const compatdata = require('../docs/assets/compat.json');
-const appVersion = require(path.join(__dirname,'../package.json')).version;
+const PolyfillSet = require('../PolyfillSet');
+const polyfillservice = require('../../lib');
+const compatdata = require('../../docs/assets/compat.json');
+const appVersion = require(path.join(__dirname,'../../package.json')).version;
 
 const cache = {};
 const cachettls = {fastly:1800, respTimes:1800, outages:86400};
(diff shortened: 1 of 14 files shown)


service/routes/docs.js
```diff
@@ -297,12 +299,10 @@ function route(req, res, next) {
 		.then(locals => {
 			template(locals.pageName)
 				.then(templFn => res.send(templFn(locals)))
-				.catch(function() { next(); })
+				.catch(() => next())
```
JakeChampion, line 302: I think we can pass `next` in to catch directly rather than wrapping in another function.

`.catch(next)`

service/routes/test.js
```diff
@@ -77,6 +83,10 @@ function createEndpoint(type, polyfillio) {
 	};
 }
 
-module.exports = {
-	createEndpoint: createEndpoint
-};
+router.use('/libs/mocha', express.static(path.join(__dirname, '/../../node_modules/mocha')));
```
JakeChampion: I think `require.resolve('mocha')` does the same job as `path.join(__dirname, '/../../node_modules/mocha')`

service/routes/test.js
```diff
@@ -77,6 +83,10 @@ function createEndpoint(type, polyfillio) {
...
+router.use('/libs/expect', express.static(path.join(__dirname, '/../../node_modules/expect.js/')));
```
JakeChampion: Same here, believe this can be replaced with `require.resolve('expect.js')`

JakeChampion:
Tested locally, passes.

service/routes/docs.js
triblondon, line 302: I think if we do that, the error will cause express to bail.  If you pass an object to `next` it tells express to skip all remaining route middleware and jump to the first error middleware.

service/routes/test.js
triblondon: Oooooooooooooooooooo

triblondon: Ah, except it doesn't quite.  It gives you the main file.

triblondon:
Updated the resolve thing. OK with that and the `next` thing?

JakeChampion:
Yup, LGTM 👍

PR #743, opened 2016-06-28
Real user monitoring

triblondon:
This is a work in progress to create a beacon sending back live measurement data from active users of the service, for two purposes:
1. Collect resourcetiming data to evaluate performance more accurately than a simple response time reported by Pingdom, which appears to be highly variable between Pingdom POPs.
2. Check that our detects are correct by evaluating detects for all polyfills that we include in the bundle

I chose to use a MySQL backend for recording the data because the statistical work that needs to be done would quickly use up all the memory allocated to the process, and we intended to add a data storage backend at some point anyway.

Current concerns:
- The detects included in the bundle are only for the polyfills included.  This will capture false positives in our config, ie browsers that we serve to without needing to, but will not capture false negatives, ie browsers that we don't serve to even though they have no native impl and we have a polyfill.  We might need to compose a separate list of features for the RUM detects - maybe even just a random sample of features rather than anything related to the features that we are including polyfills for in the bundle.

You need a MYSQL_DSN env var to run this, and a database set up with the relevant schema, which is included in the root.


triblondon:
Jake - I've got more to do here but interested in what you make of it so far.  Also paging @wheresrhys

service/routes/rum.js
```diff
@@ -0,0 +1,103 @@
...
+ *  - dns: perf.domainLookupEnd - perf.domainLookupStart
+ *  - connect: perf.connectEnd - perf.connectStart
+ *  - req: perf.responseStart - perf.requestStart
+ *  - resp: perf.responseEnd - perf.responseStart
+ */
+router.get('/v2/recordRumData', (req, res) => {
+
+	res.set('Cache-Control', 'no-store');
+
+	const logquery = 'INSERT INTO requests SET perf_dns=?, perf_connect=?, perf_req=?, perf_resp=?, ua_family=?, ua_version=?, lat=?, lng=?, country=?, data_center=?, refer_domain=?';
```
DavidBruant, line 36: I'd recommend using a string builder like https://www.npmjs.com/package/sql
It usually makes SQL queries more readable and maintainable.

service/routes/rum.js
```diff
@@ -0,0 +1,125 @@
...
+const polyfillio = require('../../lib');
+const metrics = require('../metrics');
+const express = require('express');
+
+let mysql;
+if (process.env.MYSQL_DSN) {
+	require('promise-mysql').createConnection(process.env.MYSQL_DSN).then(conn => {
+		mysql = conn;
+	});
+}
```
JakeChampion, line 20: If the connection fails to be made, should we catch that failure and either crash the process with an informative error or write to stdout/logs that it failed and carry-on?

I think a combination of the two would work best. Write to stdout and carry-on if not in production, crash process with informative error if in production.

service/routes/rum.js
```diff
@@ -0,0 +1,125 @@
...
+ *  - resp: perf.responseEnd - perf.responseStart
+ */
+router.get('/v2/recordRumData', (req, res) => {
+
+	res.set('Cache-Control', 'no-store');
+
+	const logquery = 'INSERT INTO requests SET perf_dns=?, perf_connect=?, perf_req=?, perf_resp=?, ua_family=?, ua_version=?, lat=?, lng=?, country=?, data_center=?, refer_domain=?';
+	const referer = (req.get('referer') || '').replace(/^(https?:\/\/)?(www\.)?(.+?)(\:\d+)?([\/\?].*)?$/, '$3');
+	const ua = polyfillio.normalizeUserAgent(req.get('user-agent')).split('/');
+	const logdata = [req.query.dns, req.query.connect, req.query.req, req.query.resp, ua[0], ua[1], req.get('geo-lat'), req.get('geo-lng'), req.get('geo-country'), req.get('data-center'), referer];
```
JakeChampion, line 39: Would this work without Fastly or another CDN sat in-front of the application?

service/routes/rum.js
```diff
@@ -0,0 +1,125 @@
...
+				mysql.query('SELECT data_center, country, (SELECT perf_dns FROM requests r2 WHERE data_center=r1.data_center AND country=r1.country AND '+daterange+' ORDER BY perf_dns DESC LIMIT ?, 1) as perf_dns_95 FROM requests r1 WHERE '+daterange+' GROUP BY data_center, country', [Math.floor(count*0.05)]),
+				mysql.query('SELECT data_center, country, (SELECT perf_dns FROM requests r2 WHERE data_center=r1.data_center AND country=r1.country AND '+daterange+' ORDER BY perf_dns DESC LIMIT ?, 1) as perf_dns_99 FROM requests r1 WHERE '+daterange+' GROUP BY data_center, country', [Math.floor(count*0.01)])
+			]);
+		})
+		.then(results => {
+			const makeKey = row => row.country + ' ' + row.data_center;
+			const data = results.reduce((acc, resultset) => {
+				resultset.forEach(row => {
+					acc[makeKey(row)] = Object.assign({}, acc[makeKey(row)], row);
+				})
```
JakeChampion: missing semi-colon here.

service/routes/rum.js
```diff
@@ -0,0 +1,125 @@
...
+			})
+		;
+	}
+
+	res.status(200);
+	res.set('Content-type', 'image/gif');
+	res.send(blankGif);
+});
+
+router.get('/v2/getRumPerfData', (req, res) => {
```
JakeChampion, line 65: Will this route get cached by the CDN?

service/routes/rum.js
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * GET /v2/recordData
+ *  - <featureName>: 1 if it passed the detect, else 0
+ *  - dns: perf.domainLookupEnd - perf.domainLookupStart
+ *  - connect: perf.connectEnd - perf.connectStart
+ *  - req: perf.responseStart - perf.requestStart
+ *  - resp: perf.responseEnd - perf.responseStart
+ */
+router.get('/v2/recordRumData', (req, res) => {
```
JakeChampion, line 34: This route shouldn't be cached by the CDN, have we thought about how we will handle all the traffic when we come to turn this on by default?

service/routes/rum.js
triblondon, line 36: Thanks.  I'll stick with this for now but if it does get unwieldy I'll switch to something like that.

service/routes/rum.js
triblondon, line 20: I think just throwing in all situations is probably best.  If you don't want a mySQL connectionn you can just not set the DSN

service/routes/rum.js
triblondon, line 39: Yes, it works when I run it locally.  The DC and geoip stats just get set to null.

service/routes/rum.js
triblondon, line 65: Yes.  Added specific cache headers to these endpoints to tune that.

service/routes/rum.js
triblondon, line 34: No, but I think we're a fair way away from doing that.  One possibility is to run the RUM recording and reporting endpoints as a separate Heroku app, which would enable us to scale it separately, push releases with much lower QA overhead, and also prevent failures in the RUM code from affecting the main app.  But what we have should be fine for the moment.

PR #744, opened 2016-06-28
Revert "Add commonJS support for external modules (#719)"

JakeChampion:
This reverts commit 3ba1a534d99e6ae33fb364983f487277e2fbb3d7.

Reasons being Browserify is causing issues with our detect scripts by changing the context of execution from `Window/Self/Global` to an `exports` object.


AlbertoElias:
👍

PR #753, opened 2016-07-08
Update README.md

oguzzkilic:
a small fix.


JakeChampion:
That shouldn't be required, why did you need to cd into the installed dependency?

oguzzkilic:
@JakeChampion

PR #698, opened 2016-05-13
update intl to latest version

JakeChampion:

PR #754, opened 2016-07-08
Add "android" browser to Intl locales

Turbo87:
This should hopefully fix the issue reported in https://github.com/Financial-Times/polyfill-service/pull/613#issuecomment-231381636

It seems that the same thing was done in https://github.com/Financial-Times/polyfill-service/pull/698 but that PR also contains other changes which seem to take longer to land.

### DIFF
--- a/polyfills/Intl/~locale/af-NA/config.json
+++ b/polyfills/Intl/~locale/af-NA/config.json
@@ -1,5 +1,6 @@
 {
 	"browsers": {
+		"android": "<=4.3",
 		"ie": "<11",
 		"ie_mob": "10",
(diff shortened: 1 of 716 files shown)


triblondon:
Thanks!

Turbo87:
@triblondon any idea when this change will go live?

triblondon:
It'll be a while. We're about to release 3.8 to live.  Expect 3.9 in a few weeks.

Turbo87:
hmm, no chance for a quick bugfix release? I guess the only way around it for now is using `flags=always`

PR #760, opened 2016-07-14
Symbol polyfill: Resolve cross-domain issue on IE11

JakeChampion:


rowanmanning:
Code looks good, but I'm not sure what the bug was. Fancy explaining when you're in?

triblondon:
Agree, an explanation of the bug would be useful.

JakeChampion:
This bug appears on pages which make use of iframes, an example would be `ft.com` which has ~29 iframes on it.

It appears that somehow IE11 uses the patched Object inside the iframe and can no longer access the code which originated from a different domain.

triblondon:
> IE11 uses the patched Object inside the iframe

What.

I can't comment on this extensively, but it looks OK as a defence against that behaviour.

JakeChampion:
I had a thought - This should probably be a part of the `Object.getOwnPropertyNames` polyfill and configuring that polyfill to target IE11. What do you think?

PR #762, opened 2016-07-15
Simplify installing third-party polyfills from NPM

JakeChampion:
Polyfills from NPM can now be added to the polyfill-service by adding fields to the `config.json` file for the polyfill.

There is now an `install` object which contains all the information on how to install the package into the polyfill-service. The properties of this object are as follows:
- _module_*: The name of the package on npmjs to install 
- _paths_: The paths of the files to combine together to create the polyfill. If not included, defaults to the packages default entry point.
- _postinstall_: The path to a javascript module to execute after creating the polyfill. This can be used to create any helper files for the polyfill. An example would be the localisation data files used with the `Intl` polyfill.

_Properties denoted with (*) are required_.


triblondon:
Great news that IO is now available from the npm registry.

I have a few niggles with this approach:
- installpolyfill is only used as part of tasks, but it's in lib.  It would be better in a subfolder of tasks, I think.
- Do we need installpolyfill to be a separate module, rather than just improving the update task?  Maybe we could consider putting some config in the config.json file for a common install pattern, eg an `install` section that contains something like `{"module":"intersection-observer", "paths": [...], "postinstall":"update.task.js"}`.  This would cause the generic update task to install the module, concat the files in `paths` together and copy the result to the `polyfill.js` file.  If necessary, a local script could run to do anything specific either pre or post install.
- The installpolyfill script is more functional than the rest of the polyfill service code. I'm not opposed to FP, but using eg. `ary` instead of a closure, and `sum` to conat strings is not intuitive to me. May be worth asking the rest of the Origami team to make sure the code is easy enough to maintain.

JakeChampion:
> Great news that IO is now available from the npm registry.
> 
> I have a few niggles with this approach:
> - installpolyfill is only used as part of tasks, but it's in lib.  It would be better in a subfolder of tasks, I think.

Good point, I'll move it there. I thought I should move it as it no longer makes a reference to Grunt however it is still used as a build-step, makes more sense to be in the tasks directory.

> - Do we need installpolyfill to be a separate module, rather than just improving the update task?  Maybe we could consider putting some config in the config.json file for a common install pattern, eg an `install` section that contains something like `{"module":"intersection-observer", "paths": [...], "postinstall":"update.task.js"}`.  This would cause the generic update task to install the module, concat the files in `paths` together and copy the result to the `polyfill.js` file.  If necessary, a local script could run to do anything specific either pre or post install.

I like this idea, if I've read this correctly it would also mean the polyfill's source code would no longer be in our repository as it would be pulled in when running `npm install`. Is that what you had in mind?

> - The installpolyfill script is more functional than the rest of the polyfill service code. I'm not opposed to FP, but using eg. `ary` instead of a closure, and `sum` to conat strings is not intuitive to me. May be worth asking the rest of the Origami team to make sure the code is easy enough to maintain.

This is true, I'll refactor those parts of the code.

triblondon:
> it would also mean the polyfill's source code would no longer be in our repository as it would be pulled in when running npm install

Ah, no, I hadn't thought of going that far actually.  I suppose with shrinkwrap, this would be OK, but I hate shrinkwrap and was thinking that we could now get rid of it - our deployment process now ships the same container to prod as the one that we test on staging, so there's no possibility of a change in deps.  I guess that possibility still exists between CI and the QA deploy though.

The problem with shrinkwrapping is that nothing ever gets updated, plus the shrinkwrap file appears to change based on the individual developers' npm client version and settings, which means the diff is often huge and useless.

I'm willing to feel more positively about shrinkwrap if you want to pursue this, or we could otherwise keep the polyfill.js around.  Maybe in that case the config.json property should be called `update` rather than `install`.

JakeChampion:
@triblondon This is ready for another round of code reviewing.

package.json
```diff
@@ -39,10 +39,12 @@
     "blocked": "^1.1.0",
     "denodeify": "^1.2.1",
     "dotenv": "^1.2.0",
+    "exists-sync": "0.0.3",
```
triblondon, line 42: This seems almost as inconsequential as left-pad.  As long as it doesn't have lots of dependencies, fine.

polyfills/HTMLPictureElement/config.json
```diff
@@ -11,9 +11,11 @@
 		"bb": "7 - 10",
 		"android": "* < 46"
 	},
-	"dependencies": [],
+	"dependencies": ["html5shiv"],
```
triblondon: official name for this is `~html5-elements`.  `html5shiv` is an alias.

polyfills/HTMLPictureElement/config.json
```diff
@@ -11,9 +11,11 @@
...
 	"license": "MIT",
 	"spec": "https://html.spec.whatwg.org/multipage/embedded-content.html#embedded-content",
 	"repo": "https://github.com/scottjehl/picturefill",
-	"docs": "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture"
+	"docs": "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture",
+	"module":"picturefill",
```
triblondon: I think `module` and `paths` should be grouped under something like `install` or `update`, like we do with `build`.

polyfills/IntersectionObserver/config.json
```diff
@@ -14,11 +14,15 @@
...
 		"Array.prototype.some",
 		"Function.prototype.bind"
 	],
 	"docs": "https://github.com/WICG/IntersectionObserver/blob/master/explainer.md",
 	"spec": "http://rawgit.com/WICG/IntersectionObserver/master/index.html",
-	"notes": []
+	"notes": [],
+	"repo": "https://github.com/WICG/IntersectionObserver",
+	"module":"intersection-observer",
+	"paths": ["intersection-observer"]
```
triblondon: As above, I'd group these under a single property, possibly `update`, if we are keeping the local source.

polyfills/Intl/update.task.js
```diff
@@ -1,32 +1,43 @@
 'use strict';
```
triblondon, line 14: I know it wasn't there before, but a brief header comment describing what the postinstall task is needed for would add some useful context.

polyfills/Intl/config.json
```diff
@@ -20,5 +20,8 @@
 	"docs": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl",
 	"notes": [
 		"Locales must be specified separately by prefixing the locale name with `Intl.~locale`, eg `Intl.~locale.en-GB`."
-	]
+	],
+	"module": "Intl",
+	"paths": ["dist/Intl.min.js"],
+	"postinstall": "update.task.js"
```
triblondon: consider renaming to `postupdate`?

polyfills/Intl/update.task.js
```diff
@@ -35,28 +46,29 @@ module.exports = function(grunt) {
...
 	configSource.test = { ci: false };
 
-	var configFileSource = JSON.stringify(configSource, null, 4);
+	const configFileSource = JSON.stringify(configSource, null, 4);
 
-	grunt.log.writeln('Importing Intl polyfill from ' + IntlSourcePath);
-
-	var IntlPolyfillSource = fs.readFileSync(IntlSourcePath);
-	grunt.file.write(path.join(IntlPolyfillOutput, 'polyfill.js'), IntlPolyfillSource);
+	console.log('Importing Intl.~locale.* polyfill from ' + LocalesPath);
```
triblondon: No access to `grunt.log`?

polyfills/Intl/update.task.js
```diff
@@ -35,28 +46,29 @@ module.exports = function(grunt) {
...
+		writeFileIfChanged(configOutputPath, configFileSource);
 	});
 
 
-	grunt.log.writeln(locales.length + ' imported locales');
-	grunt.log.writeln('Intl polyfill imported successfully');
-
+	console.log(locales.length + ' imported locales');
+	console.log('Intl polyfill imported successfully');
 };
```
triblondon, line 85: This all looks like it basically does the same thing and you just refactored it to modernise the code?

tasks/updatelibrary.js
```diff
@@ -1,17 +1,64 @@
 'use strict';
 
+const fs = require('fs');
 const path = require('path');
+const crypto = require('crypto');
+const existsSync = require('exists-sync');
+function hash (contents) {
```
triblondon: I'd be tempted to call this function `md5()`

tasks/updatelibrary.js
```diff
@@ -1,17 +1,64 @@
...
+	return crypto.createHash('md5').update(contents).digest('hex');
+}
+
+function constructPolyfill(polyfillPaths) {
```
triblondon: Consider renaming to `loadSource`?

tasks/updatelibrary.js
```diff
@@ -1,17 +1,64 @@
...
+	const polyfills = polyfillPaths.map((polyfill) => fs.readFileSync(polyfill));
+	return polyfills.join('');
```
triblondon, line 13: This is a lot more obvious now, thanks.

tasks/updatelibrary.js
```diff
@@ -1,17 +1,64 @@
...
+
+const installPolyfill = function(polyfillOutputFolder, module, polyfillSourcePaths) {
+	const polyfillOutputPath = path.join(polyfillOutputFolder, 'polyfill.js');
+	const polyfillAlreadyExists = existsSync(polyfillOutputPath);
+
+	polyfillSourcePaths = polyfillSourcePaths || [''];
+	polyfillSourcePaths = polyfillSourcePaths.map(polyfillPath => path.join(module, polyfillPath))
+	polyfillSourcePaths = polyfillSourcePaths.map(path => require.resolve(path));
+
+	console.log(`Importing polyfill/s from ${polyfillSourcePaths.join(',')}`);
```
triblondon: use `grunt.log` if possible?
Consider starting the comment with our polyfill feature name, I can't see that we're outputting that otherwise.

tasks/updatelibrary.js
```diff
@@ -1,17 +1,64 @@
...
+  }
+
+	fs.writeFileSync(polyfillOutputPath, newPolyfill);
+	console.log('Polyfill imported successfully');
+
+};
 
 module.exports = function(grunt) {
 
 	grunt.registerMultiTask('updatelibrary', 'Update Polyfill imported from external libraries', function() {
```
triblondon: nit: polyfills, plural.

tasks/updatelibrary.js
```diff
@@ -1,17 +1,64 @@
...
-			file.src.filter(function(src) {
-				return src.indexOf('/__versions/') === -1;
-			}).forEach(function(src){
-				grunt.log.writeln('Running ' + src + ': ');
-				require(path.resolve(src.toString()))(grunt);
+			const configs = file.src.map(src => Object.assign({src}, {
+				config: JSON.parse(fs.readFileSync(src)),
+			}));
+			configs.filter(function(config) {
+				return typeof config.config.module === 'string';
```
triblondon: So here I think it's more appropriate to be looking for a section, like `update`

triblondon:
Review complete.  Feel free to merge it once you've taken those comments in. I don't need to review again (i'm also about to get on a plane so can't anyway)

polyfills/Intl/update.task.js
JakeChampion: We can pass grunt in to the exported function as a parameter, shall I do that?

polyfills/Intl/update.task.js
JakeChampion, line 85: Not quite, it now won't modify any files which have not changed, this is a good thing as before git would think every single file has been changed when in fact they had not.

tasks/updatelibrary.js
JakeChampion: 👍

package.json
JakeChampion, line 42: It has 0 dependencies.

polyfills/Intl/update.task.js
JakeChampion, line 14: Added.

Issue #467, opened 2015-08-21
ie8 issue SCRIPT445: Object doesn't support this action


thednp:
Please check this demo

http://thednp.github.io/bootstrap.native/

There seems to be an issue with the event polyfill at this line

```
eventElement.call(element, event);
```

I believe this is because there have been some changes to the default IE8 polyfills structure that's causing this issue, because some months ago I didn't have this issue.

Can you please investigate and fix the issue?

triblondon:
Often, these issues turn out to be problems in host site code.  To help us debug this problem, it would be appreciated if you could add a test to our test suite for this polyfill that fails under the conditions you describe.

Failing that, create a page that includes the polyfill service and just enough additional code to demonstrate the problem (ideally not more than a dozen lines of markup and JS)

thednp:
Well, simple open IE10+, switch to IE8 view and refresh page. But the strange/crazy thing is, that demo works flawless today. Do you have any idea what is causing this (sometimes it works sometimes gets some error)?

EDIT: I posted some demo on codepen and jsfiddle and none of them support IE8 :|

thednp:
Now it's not working again.. very strange!

Here's a page I made for you to play around.
http://thednp.github.io/bootstrap.native/test.html

triblondon:
Most code bins/pens have an output mode that doesn't load the editor UI so that avoids any browser support problems of the app itself.  I'm afraid we aren't going to dig through the minified version of your library but if you can extract the code that triggers the issue I'd be glad to take a look.

thednp:
Updated the test page, please check again. I don't really know what I should extract. The IE11 (in IE8 mode) console only shows this error, with no reference to what or where that code is used.

triblondon:
I'm afraid I can't reproduce the issue.

Again, I'd suggest reducing the problem.  You have thousands of lines of code in this page that is not coming from the polyfill service, and you're also loading multiple polyfills.  If you can find a problem state that requires only a few lines of code and a single polyfill, we'd be happy to look into that.

Alternatively, others may be willing to offer commercial support if you'd like someone else to do the debugging for you.

thednp:
After some time to work with, I figured the issue, I've logged the `element` and the `eventElement` and it was `HTMLDocument` with empty `eventElement`. Here is the fix:
- in the `while (++index < length && !event.cancelImmediate) {}` loop locate

```
if (indexOf(list, eventElement) !== -1) {
[...]

PR #767, opened 2016-07-19
Added callback type check in addEventListener to avoid exceptions

vstanyshevskyy:
I faced the same issue described in the [Issue 467](https://github.com/Financial-Times/polyfill-service/issues/467). Looks like it is common.


triblondon:
Thanks!  I'm confused about why that variable is called `eventElement` since it seems to be (expected to be) a handler function.  I'd be happy to merge this.  @JakeChampion could you run CI and merge if you're OK with it?

PR #773, opened 2016-07-22
Pass string to factory to avoid throwing errors in modern browsers

### LINKED ISSUE #772: IE10 in IE7 compatibility mode
I am seeing a console error in the document.querySelector polyfill for IE10 in IE7 compatibility mode.

This is the request being made: (`min` has been removed for debugging)
`https://next-geebee.ft.com/polyfill/v2/polyfill.js?callback=ftNextPolyfillServiceCallback&rum=1&features=default,requestAnimationFrame,Promise,matchMedia,HTMLPictureElement,CustomEvent|always|gated,fetch|always|gated,Array.prototype.find|always|gated,Array.prototype.findIndex|always|gated,IntersectionObserver&excludes=Symbol,Symbol.iterator,Symbol.species,Map,Set`

The error from ie dev console:
`SCRIPT65535: Argument not optional`

The offending snippet:
https://github.com/Financial-Times/polyfill-service/blob/c347adb384b77b0844e9d09f25779bbe9ca13952/polyfills/document.querySelector/polyfill.js#L7

I have fixed the error locally by excluding the querySelector polyfill and using a local one which has the following difference/change: `generator = document.createElement('null')`

JakeChampion:
Resolves #772

### DIFF
--- a/polyfills/document.querySelector/polyfill.js
+++ b/polyfills/document.querySelector/polyfill.js
@@ -4,7 +4,7 @@
 
 	function getElementsByQuery(node, selector, one) {
 		var
-		generator = document.createElement(),
+		generator = document.createElement('null'),
 		id = 'qsa' + String(Math.random()).slice(3),
 		style, elements;
 
--- a/polyfills/document.querySelector/polyfill.js
+++ b/polyfills/document.querySelector/polyfill.js
@@ -4,7 +4,7 @@
 
 	function getElementsByQuery(node, selector, one) {
 		var
-		generator = document.createElement('null'),
+		generator = document.createElement('div'),
 		id = 'qsa' + String(Math.random()).slice(3),
 		style, elements;
 


polyfills/document.querySelector/polyfill.js
```diff
@@ -4,7 +4,7 @@
 
 	function getElementsByQuery(node, selector, one) {
 		var
-		generator = document.createElement(),
+		generator = document.createElement('null'),
```
triblondon: Is this valid?  Is it worth using a real element to avoid possible rarely tested edge cases?

JakeChampion: This is valid, it will create an `HTMLUnknownElement` instead of a `HTMLElement`. 
Pasted from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLUnknownElement):

> The HTMLUnknownElement interface represents an invalid HTML element and derives from the HTMLElement interface, but without implementing any additional properties or methods.

triblondon: Sure, just wondering if that might be more likely to run into unknown browser bugs.  Not a strong point.

JakeChampion: Possibly, what element would you suggest as an alternative?

triblondon: div?

JakeChampion:
Updated this PR to use a `div` element

PR #774, opened 2016-07-24
Updated localStorage detect.js for Safari private browsing mode

lelledev:
Hi,

On Safari private browsing Polyfill detect the `localStorage` availability with `'localStorage' in this` because `localStorage` exist and is an object. But Safari private browsing limit the localStorage session to 0Mb and when try to set a variable will show an error about the limit size.

With this changes Polyfill detects the Safari private browsing localStorage support.

Many thanks.

### DIFF
--- a/polyfills/localStorage/detect.js
+++ b/polyfills/localStorage/detect.js
@@ -1,1 +1,1 @@
-'localStorage' in this
+try { localStorage.setItem('localStorage', true); localStorage.removeItem('localStorage'); true; } catch (e) { false; }


triblondon:
@matthew-andrews you know about this issue, right?

matthew-andrews:
Hm, I do but I don't think it would be a good idea to try to polyfill in this case.

triblondon:
@matthew-andrews why not?

matthew-andrews:
hmm, i guess if the polyfill works it's ok, it's just that the function is there, the browser _thinks_ it can support it — it's just that it doesn't work.  are there other polyfills that fix broken implementations of things rather than just add missing ones?

triblondon:
Yes, we have quite a lot of those.  In this case I don't think the polyfill will maintain the data from one page view to the next, it's just an in-memory store.  Having said that, it does something with DOM elements that I don't really understand.

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/localStorage/polyfill.js

If it works well enough for normal environments, I see no reason not to allow it to be loaded in private browsing mode if it works there.  @lelledaniele have you tested it?

lelledev:
Hi

[I made this test and I tried It](http://codepen.io/anon/pen/WxKkKA) on Safari with/without private browsing.
Yes the data won't be maintain from one page view to the next but if now you use Polyfill on Safari private browsing will generate an error because `localStorage in this` answered `true` and all script will be blocked.

Many thanks.

safari
safari with private browsing

triblondon:
@lelledaniele to test this in private browsing mode you'd need to force-load the polyfill, using https://polyfill.io/v2/polyfill.js?features=localStorage&flags=always

lelledev:
Hi @triblondon,

if try force the Polyfill localStorage feature still have the same problem.
`QuotaExceededError: DOM Exception 22: An attempt was made to add something to storage that exceeded the quota.`
[Tested script link](http://codepen.io/anon/pen/ZOjXNa)

Many thanks

screenshot at 2016-07-30 10 44 44

triblondon:
So it looks like the polyfill does not override the native (lack of) implementation even when force-included.  It seems like that's because the polyfill includes a feature detect:

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/localStorage/polyfill.js#L73

I'm really uncomfortable with this polyfill, the more I look at it.  I have no idea why we're using a DOM element to hang properties on.  Does this give us some kind of persistence in some weird way that I'm not able to understand?  I'm tempted to swap out a simpler polyfill which doesn't include an inline detect, since the only real use case for a localStorage polyfill is private browsing.  Something like:

``` javascript
window.localStorage = (function() {
   var protected = ["setItem", "getItem", "removeItem", "clear"];
   return {
      setItem: function(id, val) { this[id] = String(val); },
      getItem: function(id) { return this.hasOwnProperty(id) ? this[id] : undefined; },
      removeItem: function(id) { (this.hasOwnProperty(id) && !protected.includes(id)) ? delete this[id] : undefined; },
      clear: function() { Object.keys(this).forEach(this.removeItem, this); }
   };
}());
```

Not tested.

triblondon:
I've done more of a review here.  The bottom line is:
- You cannot detect server side whether a browser is in private browsing mode (which is partly by design, so unlikely to ever change)
- Currently we don't target Safari with the localStorage polyfill because (outside of private browsing) it does support localStorage
- Our existing polyfill wouldn't take effect even if we did serve it in PBM because it contains a simplistic feature detect

So the best option we have here is:
- Rewrite the localStorage polyfill to incorporate a more intelligent feature detect
- Serve it in all versions of Safari regardless of PBM, and rely on the feature detect to avoid activating it outside of PBM

JakeChampion:
Hi @lelledaniele, are you wanting to carry on work with this PR or is it okay to close it?

lelledev:
Hi @JakeChampion, we can close It. I didn't find any solution because on Safari we cannot override localStorage. Many thanks.

PR #775, opened 2016-07-26
Add aliasing for FB in-app browser on iOS, fixes #561

### LINKED ISSUE #561: Facebook iOS app browser not supported
The Facebook native iOS app uses a `UIWebView` with a custom user-agent. Ideally Polyfill should return a polyfill for the current Safari/iOS version, but returns an error message.

Here is a polyfill request with a Facebook user-agent:

```
curl "https://cdn.polyfill.io/v2/polyfill.min.js?features=Object.assign,Promise,Array.prototype.find,Array.prototype.findIndex,Intl,Intl.~locale.nb&flags=always,gated" --header "User-agent: Mozilla/5.0 (iPhone; CPU iPhone OS 9_2 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Mobile/13C75 [FBAN/FBIOS;FBAV/46.0.0.54.156;FBBV/18972819;FBDV/iPhone8,1;FBMD/iPhone;FBSN/iPhone OS;FBSV/9.2;FBSS/2; FBCR/Telenor;FBID/phone;FBLC/nb_NO;FBOP/5]"
```

Returns:

```
/* Unsupported UA detected: facebook/46.0.0 */
```

triblondon:
Changes identification of FB iOS browser from facebook/80.0.0 to ios_saf/9.3.0.

Note that Android already spawns the standard OS browser from the Facebook app.

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -79,6 +79,11 @@ const aliases = {
 	"iphone": "ios_saf",
 	"iphone simulator": "ios_saf",
 	"mobile safari uiwebview": "ios_saf",
+	"facebook": function(ua) {
+		if (ua.os.family === 'iOS') {
+			return {family:'ios_saf', major: ua.os.major, minor:ua.os.minor};
+		}
+	},
 
 	"phantomjs": ["safari", 5],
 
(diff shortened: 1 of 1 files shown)


georgecrawford:
What's the situation for Android FB in-app browser? Do we know?

lib/UA.js
```diff
@@ -127,8 +132,12 @@ function UA(uaString) {
 		this.ua.family = this.ua.family.toLowerCase();
 		if (aliases[this.ua.family]) {
 
+			// Custom aliasing
+			if (typeof aliases[this.ua.family] === 'function') {
+				Object.assign(this.ua, aliases[this.ua.family](this.ua));
+
 			// Map to different family, use same version scheme
-			if (typeof aliases[this.ua.family] === 'string') {
+			} if (typeof aliases[this.ua.family] === 'string') {
```
JakeChampion: should this be `} else if (`?

triblondon: Sorry.  I should not file a PR in a hurry and then run to lunch.

triblondon:
@georgecrawford read the PR description :-)

leggsimon:
I am right in thinking that this would stop the FB in-app browser being "unknown" yes? And would just be dealt with as and when and we wouldn't need `&unknown=polyfill` to cover it?

PR #777, opened 2016-07-26
Enable Object.create(Function.prototype) to work

JakeChampion:
This throws with our polyfill but works in modern browsers and in es5-shim. This PR enables it to work with our polyfill.


rowanmanning:
Looks good to me 👍

PR #779, opened 2016-07-26
Minified reserved words

### LINKED ISSUE #778: IE8 console error on minfication
Have recently found that our call to the Polyfill service is producing a console error on IE8.

The call we make to Polyfill.io is the following URL:

`//cdn.polyfill.io/v2/polyfill.min.js?features=default,console,Array.prototype.includes,fetch|always,promise|always&flags=gated`

And in IE8 we get a console error of:

`Expected identifier – polyfill.min.js?features=default,console,fetch|always,promise|always&flags=gated, line 13 character 3132`

When taking off the minification, the console error disappears, so this must come about as a result of the minified code(?).  I’ve also tried reducing the call down to just have the default feature set and it still returns the console error, so it doesn’t seem to be caused by one of the additional feature requests.

Any help on this would be appreciated.

JakeChampion:
It seems Uglify-js 2.7.0 began to set `screw_ie8` to true by default which broke some of our minified polyfills in IE8.
This PR sets `screw_ie8` to false to enable IE8 support in all polyfills once again.

Resolves https://github.com/Financial-Times/polyfill-service/issues/778

### DIFF
--- a/tasks/buildsources.js
+++ b/tasks/buildsources.js
@@ -93,7 +93,12 @@ module.exports = function(grunt) {
 					config.minSource = config.rawSource;
 				} else {
 					validateSource(config.rawSource, featureName+' from '+polyfillSourcePath);
-					config.minSource = uglify.minify(config.rawSource, {fromString: true}).code;
+					config.minSource = uglify.minify(config.rawSource, {
+						fromString: true,
+						compress: { screw_ie8: false},
+						mangle:   { screw_ie8: false},
+						output:   { screw_ie8: false, beautify: false }
+					}).code;
 				}
 
 				if (fs.existsSync(detectPath)) {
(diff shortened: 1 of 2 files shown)


kzc:
> Revert "Run tests against minified code"

 Just curious - why aren't polyfill tests run against minified code?

JakeChampion:
Because we test for correct function names which would fail on minified code

kzc:
There's an uglify option `keep_fnames` for `mangle` and `compress` that might be of interest. Don't know how much bigger your minified code would be as result.

https://github.com/mishoo/UglifyJS2/commit/9edbe93df5633bd13b1c001d066887e011ce767c

https://github.com/mishoo/UglifyJS2/issues/1187#issuecomment-230349336

JakeChampion:
Yep, I saw that option but didn't want to implement it as it results in a bigger file size.

PR #780, opened 2016-07-26
Promises: Switch from Octane to Yaku

JakeChampion:
This PR switches our Promise polyfill from [Octane](https://github.com/Octane/Promise) to [Yaku](https://github.com/ysmood/yaku).

Yaku will report any unhandled rejection via console.error by default, in case you forget to write catch. You can also catch them manually:

Browser: `window.onunhandledrejection = ({ promise, reason }) => { /* Your Code */ };`


JakeChampion:
I added and then removed the promises a+ and promises es6 test suites from this PR as those tests require functions such as `Object.create`/`Object.key`/`[].forEach` which we would need to polyfill. Another reason for their removal is that Yaku runs all these tests in it's own repository meaning we don't need to here, this is in line with how we handle the tests for the third-party polyfill `Intl`.

triblondon:
👍

PR #783, opened 2016-07-27
Identify the polyfills needed for Samsung Internet 4

AdaRoseCannon:
I think this is correct, but would like confirmation.

I ran the control test on the device to identify what features were required.

I then added the samsung_mob to those configs.

### DIFF
--- a/docs/contributing/authoring-polyfills.html
+++ b/docs/contributing/authoring-polyfills.html
@@ -67,7 +67,8 @@
         "firefox": "&lt;=20",
         "opera": "11 || 14",
         "safari": "&lt;=4",
-        "ios_saf" "&lt;=6"
+        "ios_saf" "&lt;=6",
+        "samsing_mob": "*"
     },
     "aliases": [
         "modernizr:es5array"
@@ -165,6 +166,7 @@ module.exports = function(grunt) {
 			<tr><td><code>opera</code></td><td>Opera</td><td>{{baselines.opera}}</td></tr>
 			<tr><td><code>op_mob</code></td><td>Opera Mobile</td><td>{{baselines.op_mob}}</td></tr>
 			<tr><td><code>op_mini</code></td><td>Opera Mini</td><td>{{baselines.op_mini}}</td></tr>
+			<tr><td><code>samsung_mob</code></td><td>Samsung Internet</td><td>{{baselines.samsung_mob}}</td></tr>
 			<tr><td><code>bb</code></td><td>Blackberry</td><td>{{baselines.bb}}</td></tr>
 			</tbody>
 		</table>
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -31,7 +31,8 @@ const baseLineVersions = {
 	"opera": ">=11",
 	"op_mob": ">=10",
 	"op_mini": ">=5",
-	"bb": ">=6"
+	"bb": ">=6",
+	"samsung_mob": "*"
 };
 
 /* Aliases may be expressed in three formats:
@@ -79,6 +80,8 @@ const aliases = {
 	"iphone": "ios_saf",
 	"iphone simulator": "ios_saf",
 	"mobile safari uiwebview": "ios_saf",
+
+	"samsung internet": "samsung_mob",
 
 	"phantomjs": ["safari", 5],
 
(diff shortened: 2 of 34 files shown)


JakeChampion:
This should resolve #509 and #652

triblondon:
- No `Intl`?
- What's the naming rational for `samsung_mob` rather than just `samsung`?  Is samsung planning to make multiple browsers or would that at least be a viable possibility?  If not why do we need to be specific here?
- Missing docs in https://github.com/Financial-Times/polyfill-service/blob/ae4aa7477c211a699cf0b700c0124a976ad8b4ea/docs/contributing/authoring-polyfills.html
- Will Samsung be officially supporting this config for future versions?

AdaRoseCannon:
- Intl is there: polyfills/Intl/config.json
- samsung_mob is the term Jake used in an earlier PR I used the same name. Would you prefer me to change it?
-  will update the doc, thanks for noticing.
- Yes, the Samsung developer advocate team, will continue to support this. ('cause the polyfill service rocks)

JakeChampion:
@AdaRoseEdwards I thought Samsung Browser supported `Intl` natively, is that not the case?

AdaRoseCannon:
Not on the earlier versions of Samsung Internet, I am trying to get my hands on previous versions so I can add some '2-4' for features missing on just those versions. I will also specify Intl for just that range too. 

Is there a way to work return what the polyfill service is working out the platform/version as from parsing the UA string?

JakeChampion:
We use [useragent](https://www.npmjs.com/package/useragent) to give use browser versions. We also just recently added the ability to configure the versions within the [polyfill-service itself](https://github.com/Financial-Times/polyfill-service/pull/775/files#diff-6f0f32bc2b556db6b5795e1ee9170443R82)

AdaRoseCannon:
Looking at the docs: http://developer.samsung.com/technical-doc/view.do?v=T000000203 and testing the UAs with useragent. Samsung Browser is either Chrome 34 (Old), Or Samsung 4 (Current) 

So I will leave it at \* for now, until the next version of Samsung Internet gets released.

(@JakeChampion i'll test Intl again see if it isupported)

AdaRoseCannon:
Intl is not supported natively. Fails feature detect.

JakeChampion:
You can test what our service thinks the UA is, here is an example URL using the older Samsung UA:
`https://cdn.polyfill.io/v2/polyfill.js?features=Promise&ua=Mozilla/5.0%20(Linux;%20Android%204.4.4;%20en-au;%20SAMSUNG%20SM-N915G%20Build/KTU84P)%20AppleWebKit/537.36%20(KTHML,%20like%20Gecko)%20Version/2.0%20Chrome/34.0.1847.76%20Mobile%20Safari/537.36`

hannesj:
@AdaRoseEdwards Could you check support for [these](https://github.com/Financial-Times/polyfill-service/pull/726) newly added polyfills?

AdaRoseCannon:
Sure!

JakeChampion:
@AdaRoseEdwards When would we expect these to be added? We are wanting to cut a release by end of today if that is possible.

AdaRoseCannon:
I'll do it now!

AdaRoseCannon:
Array.prototype.@@iterator supported natively
Array.prototype.entries supported natively
Array.prototype.keys supported natively
X Array.prototype.values is not supported natively

JakeChampion:
Can you make a PR for this?

docs/contributing/authoring-polyfills.html
```diff
@@ -67,7 +67,8 @@ <h2 id='configuration'>Configuration</h2>
         "firefox": "&lt;=20",
         "opera": "11 || 14",
         "safari": "&lt;=4",
-        "ios_saf" "&lt;=6"
+        "ios_saf" "&lt;=6",
+        "samsing_mob": "*"
```
piuccio: this should probably be `samsung` not `samsing`

skv-headless:
@AdaRoseCannon thank you for the samsung browser support. Any plans for supporting 2 and 3 versions? We still have plenty of such users 😞

AdaRoseCannon:
Sorry there are no plans for supporting those versions.

PR #786, opened 2016-07-29
Fix Math.clz32

JakeChampion:
Before it was not using the parameter in the function but instead using `undefined` in all calculations. I've added tests from Test262 to cover the issue.


rowanmanning:
This looks sensible 👍

Issue #764, opened 2016-07-18
Expose URLSearchParams on window


rupesh1:
Do you think its possible to expose the existing URLSearchParams function (available privately to the URL polyfill https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/URL/polyfill.js#L133) on the window/global object?

Example usage as a global: 
https://googlechrome.github.io/samples/urlsearchparams/index.html
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

Spec: 
https://url.spec.whatwg.org/#urlsearchparams

triblondon:
Great idea.  This is a third party polyfill, so the best place to start is https://github.com/inexorabletash/polyfill, and if they're happy with the idea maybe raise a PR there?  If the polyfill is updated to expose URLSearchParams, we would update our copy, and add `URLSearchParams` as an alias of `URL`, since they are actually both independent web APIs but contained in one polyfill.

rupesh1:
Hi Andrew - thanks for the quick reply. It would appear that its been available in the source polyfill since this commit https://github.com/inexorabletash/polyfill/commit/fe3084974d83c909e7d9802fa8609aab5f93cd55#diff-72d74024ea723e098405b46585936940R5301

That commit is dated Nov 2015, maybe the reference to the source `URL` polyfill could just be updated in general (and we get this for free)?

Getting `URLSearchParams` with `URL` makes perfect sense and would be ideal for my use-case.

PR #800, opened 2016-08-03
Update to latest version of URL, use new install mechanism

triblondon:
See #764.


JakeChampion:
This seems to fail in IE 8/12/13 :

``` bash
 - ie 13 (Sauce results: https://saucelabs.com/tests/842735ead22d42cf9eaeb9424189bf2d)
    -> URL       http://127.0.0.1:3000/test/tests?mode=targeted&feature=URL
 - ie 12 (Sauce results: https://saucelabs.com/tests/49bbe4a2c26f4bedb589c07c2d18b30f)
    -> URL       http://127.0.0.1:3000/test/tests?mode=targeted&feature=URL
 - ie 8 (Sauce results: https://saucelabs.com/tests/99688b1ce0624bd1931ad3cd6e75cd95)
    -> URL       http://127.0.0.1:3000/test/tests?mode=targeted&feature=URL
```

triblondon:
Yep, will look at those.  I lazily relied on the PR to run CI rather than running it from my local machine first.

triblondon:
@JakeChampion success! I had to disable number of tests because of buggy native implementations in Firefox and edge, and because some of the features tested are not actually implemented by the polyfill.  But I also added a slew of extra tests by updating from the upstream repo. Ended up removing edge targeting because native support is almost perfect in edge, and adding iOS Safari targeting because URL is not supported on Safari on any platform.

Could you merge if you're happy with this, and tag for whatever your next release is?

polyfills/URL/tests.js
```diff
@@ -1,28 +1,35 @@
 /* eslint-env mocha, browser*/
 /* global proclaim, it */
 
+function toArray(it) {
```
JakeChampion, line 4: The tests which use this function are commented out, should we comment this out or remove it?

triblondon: I wanted to make as few changes to the tests as possible.  I don't really mind either of those options, or just leaving it as-is.

rowanmanning:
This looks good to me 👍

PR #805, opened 2016-08-05
Add array.of via new third party install method

JakeChampion:


polyfills/Array/of/config.json
```diff
@@ -21,5 +21,9 @@
 		"Object.defineProperty"
 	],
 	"spec": "http://people.mozilla.org/~jorendorff/es6-draft.html#sec-array.of",
-	"docs": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of"
+	"docs": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of",
+	"install": {
+		"module": "array.of",
+		"paths": ["array-of.js"]
+	}
 }
```
triblondon, line 31: Need to add `repo` and `licence` fields

triblondon:
New polyfill is about 3x the size of the current one.  Good to delegate maintenance to someone else but is that necessary?

JakeChampion:
It is necessary as ours is not fully spec compliant. Our polyfill would not let a user pass a different constructor to the function which is defined in [step 4a](http://ecma-international.org/ecma-262/6.0/#sec-array.of).

PR #806, opened 2016-08-05
Add atob btoa via new third party install method

JakeChampion:


.gitignore
```diff
@@ -16,3 +16,5 @@
 /polyfills/Intl/~locale
 
 polyfills/Promise/polyfill.js
+
+polyfills/atob/polyfill.js
```
triblondon: Line spacing here seems a bit strange.

Is it possible to have .gitignore files at sublevels within the repo?  If so we could put it in the polyfill directory.  Alternatively, we could have a different name for polyfill files that are auto-installed, so that they can be ignored with a single pattern.  But that's probably unnecessary.

triblondon:
shirnkwrap always seems to change more than I expect it to...

.gitignore
JakeChampion: I'd be tempted to not even have the polyfills copied to this location and instead have the coed in `lib` deal with the third party install method directly.

rowanmanning:
Looks good 👍

PR #808, opened 2016-08-05
Add AudioContext via new third party install method

JakeChampion:


triblondon:
I've never actually used this and no-one has ever reported any issues with it.  So I don't think it's one that gets used all that much.  The code looks to be fine, assuming the linked polyfill works!

PR #811, opened 2016-08-06
Update dependencies

JakeChampion:
This updates all our dependencies to the latest versions and updates the `npm-shrinkwrap.json` file.

I've tested this by visiting every documentation page, by running the test suite and by running commands which override the default port E.G. `npm start -- --port 4000` (this was to test the update to `minimist`, which I honestly believe no-one is using).


JakeChampion:
``` bash
❯ npm-check --skip-unused
❤️  Your modules look amazing. Keep up the great work. ❤️
```

triblondon:
Nice.  Thanks for updating these.  I only discovered npm-check recently and am using it at Nikkei but hadn't gotten around to thinking about the polyfill service.

Issue #803, opened 2016-08-05
Add boltsort to vcl


wheresrhys:
It should improve the already high cache hit rate even further https://www.fastly.com/blog/varnish-tip-boltsort

JakeChampion:
Great idea! Thanks for bringing this to our attention :-)

JakeChampion:
This work is in master and will be released in next release - `3.10`

PR #812, opened 2016-08-06
Sort querystring using boltsort for better caching

JakeChampion:
_Disclaimer: Everytime I touch VCL I have to read through the documentation on what each function is meant to do._

This pull request takes the advice given in #803 to use `boltsort` for a higher caching hit ratio.

I've used the `boltsort` function inside the `vcl_recv` function to ensure every url in our cache has their querystring normalised.

I believe this is the correct location to add it as every URL except for `^/v2/recordRumData` should hit this code path.

I'm not entirely sure how to test this change except to deploy it to qa.polyfill.io and test it manually.

### DIFF
--- a/fastly-config.vcl
+++ b/fastly-config.vcl
@@ -15,6 +15,7 @@ sub vcl_recv {
 	}
 
 	if (req.url ~ "^/v2/polyfill\." && req.url !~ "[\?\&]ua=") {
+		set req.http.X-Orig-URL = req.url;
 		set req.url = "/v2/normalizeUa?ua=" urlencode(req.http.User-Agent);
 	}
 
--- a/fastly-config.vcl
+++ b/fastly-config.vcl
@@ -1,3 +1,5 @@
+import boltsort;
+
 sub vcl_recv {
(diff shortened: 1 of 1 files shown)


fastly-config.vcl
```diff
@@ -13,14 +15,15 @@ sub vcl_recv {
 	}
 
 	if (req.url ~ "^/v2/polyfill\." && req.url !~ "[\?\&]ua=") {
-		set req.http.X-Orig-URL = req.url;
```
triblondon, line 16: Removing this line will break everything :-)

triblondon:
> Disclaimer: Everytime I touch VCL I have to read through the documentation on what each function is meant to do.

You're not the only one.

Looks like a great idea apart from the line that breaks everything.

fastly-config.vcl
JakeChampion, line 16: How did I remove that?! Thank goodness for your reviewing!

triblondon:
Reassigning to me there was a bit overcautious!  It's fine now :-)

PR #813, opened 2016-08-07
Switch from expect.js to Proclaim

JakeChampion:
Reasons for switching away from expect.js:
- Issues with sparse arrays on ie8 and lower
- Does not look to be merging PRs or being receptive in issues.

Reasons to switch to Proclaim:
- Created and maintained by @rowanmanning, who works for the FT and is a contributor to polyfill-service
- Works down to Android Browser 2.2, Firefox 3.6, Google Chrome 14, Internet Explorer 6, Safari iOS 3 & Safari 5
- Implements all of the assertions in [CommonJS Unit Testing 1.0](http://wiki.commonjs.org/wiki/Unit_Testing/1.0#Assert), meaning if down the road we need to swap assertion library, it will be less painful than swapping from expect.js is.


triblondon:
- Good to get improved compatibility and active development / bug fixing 👍 
- Internally maintained is good and bad - though I appreciate Rowan doesn't make Proclaim on FT time, it would be fair for him to do so if we're using it.  It's obviously cheaper for us to use an expectation library that is a large community project rather than something we have to maintain. However, we also get to prioritise bug fixes if we're sponsoring them!
- I do prefer the syntax style of expect.  It is easier to remember which way around the candidate value and  the expected value go if you write `expect(foo).to.be(42)` than if you write `proclaim.equal(foo, 42)`.  Is there any proclaim support for that / do you violently hate the expect syntax?  I wouldn't veto the change though.

I assume you automated all those commits!?

rowanmanning:
Looks good to me, as long as we're happy maybe contributing some automated browser testing back to Proclaim at some point. In answer to Andrew:
- Yep admittedly Proclaim isn't as widely adopted as expect.js, if I can respond to issues in work time then this is probably OK but it's a valid concern.
- I'm not a fan of the expect syntax, so unlikely to implement in Proclaim. You could reasonably easily write a wrapper library if it's an issue but I prefer the terse assert style and it's kind of standardised in [CommonJS Unit Testing 1.0](http://wiki.commonjs.org/wiki/Unit_Testing/1.0#Assert). The rule is that the order of arguments is always `actual` then `expected`.

JakeChampion:
I've made a PR to Proclaim to add automated browser testing against every version of every browser that SauceLabs has available at the time of testing -- https://github.com/rowanmanning/proclaim/pull/32

PR #815, opened 2016-08-07
Update Element.prototype.after et al. config

simevidas:
Firefox 49 supports the new after/before/append/prepend/replaceWith DOM methods; see the relevant [Mozilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=911477#c40). (Notice how MDN has already been updated.)

I wanted to run your test locally, but I wasn’t able to install this project on my local machine ([screenshot](http://imgur.com/a/oyA5e); on Windows, there is no way to switch to a different version of Node, in case that’s the culprit here).

Instead, I’ve tested [my demo](https://jsbin.com/dewosa/edit?html,js,output) in Firefox 49 on both Windows and Android, and everything checks out.

### DIFF
--- a/polyfills/Element/prototype/after/config.json
+++ b/polyfills/Element/prototype/after/config.json
@@ -9,7 +9,7 @@
 		"android": "*",
 		"bb": "*",
 		"chrome": "*",
-		"firefox": "3.6 - *",
+		"firefox": "<49",
 		"ios_chr": "*",
 		"ios_saf": "*",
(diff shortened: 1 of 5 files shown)


JakeChampion:
Thanks, yes this project requires node 6 to run. I will get this tested on CI for you :-) Thanks again for taking the time to create a pull request and testing it.

JakeChampion:
@simevidas Your commit now passes CI 😄  🎉

PR #817, opened 2016-08-08
Remove a not own property constraint for iterators in Array.from

### LINKED ISSUE #816: Array.from function doesn't work with iterators such as Map/Set methods entries/keys/values
`Array.from`, `Map`, and `Set` are polyfills.

``` js
Array.from(new Map().set(0, 0).entries());
Array.from(new Set().add(0).entries());
```

expected:

```
[[0, 0]]
[0]
```

actual:

```
[]
[]
```

falsandtru:
Fixes #816.

### DIFF
--- a/polyfills/Array/from/polyfill.js
+++ b/polyfills/Array/from/polyfill.js
@@ -9,7 +9,7 @@
 		// if the iterable doesn't have next;
 		// it is an iterable if 'next' is a function but it has not been defined on
 		// the object itself.
-		if (typeof arraylike.next === 'function' && arraylike.hasOwnProperty('next') === false) {
+		if (typeof arraylike.next === 'function') {
 			while (!done) {
 				iterableResponse = arraylike.next();
 				if (
--- a/polyfills/Array/from/tests.js
+++ b/polyfills/Array/from/tests.js
@@ -87,6 +87,27 @@ describe('returns an array with', function () {
 			}
 		}
 
+		it('can convert from a user-defined iterator', function () {
+			function iterator(cnt) {
+				return {
+					next: function () {
+						return cnt === 0
+							? {
+								done: true
+							}
+							: {
+								value: cnt--,
+								done: false
+							};
+					}
+				};
+			}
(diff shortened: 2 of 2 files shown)


falsandtru:
@triblondon @JakeChampion Can you take a look? Note, I don't have api keys of saucelabs. How can I resolve the following error?

```
Warning: SAUCE_USER_NAME and SAUCE_API_KEY must be set in the environment Use --force to continue.
```

JakeChampion:
I've resolved this issue by bring your forked repository's branch into our origin repository. CI now passes :-)

falsandtru:
Thanks! Could you merge this?

JakeChampion:
Would you be able to add a test to cover this change?

falsandtru:
This change is already covered in the following test.

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Array/from/tests.js#L46

But old code passed this test even though it should be failed. Then I don't know how can I test this change using polyfills. Have you any ideas?

JakeChampion:
Ah yes, that is because we only install the polyfills which the requested polyfill needs to function. `Array.from` does not need `Set` or `Map` polyfills to function so they are not downloaded. This in turn means those tests you linked to would not run as `Set` and `Map` would not be present.

If we create a custom iterable object and use that in the tests then we can test this code path in older browsers which don't have `Set` or `Map`.

falsandtru:
Okay, I'll try it.

JakeChampion:
Great work @falsandtru !

falsandtru:
Thanks @JakeChampion !

JakeChampion:
@falsandtru I forgot to ask if you are okay with our contribution terms, apologies for that. Our terms can be [found here](https://cdn.polyfill.io/v2/docs/contributing#contribution-terms).

falsandtru:
I agree, this is a wholly my own work, and you can use this under your license.

falsandtru:
@JakeChampion Could you delete the `array.from` branch for the next pr?

falsandtru:
Sorry, I found the workaround.

PR #818, opened 2016-08-08
Fix browser support configurations of Map/Set/WeakMap/WeakSet

### LINKED ISSUE #793: Some polyfills are not provided for Android browser
Android browsers that have a following user-agent need Map/Set/WeakMap/WeakSet polyfills, but this polyfill service provides only Map/Set polyfills. So I got an error `ReferenceError: WeakSet is not defined`.

```
"Mozilla/5.0 (Linux; Android 4.4.2; KYY24 Build/108.1.2110) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/30.0.0.0 Mobile Safari/537.36"
```

falsandtru:
Fixes #793.

### DIFF
--- a/polyfills/Map/config.json
+++ b/polyfills/Map/config.json
@@ -5,11 +5,14 @@
 	],
 	"browsers": {
 		"ie": "<12",
+		"safari": "<7.1",
 		"firefox": "<29",
-		"safari": "<8",
-		"ios_saf": "<7.1",
 		"chrome": "<38",
-		"firefox_mob": "<29"
+		"android": "<5.1",
+		"firefox_mob": "<29",
+		"ie_mob": "*",
+		"opera_mob": "*",
+		"ios_saf": "<8"
 	},
 	"dependencies": [
 		"Array.prototype.indexOf",
--- a/polyfills/WeakMap/config.json
+++ b/polyfills/WeakMap/config.json
@@ -5,10 +5,11 @@
 	"browsers": {
 		"ie": "7 - 11",
 		"ie_mob": "10 - 11",
-		"safari": "*",
+		"safari": "<7.1",
 		"chrome": "<38",
 		"firefox": "<36",
-		"ios_saf": "*",
+		"android": "<5",
+		"ios_saf": "<8",
 		"opera": "<25",
 		"firefox_mob": "<36"
 	},
--- a/polyfills/WeakMap/config.json
+++ b/polyfills/WeakMap/config.json
@@ -5,11 +5,11 @@
 	"browsers": {
 		"ie": "7 - 11",
 		"ie_mob": "10 - 11",
-		"safari": "<7.1",
+		"safari": "<9",
 		"chrome": "<38",
 		"firefox": "<36",
-		"android": "<5",
-		"ios_saf": "<8",
+		"android": "*",
+		"ios_saf": "<9",
 		"opera": "<25",
 		"firefox_mob": "<36"
 	},
--- a/polyfills/WeakSet/config.json
+++ b/polyfills/WeakSet/config.json
@@ -3,12 +3,13 @@
 		"es6"
 	],
 	"browsers": {
-		"ie": "7 - *",
-		"ie_mob": "10 - *",
-		"safari": "*",
-		"chrome": "<38",
+		"ie": "7 - 11",
+		"ie_mob": "10 - 11",
+		"safari": "<9",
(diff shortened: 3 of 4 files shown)


polyfills/Map/config.json
```diff
@@ -5,11 +5,14 @@
...
 		"ie": "<12",
+		"safari": "<7.1",
 		"firefox": "<29",
-		"safari": "<8",
-		"ios_saf": "<7.1",
 		"chrome": "<38",
-		"firefox_mob": "<29"
+		"android": "*",
+		"firefox_mob": "<29",
+		"ie_mob": "*",
```
JakeChampion, line 13: Has this been tested?

polyfills/Map/config.json
```diff
@@ -5,11 +5,14 @@
...
+		"opera_mob": "*",
+		"ios_saf": "<7.1"
```
JakeChampion: Same question for `opera_mob`.

falsandtru: These changes come from Set/config.json. This config possibly will be working because Set/config.json is working, and already has `"android": "*"`. WeakSet's changes also come from WeakMap.

triblondon:
Super useful if you could drop in a screengrab of the compat table you're using to derive these numbers.  Cautious here as we don't test Opera, ie mobile or FF mobile in CI.

falsandtru:
Okay, is there an example screenshot?

> Opera, ie mobile or FF mobile

I also don't know those browsers. But old configs probably doesn't work with those browsers.

triblondon:
Example for `Map` from https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map:

image

We'd like to make sure that proposed changes to targeting configs are not guesses, but are based on tests. A screenshot of our tests running in that browser before and after the polyfill is ideal. but failing that, citing an authoritative source of compat data is good enough.

falsandtru:
Thanks.

image

## Map

image

## Set

image

## WeakMap

image

## WeakSet

image

falsandtru:
I'll be absent for a few days because I'm going to the center of universe 🚀

falsandtru:
I agree with contribution terms of this project.

polyfills/WeakSet/config.json
```diff
@@ -3,12 +3,13 @@
...
 	],
 	"browsers": {
-		"ie": "7 - *",
-		"ie_mob": "10 - *",
-		"safari": "*",
-		"chrome": "<38",
+		"ie": "7 - 11",
+		"ie_mob": "10 - 11",
+		"safari": "<9",
+		"chrome": "<36",
```
JakeChampion: chrome 36 doesn't support `new WeakSet(iterable)` - could we revert this back to chrome 38?

polyfills/WeakSet/config.json
```diff
@@ -3,12 +3,13 @@
...
 		"firefox": "<34",
-		"ios_saf": "*",
+		"android": "<5",
```
JakeChampion: android 5.1 does not have full WeakSet support, it is missing 3 features:
- iterator closing
- no `WeakSet.prototype.clear` method
- `.has` and `.delete` methods accept primitives

polyfills/WeakMap/config.json
```diff
@@ -5,10 +5,11 @@
 	"browsers": {
 		"ie": "7 - 11",
 		"ie_mob": "10 - 11",
-		"safari": "*",
+		"safari": "<7.1",
```
JakeChampion: Full `WeakMap` support was added in safari 9.

polyfills/WeakMap/config.json
```diff
@@ -5,10 +5,11 @@
...
 		"chrome": "<38",
 		"firefox": "<36",
-		"ios_saf": "*",
+		"android": "<5",
+		"ios_saf": "<8",
```
JakeChampion: Full `WeakMap` support was added in ios safari 9

falsandtru:
I see, wait a moment please.

falsandtru:
Fixed. If it is okay, I'll squash my commits.

JakeChampion:
Of course :-) Thanks for responding so swiftly!

JakeChampion:
Thanks for improving our browser support coverage for these polyfills. This will be released as part of version `3.11.0`.

PR #821, opened 2016-08-09
Sourcemaps issue

### LINKED ISSUE #820: SyntaxError: Unexpected end of script with minified version on safari
Below is the url we are hitting.  Something must have changed recently as this used to work on Safari.

https://cdn.polyfill.io/v2/polyfill.min.js?features=default,Intl.~locale.fr,Intl.~locale.en,Intl.~locale.es

In the minified version you can find this offending code.  It is commenting out the rest of the file and causing the issue.  If we remove the "min" from the above URL it solves the issue.

//# sourceMappingURL=Intl.min.js.mapIntlPolyfill.__addLocaleData({locale:"en",date:{ca:

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-service/issues/820

Before this PR:
When requesting minified polyfill bundle we would return all required polyfills concatenated together. The issue with this approach is some third-party polyfills include a sourcemap comment at the end of their code, this causes our bundle to end execution early.

After this PR:
When requesting minified polyfill bundle we would return all required polyfills concatenated together with a new line between each polyfill. This will stop the issue of a comment at the end of a polyfill file from commenting out the rest of the bundle.


JakeChampion:
This is the PR which introduced the sourcemap in Intl. @caridy, do you think this is better solved here (what this PR proposes) or by Intl?

caridy:
yeah, that should work just fine, but ideally we can just get the sourcemap annotation removed from the polyfills when they are added to the service, there are a couple of npm pkgs that can do that as part of the build process.

caridy:
https://github.com/thlorenz/convert-source-map#removecommentssrc

JakeChampion:
Great idea @caridy, thanks for suggesting the package, it works as expected. I've updated the PR with your recommendation :-)

triblondon:
@JakeChampion I couldn't make this code work for me, either in QA or locally, or in a simple reduced test case.  Weird.  I ended up doing this:

https://github.com/Financial-Times/polyfill-service/compare/v3.9.0...v3.9.2

If you're happy with that, I'll make the same change to master.

barrett-vegas-com:
This problem is happening for us still when hitting
https://cdn.polyfill.io/v2/polyfill.min.js?features=Intl.~locale.en&unknown=polyfill

With UA: Mozilla/5.0 (Linux; Android 6.0.1; SAMSUNG-SM-G920A Build/MMB29K) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/4.0 Chrome/44.0.2403.133 Mobile Safari/537.36

Service version states 3.11.0 when hitting non-minified version.

PR #829, opened 2016-08-11
Handle response from file:// urls correctly in fetch polyfill

ahocevar:
Since file:// urls do not respond with a status code, we can assume an equivalent of 200 when we have reached the onload handler.

This was encountered in openlayers/ol3#5712, using PhantomJS to request file:// urls.

See openlayers/ol3#5652 for a similar situation that was fixed.


ahocevar:
Closing, because rejecting file:// urls is the native behavior of current implementations as well. See github/fetch#186.

PR #831, opened 2016-08-12
install json via third party install method

JakeChampion:


triblondon:
You changed the choice of polyfill here, are we still using the right browser targeting?

JakeChampion:
It actually targets even more browsers :-)

triblondon (approved):
OK with me.

PR #836, opened 2016-08-16
Add a script to purge assets from the CDN, run this script after each deploy

JakeChampion:
Currently whenever a deployment takes place we need to manually purge the assets from the CDN in order for users to see the new documentation. This PR adds a script which we can run to purge the assets from the CDN based upon the environment we are targeting, defaulting with QA.

The PR also adds this script to the end of the deployment scripts, ensuring that after each deployment, the cache is purged.


triblondon:
This is fine.  In my rum branch i'll move this into tasks/node

rowanmanning:
Looks good to me 👍

PR #838, opened 2016-08-18
Add StyleSheetList[Symbol.iterator] for browsers which have StyleSheetList

JakeChampion:
There is no specification for this feature that I can find, but it does seem to exists in Chrome and Firefox. It does not exist in Safari.

This polyfill currently doesn't target IE8 as IE8 doesn't have `document.styleSheets` and we do not have a polyfill for this as of yet.

### DIFF
--- a/polyfills/Object/defineProperties/polyfill.js
+++ b/polyfills/Object/defineProperties/polyfill.js
@@ -1,8 +1,6 @@
 Object.defineProperties = function defineProperties(object, descriptors) {
 	for (var property in descriptors) {
-		if (Object.hasOwnProperty(property)) {
-			Object.defineProperty(object, property, descriptors[property]);
-		}
+		Object.defineProperty(object, property, descriptors[property]);
 	}
 
 	return object;
--- a/polyfills/Object/defineProperties/polyfill.js
+++ b/polyfills/Object/defineProperties/polyfill.js
@@ -1,6 +1,8 @@
 Object.defineProperties = function defineProperties(object, descriptors) {
 	for (var property in descriptors) {
-		Object.defineProperty(object, property, descriptors[property]);
+		if (Object.hasOwnProperty(property)) {
+			Object.defineProperty(object, property, descriptors[property]);
+		}
 	}
 
 	return object;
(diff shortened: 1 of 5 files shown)


polyfills/Object/defineProperties/polyfill.js
```diff
@@ -1,6 +1,8 @@
 Object.defineProperties = function defineProperties(object, descriptors) {
 	for (var property in descriptors) {
-		Object.defineProperty(object, property, descriptors[property]);
+		if (Object.hasOwnProperty(property)) {
```
triblondon: Are you sure about this?  Seems like if anything it should be checking the inverse of this condition.

polyfills/StyleSheetList/@@iterator/config.json
```diff
@@ -0,0 +1,20 @@
+{
+	"browsers": {
+		"ie": "8 <",
```
triblondon: Uncommon semver syntax - we would commonly write `8 - *`

polyfills/StyleSheetList/@@iterator/config.json
```diff
@@ -0,0 +1,20 @@
...
+		"firefox": "*",
+		"chrome": "*",
+		"safari": "*",
+		"android": "*",
+		"ios_saf": "*",
+		"opera": "*"
+	},
+	"dependencies": [
+		"_ArrayIterator",
+		"Symbol.iterator"
```
triblondon: Does the polyfill work without a `StyleSheetList` polyfill?

polyfills/StyleSheetList/@@iterator/detect.js
```diff
@@ -0,0 +1 @@
+'Symbol' in this && 'iterator' in this.Symbol && !!StyleSheetList.prototype[Symbol.iterator]
```
triblondon: I would suggest adding a test for `StyleSheetList` as well.

polyfills/StyleSheetList/@@iterator/config.json
JakeChampion: No it doesn't, IE8 needs a `StyleSheetList` polyfill. I decided to not target IE8 for now.

PR #840, opened 2016-08-18
Fix Array.from static method

falsandtru:
Support Symbol.iterator and surrogate pairs.


polyfills/Array/from/tests.js
```diff
@@ -108,11 +108,37 @@ describe('returns an array with', function () {
 			proclaim.deepEqual(Array.from(iterator(3)), [3, 2, 1]);
 		});
 
+		it('can convert from a user-defined iterator using the Symbol.iterator', function () {
```
JakeChampion: This will needs a skips option for browsers which don't support `Symbol.iterator`.

JakeChampion:
@falsandtru I've been working on full iterator support for `Array.from` on the upstream polyfill. You may be interested in checking that out :-) https://github.com/mathiasbynens/Array.from/pull/27

polyfills/Array/from/tests.js
falsandtru: Is there a case that this service doesn't provide Symbol polyfills even when browsers don't support the Symbol?

falsandtru:
When can I use its upstream polyfills?

JakeChampion:
Not for a while, we need to add support for polyfills written in using module systems such as commonjs before we will be able to use the upstream polyfill directly. Perhaps we can copy some of the changes in the upstream pull request into our local polyfill? Namely the Safari browser support for `Set` and `Map` iteration.

falsandtru:
I see, where can I fix the polyfill? Here, or upstream?

JakeChampion:
Fixing here for now, I can take on the work to push upstream :-)

JakeChampion:
I'm running the tests on CI for this PR, thanks for the contribution!

falsandtru:
Sorry, I misunderstand about iterators. This code is wrong, and is a partial support. How should I do?

polyfills/Array/from/polyfill.js
```diff
@@ -1,45 +1,34 @@
 
 // Wrapped in IIFE to prevent leaking to global scope.
 (function () {
+	function isIterable (arraylike) {
+		switch ('function') {
+			case typeof arraylike['next']:
+			case typeof arraylike[Symbol.iterator]:
```
JakeChampion: We can't assume `Symbol` will exist, Firefox has versions which use `@@iterator` instead.

JakeChampion:
I will be attempting to complete my pull-request made upstream by the end of this week. I have two tasks left to do on it:
- Get the polyfill working across different realms
- Fix the failing tests that are present on `master`

If you like, you can help by cloning [`Array.from`](https://github.com/mathiasbynens/Array.from/) and running the tests against a browser such safari ios 7, the test `no setters are called for indexes` fails.

falsandtru:
I see, then I close this pr, and wait the upstream fix. I'm very glad. Thanks a lot!

JakeChampion:
You are more than welcome. Thank you for taking the time to create a pull request, you've been quite active on this project recently, I'm very happy that you find it useful. How have you found contributing to this project? Was there anything in the documentation or contribution guide that you found was confusing or could be worded in a better way? We're are always trying to make contributing easier for members of the community and love hearing feedback and how we can improve :-)

falsandtru:
I'll create some pr for Android support after several days. Many configs are not defined android versions. Thanks again 👍

polyfills/Array/from/tests.js
```diff
@@ -87,32 +87,46 @@ describe('returns an array with', function () {
...
-								done: false
-							};
+		it('can convert from a custom iterator', function () {
+			function iterable(count) {
+				var iterator = {
+					'@@iterator': function () {
+						throw new Error();
 					}
 				};
+				iterator[Symbol.iterator] = function () {
```
JakeChampion, line 99: I think you need to add `Symbol.iterator` as a dependency in the `config.json` for this polyfill otherwise this will throw in any browser which don't support `Symbol.iterator`.

falsandtru: Okay, I added a dependency.

JakeChampion: Thanks, rerunning the tests for this.

polyfills/Array/from/polyfill.js
```diff
@@ -1,14 +1,15 @@
 
 // Wrapped in IIFE to prevent leaking to global scope.
 (function () {
-	function parseIterable(arraylike) {
-		if (typeof arraylike[Symbol.iterator] !== 'function') {
+	var iterator = typeof Symbol === 'function' ? Symbol.iterator : '@@iterator';
```
JakeChampion: Thanks for reverting this. Polyfilling iterators is a rather tough challenge to get 100% correct.

JakeChampion:
@falsandtru Re-running the tests for you now :-)

falsandtru:
Hmm, it seems to have caused by conflict between Symbol.iterator and `@@iterator`. I'll try to fix it.

falsandtru:
Probably there are two ways.
1. Check `@@iterator` in all polyfills, per polyfill.
2. Change the Symbol.iterator polyfill to use `@@iterator` in old firefox.

I recommend the second way that change Symbol.iterator. The first way will make bugs.

falsandtru:
I'm trying the first way only for Array.from because the second way is difficult to pass the tests of Symbol.

falsandtru:
Fixed. Please run ci with latest 2 commits.

JakeChampion:
Re-running :-)

falsandtru:
Thanks :) If the test for `73227c3` is passed, I recommend that we drop the last commit `767c1e2` for simplicity and consistency with other polyfills.

polyfills/Array/from/tests.js
```diff
@@ -87,35 +87,37 @@ describe('returns an array with', function () {
...
-						next: function () {
-							return cnt === 0
-								? {
-									done: true
-								}
-								: {
-									value: cnt--,
-									done: false
-								};
+		if ([][Symbol.iterator]) {
```
JakeChampion: Can we place this `if` statement inside the `it` function and add an `else` statement where we call `this.skip();`. This would let mocha know that a test was skipped and helps us when it comes to viewing the test reports.

falsandtru: Is this ok?

```
it('can convert from a custom iterator', function () {
    if (![][Symbol.iterator]) this.skip();
```

JakeChampion: Nearly, I would also check that `Symbol` itself exists.

falsandtru: I see, fixed.

falsandtru:
@JakeChampion Unfortunately, we cannot use the upstream polyfill because it cannot use with Symbol.iterator polyfill https://github.com/mathiasbynens/Array.from/issues/36. So can we use this fix?

JakeChampion:
@falsandtru I think it would be better to copy the upstream polyfill into the polyfill-service and modify it to work with Symbol polyfills.

falsandtru:
However, then we need large changes to work with Symbol polyfills.

falsandtru:
Seems like we cannot remove fixes for legacy Firefox support. So I added it https://github.com/Financial-Times/polyfill-service/pull/840/commits/d4d0f9a81032e82462bed2a33f770c97fd9cea2c.

polyfills/Array/from/polyfill.js
```diff
@@ -1,48 +1,33 @@
...
-					tempArray.push(iterableResponse.value);
-				} else {
-
-					// it does not conform to the iterable pattern
-					break;
-				}
-			}
+	var mozIterator = '@@iterator';
+	var hasSymbol = typeof Symbol === 'function';
+	var hasMozIterator = typeof [][mozIterator] === 'function';
```
JakeChampion, line 6: This makes it work in Firefox but it will still fail in Safari versions which only support iteration via `for...of` - https://github.com/mathiasbynens/Array.from/blob/master/implementation.js#L168

falsandtru: I see, I want to copy your impl. But I have a question. The following your code will possibly always throw an error because a generated function cannot change `y`, `y` would always be `undefined`. Why it works correctly? Safari's bug?

``` js
        var map = new Map();
        map.set(1, 2);
        var y;
        Function('map', 'y', 'for (var x of map) {y = x}')(map, y); // eslint-disable-line no-new-func
        if (y[0] === 1 && y[1] === 2) {
            forOfOnly = true;
        }
```

JakeChampion: This should do it:

``` js
var map = new Map();
map.set(1, 2);
var y = Function('map', 'for (var x of map) {y = x} return y')(map); // eslint-disable-line no-new-func
if (y[0] === 1 && y[1] === 2) {
    forOfOnly = true;
}
```

JakeChampion: Good catch, it's a bug in my code where I confused the execution rules of `eval` with `Function`, this should be pushed back upstream.

falsandtru: Okay, do you fix it yourself in upstream? Otherwise, I can fix it today.

JakeChampion: If you have time to fix today that would be great, I'm busy for the rest of today on other work.

falsandtru: Okay, I made https://github.com/mathiasbynens/Array.from/pull/37.

PR #841, opened 2016-08-18
Add ability to use live feature detection instead of being based on User-Agent

JakeChampion:
This is a proposed solution to the live feature detection requested in issue #84.

This PR adds a new endpoint (`/v2/detect.js`) for serving the feature detects only.

The endpoint `/v2/detect.js` accepts two query parameters:
- `features` : List of browser features to feature-detect. Accepts a comma-separated list of feature names. Available feature names are shown on the [Browsers and Features page](https://cdn.polyfill.io/v2/docs/features/), as well as group aliases such as 'es5' and 'es6'.
- `callback` : Name of JavaScript function to call after detects are loaded. The function will be invoked with an array containing the names of each feature which was not detected from the `features` list provided. If all features are supported by the browser, an empty array will be passed into the callback function.

One way to use this endpoint is to make a custom request to our polyfill endpoint with the missing features for the browser. Here is one example of this:

``` js
<script>
var callback = function(features) {
    var url = 'https://cdn.polyfill.io/v1/polyfill.min.js?ua=false&unknown=polyfill&flags=always&features=' + features.join(',');
    var script = document.createElement('script');
    script.src = url;
    document.body.appendChild(script);
};
</script>
<script src="http://localhost:3000/v2/detect.js?features=Event.focusin&callback=callback"></script>
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -231,6 +231,8 @@ function getDetectString(options) {
 	let callback;
 	if (options && typeof options.callback === 'string') {
 		callback = options.callback;
+	} else {
+		throw new TypeError('`getDetectString` requires a callback function to be passed. `getDetectString({callback: "callback"})`.');
 	}
 	const unknown = [];
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,6 +8,8 @@ const UA = require('./UA');
 const sourceslib = require('./sources').getCollection();
 const appVersion = require(path.join(__dirname,'../package.json')).version;
 const Handlebars = require('handlebars');
+const {stripIndents} = require('common-tags');
+const co = require('co');
 
 // Load additional useragent features: primarily to use: agent.satisfies to
 // test a browser version against a semver string
@@ -34,6 +36,15 @@ function getOptions(opts) {
 	}, opts);
 }
 
+const resolveAliases = createAliasResolver([
+	function aliasFromConfig(featureName) {
+		return sourceslib.getConfigAliases(featureName);
+	},
+	function aliasAll(featureName) {
+		return (featureName === 'all') ? sourceslib.listPolyfills() : undefined;
+	}
+]);
+
 /**
  * Given a set of features that should be polyfilled in 'options.features' (with flags i.e. `{<featurename>: {flags:[<flaglist>]}, ...}`), determine which have a configuration valid for the given options.uaString, and return a promise of set of canonical (unaliased) features (with flags) and polyfills.
  *
@@ -44,15 +55,6 @@ function getPolyfills(options) {
 
 	options = getOptions(options);
 	const ua = new UA(options.uaString);
-
-	const resolveAliases = createAliasResolver([
-		function aliasFromConfig(featureName) {
-			return sourceslib.getConfigAliases(featureName);
-		},
-		function aliasAll(featureName) {
-			return (featureName === 'all') ? sourceslib.listPolyfills() : undefined;
-		}
-	]);
 
 	const resolveDependencies = createAliasResolver(
 		function aliasDependencies(featureName) {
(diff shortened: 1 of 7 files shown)


JakeChampion:
@simevidas, I know it was a long time ago when you requested this feature, apologies if you no longer need it. If you are still interested in this feature, would you be able to give this branch a spin and see if it is what you are wanting?

triblondon:
I'll try this out next week, when I get my cast shortened so I can straighten my arm :-(

lib/index.js
```diff
@@ -224,10 +226,86 @@ function getPolyfillString(options) {
 	;
 }
 
+function getDetectString(options) {
```
triblondon, line 229: Isn't this kind of just reimplementing `getPolyfillString`?  Could the existing `getPolyfillString` be abstracted into something like `buildOutputString` and then be called in two different modes?  You could also avoid adding the `co` dependency`.  I'm not against using third party code but if we already have a way of doing something in one part of the code it seems like a good idea to be consistent with that.

lib/index.js
```diff
@@ -224,10 +226,86 @@ function getPolyfillString(options) {
...
+		})
+		.then(detects => {
+
+			// Outer closure hides private features from global scope
+			// Invoke the closure, binding `this` to window (in a browser), self (in a web worker), or global (in Node/IOjs)
+			let featureDetectBundle = stripIndents`(function(undefined) {
+				var featuresToPolyfill = [];
+				${detects.join('')}
+				${callback ? `${callback}(featuresToPolyfill);` : ''}
+				return featuresToPolyfill;
```
triblondon, line 287: Where would this return statement be returning to?

lib/index.js
```diff
@@ -224,10 +226,86 @@ function getPolyfillString(options) {
...
+			}).call('object' === typeof window && window || 'object' === typeof self && self || 'object' === typeof global && global || {});`;
```
triblondon, line 288: A template string inside another template string strikes me as confusing code.  It might seem weird to use a library over the built in feature, but we already have handlebars as a dependency of the project,and id be tempted to use it here like i've done in the RUM template.  This then also avoids strings of client side code being interspersed with servr side code in the same file, and means you can avoid adding a dependency (`common-tags`) to the project.

polyfills/~html5-elements/config.json
```diff
@@ -16,5 +16,8 @@
 	"install": {
 		"module": "html5shiv",
 		"paths": ["dist/html5shiv.min.js"]
-	}
+	},
+	"dependencies": [
+		"Document"
+	]
```
triblondon, line 22: This seems a bit out of place in this PR

service/routes/api.js
```diff
@@ -84,6 +84,22 @@ router.get(/^\/v2\/polyfill(\.\w+)(\.\w+)?/, (req, res) => {
 
 });
 
+router.get(/^\/v2\/detect(\.\w+)(\.\w+)?/, (req, res) => {
+	const polyfills = PolyfillSet.fromQueryParam(req.query.features);
+
+	const params = {
+		features: polyfills.get(),
+		callback: req.query.callback
```
triblondon, line 92: Callback is needed for this feature to work, right?  May be worth considering what happens if its missing.

triblondon:
I think this is a good idea, though i think it could be lighter, both in terms of our code and number of new dependencies.  Also maybe you could consider offering an option to built the boilerplate that glues together detect and polyfill into detect?

polyfills/~html5-elements/config.json
JakeChampion, line 22: When running the bundled detect code in NodeJS, it will fail because `Document` does not exist. This change resolves that issue :-)

PR #848, opened 2016-08-23
widen browser delivery of iterators

wheresrhys:
@JakeChampion 

The ie_mob rule is definitely needed (and also makes me think we should perhaps fallback to ie on ie_mob if ie_mob isn't defined)

In the absence of data I'm going for '*' for samsung browsers. We can make it more stringent using RUM in the near future I suppose, but until then not covering samsung is a blocker to next using polyfill.io for symbol

### DIFF
--- a/polyfills/Array/prototype/@@iterator/config.json
+++ b/polyfills/Array/prototype/@@iterator/config.json
@@ -5,12 +5,14 @@
 	],
 	"browsers": {
 		"ie": "9 - 12",
+		"ie_mob": "9 - 12",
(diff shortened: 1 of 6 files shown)


JakeChampion:
@AdaRoseEdwards Are you able to test this on Samsung Browser?

AdaRoseCannon:
Sure I'll do it tomorrow.

JakeChampion:
@AdaRoseEdwards Have you had time to check this out for Samsung?

PR #851, opened 2016-08-29
Lambda task for RUM processing

triblondon:
OK, squashed out all the authentication tokens and reduced my work to one commit.

Issues we may need to resolve:
- compatibility testing largely excludes the possibility of finding false negatives, because a request for 10 features, which delivers polyfills for five of them, will only include RUM tests for the five features we have polyfilled. We should probably include RUM tests for all requested features regardless of whether they got polyfilled or not.

Sample output from `/v2/getRumPerfData`:

``` csv
"data_center","country","count","perf_dns_count_notnull","perf_dns_min","perf_dns_max","perf_dns_mean","perf_dns_std","perf_dns_95","perf_dns_99","perf_connect_count_notnull","perf_connect_min","perf_connect_max","perf_connect_mean","perf_connect_std","perf_connect_95","perf_connect_99","perf_req_count_notnull","perf_req_min","perf_req_max","perf_req_mean","perf_req_std","perf_req_95","perf_req_99","perf_resp_count_notnull","perf_resp_min","perf_resp_max","perf_resp_mean","perf_resp_std","perf_resp_95","perf_resp_99"
"IAD","US","149","13","0","91","21.17","22.52","20","33","14","0","238","65.27","60.21","38","142","130","0","505","28.64","65.96","96","383","130","0","46","5.36","7.08","21","38"
"JFK","US","89","13","0","418","78.26","107.66","57","418","12","0","311","124.04","87.8","147","311","67","0","62","12.81","13.27","48","62","66","0","39","4.25","6.9","15","39"
"LCY","GB","66","1","0","0","0","0","0","0","2","0","10","8.14","1.88","0","10","64","1","275","24.39","51.87","112","275","64","0","23","2.82","4.02","13","23"
```

This includes performance data for each of the DNS look up, TCP connect, request and response, in each case providing the following statistics:
- Count of non-null values recorded (null values can result from beacon data from browsers that do not support the resource timing API)
- minimum value
- maximum value
- mean value
- Standard deviation
- 95th percentile value (probably the most useful)
- 99th percentile value

Sample output from `/v2/getRumCompatData`:

``` csv
"feature_name","ua_family","ua_version","pass_rate","ip_count","total_count","is_targeted","targeting_status"
"Element.prototype.before","chrome","52.0.0","1","103","338","Yes","False positive"
"Element.prototype.prepend","chrome","52.0.0","1","103","338","Yes","False positive"
"Element.prototype.replaceWith","chrome","52.0.0","1","103","338","Yes","False positive"
```

These reports are still not very user-friendly. Arguably the compatibility data could be used as-is (although the current logic for measuring compatibility will exclude false negatives because we will not include tests for features that we are not pollyfilling).  
However the performance data is difficult to read, so I'm thinking of adding a new section to the usage data page, potentially like this:

image

(this network timeline mock up uses the same colour scheme as chrome network timeline)


triblondon:
Done the perf visualisation:

image

service/routes/docs.js
```diff
@@ -14,8 +14,19 @@ const compatdata = require('../../docs/assets/compat.json');
...
-const cachettls = {fastly:1800, respTimes:1800, outages:86400};
+const cachettls = {fastly:1800, respTimes:1800, outages:86400, rumPerf:86400};
 const templCache = {};
+let mysql;
+
+if (process.env.RUM_MYSQL_DSN) {
+	require('promise-mysql')
+		.createConnection(process.env.RUM_MYSQL_DSN)
+		.then(conn => {
+			mysql = conn;
```
JakeChampion, line 25: If the connection fails to be made, `mysql` will be undefined.

service/routes/docs.js
```diff
@@ -164,6 +176,31 @@ function getData(type) {
 				}));
 			}));
 		},
+		rumPerf: () => {
+			if (!mysql) return [];
```
JakeChampion, line 180: Cool, we deal with an undefined `mysql` here. :+1:

tasks/lambda/project.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "origami-polyfill",
+  "description": "AWS Lambda functions for Polyfill service tasks",
+  "role": "arn:aws:iam::371548805176:role/origami-polyfill_lambda",
```
JakeChampion: Is it safe to commit this to a public repo?

Gruntfile.js
```diff
@@ -88,18 +88,34 @@ module.exports = function(grunt) {
 			dryrun: { options: {service: "qa", dryRun: true} },
 			qa: { options: {service: "qa"} },
 			prod: { options: {service: "prod"} }
+		},
+		"shell": {
+			"deployrumlambda": {
+				command: env => {
+
+					// These AWS profiles are expected to be in ~/.aws/credentials
```
JakeChampion: Do you have an example of what these should be? Should we store them in LastPass?

README.md
```diff
@@ -85,8 +75,40 @@ grunt ci             # run the service tests and polyfills tests on a large set
...
 
-Release Process
----------------
+
+## Real User Monitoring
+
+We have shipped experimental support for using RUM to monitor feature support and performance in browsers.  This involves a number of parts, all activated by the presence `RUM_MYSQL_DSN` env var:
+
+* **RUM client code**: a [small snippet](lib/rumTemplate.js.handlebars) of legacy-compatible code that will evaluate feature detects on the client, sample resource timing data, and beacon the results back to the service. This is shipped as part of the Node app.
+* **Beacon endpoint**: an [endpoint to collect RUM data](fastly-config.vcl), terminated at the CDN, logging query data out to Amazon S3 to avoid overloading the backend.  To clarify, the backend node server does not provide a route handler for the RUM data collection URL, so this is shipped when we deploy VCL to Fastly.  It also requires log streaming to be configured in the Fastly UI.
```
JakeChampion, line 84: Could we add `#L22` to the link, that way we direct the reader to the part of the file which is doing this work.

README.md
```diff
@@ -85,8 +75,40 @@ grunt ci             # run the service tests and polyfills tests on a large set
...
+* **Lambda processing function**: an [AWS Lambda function](tasks/lambda/functions/rum-process/index.js) is used to move the data from the raw log files on S3 into the MySQL backend.  This is shipped using a dedicated process with [Apex](http://apex.run), see below.
+* **Reporting endpoints**: [API routes that deliver useful analysis of the RUM data](service/routes/rum.js) are provided in the node server.  These return CSV data intended to populate a spreadsheet.  This is shipped as part of the Node app.
+
+Because this requires a fair amount of orchestration, we recommend only enabling it for the FT hosted version.  If you want to run the service yourself, you can opt out of this RUM feature by not setting a `RUM_MYSQL_DSN`.
```
JakeChampion, line 88: Should we look into automating this orchestration using something like terraform?

README.md
```diff
@@ -85,8 +75,40 @@ grunt ci             # run the service tests and polyfills tests on a large set
...
+
+### Routes
+
+* `/v2/getRumPerfData`: Return stats for resource timing metrics, grouped by CDN POP and country of request origin
+* `/v2/getRumCompatData`: Return stats for differences between current targeting configuration and live feature detect results from browers, grouped by feature name, browser family and version.
+
+### Deploying Lambda
+
+All the bits of the RUM solution are deployed as part of our existing deployment workflow except the Lambda functions, which require [Apex](http://apex.run).  To deploy the Lambda functions:
```
JakeChampion: Does this mean we require a global installation of Apex? Is it possible to have it as a development dependency and have the grunt task use it directly?

README.md
```diff
@@ -85,8 +75,40 @@ grunt ci             # run the service tests and polyfills tests on a large set
...
+
+1. Set up AWS profiles in your `~/.aws/credentials` file for `[polyfill-qa]` and `[polyfill]`, and configure the default region to match the location you want to run the Lambdas. We typically use `eu-west-1`.
```
JakeChampion: Could these instead live in a local `.env` file?

Gruntfile.js
triblondon: Yes, they're in Lastpass - created by Cloud enablement team today, and configuring them is documented in the changes I made to README.md.  I've also added a TODO to the PR OP to share the lastpass secret appropriately.

README.md
```diff
@@ -85,8 +75,40 @@ grunt ci             # run the service tests and polyfills tests on a large set
...
+2. Set the `RUM_MYSQL_DSN` and/or `RUM_MYSQL_DSN_QA` environment variables in your own local environment (you can also set them in `.env`).  FT devs with Heroku privs can get the relevant values with `heroku config` against both the QA and prod apps.
```
JakeChampion: We should store the values inside LastPass.

README.md
```diff
@@ -85,8 +75,40 @@ grunt ci             # run the service tests and polyfills tests on a large set
...
+3. Run `grunt shell:deployrumlambda:qa` or `grunt shell:deployrumlambda:prod` as appropriate
+4. If this is the first time you've deployed the function to this AWS profile, you then need to configure the function in the AWS UI:
```
JakeChampion, line 101: It's a shame this can't be done via an API of some kind.

service/routes/docs.js
triblondon: Which should be fine - it's tested for whenver it is used (because it is also null _until_ a connection is made)

docs/api.html
```diff
@@ -16,7 +16,7 @@
 				<li>There are changes to the Node API.  See the README for details.</li>
 			</ul>
 
-			<h2>GET /v{{apiversion}}/polyfill<var>:minify</var>.<var>:type</var></h2>
+			<h2>GET /v2/polyfill<var>:minify</var>.<var>:type</var></h2>
```
JakeChampion, line 19: 👍

service/routes/docs.js
triblondon: We should both read down before we comment, it seems :-)

tasks/lambda/project.json
triblondon: I think so - it's not a credential.  I don't think Apex provides an option to configure roles using env vars.

docs/api.html
```diff
@@ -83,6 +83,53 @@
 					<td>Set to <code>1</code> to enable <a href='https://en.wikipedia.org/wiki/Real_user_monitoring'>real user monitoring</a>, allowing the polyfill service to gather performance data about itself using the <a href='https://developer.mozilla.org/en-US/docs/Web/API/Resource_Timing_API/Using_the_Resource_Timing_API'>resource timing API</a>, where available, and to perform feature detection to improve our browser targeting.  Turning this on will help us improve the service, and tune our browser targeting to better serve your particular users.  Defaults to 0.  In a future release this will be on by default, so if you want to opt out permanently, set it explicitly to <code>0</code> now.</td>
 				</tr>
 			</table>
+
+			<h2>GET /v2/getRumPerfData</h2>
```
JakeChampion, line 88: Is it possible to not show these docs if the `RUM_MYSQL_DSN` is not set?

tasks/lambda/project.json
triblondon: There's precedent...

https://github.com/Financial-Times/next-url-management-api/blob/master/project.json

README.md
triblondon: I think line-number-anchored links are highly fragile, so would prefer not to.

triblondon: Unless we linked to the file at a commit ref

README.md
triblondon: only a small portion of this is actually AWS, the rest is node and heroku.  My preference is not to make it more complicated (or learn another tool, admittedly)

README.md
triblondon: Yes, it's a devdep

README.md
triblondon: Possibly.  I'll look into that.

docs/api.html
triblondon: Although the docs are versioned, we currently only support v2, so this makes sense,but we should retain the ability to have separate docs for v2 and a potential v3.

docs/api.html
triblondon: Good point, wilco.

service/routes/docs.js
```diff
@@ -299,7 +341,10 @@ function route(req, res, next) {
 		.then(locals => {
 			template(locals.pageName)
 				.then(templFn => res.send(templFn(locals)))
-				.catch(() => next())
+				.catch(err => {
+					console.log(err);
+					next();
```
rowanmanning, line 347: Just checking we want to ignore this error. Fine if we do, I'm not _that_ familiar with this code

README.md
JakeChampion: I can't see it in the dev-deps, are you sure? I can see `grunt-shell` was added but not `apex`.

JakeChampion:
Is it possible to cover these new features with tests?

triblondon:
Updated:

> there is still a hardcoded AWS role in the Apex project.json file

There's precedent for this, and @matthew-andrews recommended using Apex's 'environments' feature, which I've done

> Deploying the Lambda functions requires the AWS credentials to be in your `~/.aws/credentials` file as `polyfill` and `polyfill-qa` profiles, and for the RUM_MYSQL_DSN and RUM_MYSQL_DSN_QA env vars to be defined in your `.env` file (possible solution from @JakeChampion: use env vars on command line when invoking `apex deploy`)

Used Jake's suggestion.

> HTTP API docs are shown even if RUM env var is not set.  Should probably suppress these.

Done.

triblondon:
Regarding tests, what do you want to test, and how?  I'm not bothered about breaking RUM.  I appreciate that we don't want to break the service, but that should already be tested.

triblondon:
> The AWS credentials are in Lastpass but they need sharing with the origami team.

Done

JakeChampion:
I think we should be bothered about breaking RUM, if in the future we are going to use it to inform us what features each browser supports. It is easier to add tests for it now, before it is in master than later on down the road.

triblondon:
Agree that if we are using RUM to directly affect targeting automatically, it needs to be tested and monitored, but right now it's an experiment.  I just want to collect some data and see if it makes sense.

Again, I don't know how you'd test it effectively.  The most serious failure is taking down the service (already covered by tests) and I'd say the second most serious is producing incorrect data, and it's impossible to know if the reported data is right until we collect some.  I'm open to suggestions though - what tests would you suggest we implement before merging?

JakeChampion:
unit level tests would be good, we can worry over intergration tests later

service/routes/docs.js
triblondon: Yes, most likely it's a 404 - an unmatched route.

triblondon:
I don't know how I would unit test the RUM reporting routes, they require MySQL and hook directly into Express router meaning they can really only be subject to an integration test. I could unit test the lambda function but I just don't really care if it breaks, and it's totally isolated from heroku.  My tests would also probably not cover the failure scenarios because I don't know what they are yet.  The function leaves behind any files it cannot process so I can monitor for unprocessable logs.

If you can highlight exactly what you want tested, I'll look at doing that.

Issue #846, opened 2016-08-22
UC Browser support; general config question


zslabs:
Hi all! Bringing this up here to get a little clarification on support for [UC Browser](https://en.wikipedia.org/wiki/UC_Browser) since I'm not seeing other issues with that (keyword) mentioned. Has this been brought up before? Is there anything in the service that would allow us to pass through some default polyfills for browsers that are not recognized?

mudcube:
I believe this is the appropriate documentation:
https://polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers
Adding UC Browser support would be a nice addition.

zslabs:
Ah ok great! It looks like UC Browser is normally expressed in the UA agent as as `UCBrowser` according to https://www.whatismybrowser.com/developers/tools/user-agent-parser/browse/browser-name/uc-browser-user-agents 

@JakeChampion Since my tests for support with UC Browser seem to align somewhat close to IE11 (as far as `fetch` and `closest` support), how would we go about mimicking a browser to act as another? IE10 may even be the safer bet based on older versions of UC Browser, but I'm not 100% on coverage there. https://github.com/ua-parser/uap-core/blob/master/regexes.yaml#L110

triblondon:
@zslabs try here:

https://github.com/Financial-Times/polyfill-service/blob/master/lib/UA.js#L46

zslabs:
Submitted #853 - but it's failing and I'm a little unsure why. Have a minute to look? From my tests, IE10 is a close equivalent to UC Browser.

PR #853, opened 2016-08-29
Add UC Browser - mimic IE

zslabs:
Related: https://github.com/Financial-Times/polyfill-service/issues/846

Grabbing name from https://github.com/ua-parser/uap-core/blob/master/regexes.yaml#L113

Using caniuse.com I've found that a majority of items UC Browser does not support closely match IE and am proposing this browser to act as IE (IE10 more specifically) on recent versions of UC Browser `9.9`.

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -73,7 +73,7 @@ const aliases = {
 	"internet explorer": "ie",
 	"edge": "ie",
 	"edge mobile": "ie",
-	"uc browser": ["ie", 10],
+	"uc browser": "ie",
 
 	"chrome mobile ios": "ios_chr",
 
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -73,7 +73,7 @@ const aliases = {
 	"internet explorer": "ie",
 	"edge": "ie",
 	"edge mobile": "ie",
-	"uc browser": "ie",
+	"uc browser": ["ie", 10],
 
 	"chrome mobile ios": "ios_chr",
 
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -73,7 +73,9 @@ const aliases = {
 	"internet explorer": "ie",
 	"edge": "ie",
 	"edge mobile": "ie",
-	"uc browser": ["ie", 10],
+	"uc browser": {
+		"9.9.*": ["ie", 10]
+	},
 
 	"chrome mobile ios": "ios_chr",
 
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -73,6 +73,7 @@ const aliases = {
 	"internet explorer": "ie",
 	"edge": "ie",
 	"edge mobile": "ie",
+	"uc browser": ["ie", 10],
 
 	"chrome mobile ios": "ios_chr",
 


JakeChampion:
Hi @zslabs, thanks for taking the time to create a pull request to this project, I'm always happy to see new faces offering to help out :-)

You can ignore the CI issues, they are happening because your fork doesn't contain SauceLabs environment variables.

Are you able to test this on UC browser and confirm your assumption that we can alias UC to IE? It would help us have confidence that this is in fact true.

I also need to check that you have read and agreed to our [contribution terms](https://cdn.polyfill.io/v2/docs/contributing#contribution-terms) (pasted inline for convenience):

> All contributions to the polyfill service are made under the MIT licence. It is important therefore that you agree to the following when contributing code to the service:
> - Any contribution I make to the polyfill service by way of Pull Request contains only data that is:
>   - wholly my own work, upon which I am legally permitted to grant a licence; or
>   - wholly the work of individuals within the same organisation to which the copyright in all work created by employees is assigned, of which I am a current employee and authorised representive for the purposes of this contribution, and which is legally permitted to grant a licence; or
>   - licensed by the original author under the MIT licence or equivalent terms
> - In respect of any part of my contribution that is not already available under an MIT or compatible licence, I hereby grant to the Financial Times Limited (FT) and to recipients of software distributed by the FT a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable license to make, have made, use, offer to sell, sell, import, and otherwise transfer the data.
> 
> It is helpful to us if you can clearly state your agreement to the above terms when creating a pull request. If you don't, you may be asked to post a followup comment to accept the terms, before we can merge your PR.

Thanks again for this, if there is anything I can help with, feel free to ask.

zslabs:
Hi @JakeChampion, sure - I'm fine with the above terms. In terms of the coverage, I used caniuse.com to run through as many properties as I could and related them more closely to IE than any other browser. I don't have UC Browser personally, but it's come up enough times on Bugsnag that I figured it'd be a good one to patch if possible ;) so I'm leaning on caniuse.com to help.

triblondon:
@zslabs Hey Zack.  Currently your change will map every version of uc browser to IE10 (including all future versions of UC Browser).  That seems like it might not be a great match.  I think a couple of specific version mappings would be OK, but as proposed I think this change would make targeting worse over time (because the gap between UC latest and IE10 would widen)

zslabs:
Hey @triblondon yeah I 100% agree. The version area supports wildcards, correct? I've added in a more specific version-range based on the research I was able to compile with caniuse.com

Current usage of 9.9 is at 49% according to the aforementioned, so hopefully this would help a few others out 😄 

Thanks!

triblondon:
@zslabs can you screenshot and paste the docs that you're using as a reference for those figures?  And also a sample uc browser UA string which we can use to test the browser identification?

zslabs:
Hey @triblondon,
While pasting a screen shot of all of the comparables between the two browsers would be pretty darn big, if you visit http://www.caniuse.com and click the "Settings" link at the top, you can turn on/off specific browsers for a more detail oriented look between the two.

Here's a screen shot from the page for version usage:

screen shot 2016-09-08 at 1 29 37 pm

[This site](http://www.webapps-online.com/online-tools/user-agent-strings/dv/browser212364/uc-browser) has some example UA headers that can be used for testing, with a 9 example being `Mozilla/5.0 (Linux; U; Android 4.4.2; en-us; Aqua 3G Build/Aqua_3G) AppleWebKit/533.1 (KHTML, like Gecko) Version/4.0 UCBrowser/9.9.0.459 U3/0.8.0 Mobile Safari/533.1` taken from an Android 4.2 device.

triblondon:
OK.  I tweeted at UC to ask them if this seems like a reasonable mapping to them:

https://twitter.com/triblondon/status/774065264736415745

If they are OK with it or don't respond then I'm happy to merge it.

zslabs:
Hi @triblondon,
It doesn't look like they've responded yet - how are we feeling about this one?

JakeChampion:
@zslabs This is now live on http://qa.polyfill.io - would you be able to report back whether we have fixed your issue?

PR #856, opened 2016-08-31
move grunt tasks to grunt folder

JakeChampion:
Right now the line, `grunt.loadTasks('tasks');` will execute the `purge.js` task which it shouldn't be doing.

Moving all grunt tasks to a grunt folder will mean that the `purge.js` file is not present for grunt to accidentally execute.


triblondon:
I think the change in the rum-lambda branch is a better option than this - we moved grunt into tasks/grunt and lambda into tasks/lambda

JakeChampion:
Good point, your PR should be merged soon, I will close this in anticipation.

PR #864, opened 2016-09-08
Change from ESI to cookie for data centre detection

triblondon:
Fastly cannot enable ESI and also gzip responses.  That sucks, and we can't sacrifice gzip.  So instead of splicing the data centre ID into the response body using ESI, set a 1 minute cookie with the value.  Read the cookie when receiving the beacon request, falling back to the ID of the data center processing the beacon which is _probably_ the same.

### DIFF
--- a/fastly-config.vcl
+++ b/fastly-config.vcl
@@ -20,7 +20,7 @@ sub vcl_recv {
 	}
 
 	if (req.url ~ "^/v2/recordRumData" && req.http.Normalized-User-Agent) {
-		set req.http.Log = regsub(req.url, "^.*?\?(.*)$", "\1") "&ip=" client.ip "&refer_domain=" regsub(req.http.Referer, "^(https?\:\/\/)?(www\.)?(.+?)(\:\d+)?([\/\?].*)?$", "\3") "&country=" geoip.country_code;
+		set req.http.Log = regsub(req.url, "^.*?\?(.*)$", "\1") "&ip=" client.ip "&refer_domain=" regsub(req.http.Referer, "^(https?\:\/\/)?(www\.)?(.+?)(\:\d+)?([\/\?].*)?$", "\3") "&country=" geoip.country_code "&data_center=" if(req.http.Cookie:FastlyDC, req.http.Cookie:FastlyDC, server.datacenter);
 		error 204 "No Content";
 	}
 
@@ -32,11 +32,6 @@ sub vcl_recv {
 	set req.url = boltsort.sort(req.url);
 
 	return(lookup);
-}
-
-sub vcl_fetch {
-#FASTLY fetch
-	esi;
 }
 
 sub vcl_deliver {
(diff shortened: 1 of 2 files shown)


fastly-config.vcl
```diff
@@ -20,25 +20,15 @@ sub vcl_recv {
 	}
 
 	if (req.url ~ "^/v2/recordRumData" && req.http.Normalized-User-Agent) {
-		set req.http.Log = regsub(req.url, "^.*?\?(.*)$", "\1") "&ip=" client.ip "&refer_domain=" regsub(req.http.Referer, "^(https?\:\/\/)?(www\.)?(.+?)(\:\d+)?([\/\?].*)?$", "\3") "&country=" geoip.country_code;
+		set req.http.Log = regsub(req.url, "^.*?\?(.*)$", "\1") "&ip=" client.ip "&refer_domain=" regsub(req.http.Referer, "^(https?\:\/\/)?(www\.)?(.+?)(\:\d+)?([\/\?].*)?$", "\3") "&country=" geoip.country_code "&data_center=" if(req.http.Cookie:FastlyDC, req.http.Cookie:FastlyDC, server.datacenter);
```
JakeChampion, line 23: I can't see a definition for the `if` function in the [VCL docs](https://www.varnish-cache.org/docs/4.1/reference/vcl.html#functions). Is that a Fastly specific function?

triblondon: hm, I don't know.  I can't remember where I learnt about it, possibly from Fastly's boilerplate VCL.

fastly-config.vcl
```diff
@@ -57,6 +47,11 @@ sub vcl_deliver {
 	} else if (req.url ~ "^/v\d/polyfill\..*[\?\&]ua=" && req.http.X-Orig-URL && req.http.X-Orig-URL !~ "[\?\&]ua=") {
 		add resp.http.Vary = "User-Agent";
 	}
+
+	if (req.url ~ "[\&\?]rum=1") {
+    	add resp.http.Set-Cookie = "FastlyDC=" server.datacenter "; Path=/; HttpOnly; max-age=60";
```
JakeChampion: Should we also add `secure` flag?

fastly-config.vcl
JakeChampion: It is Fastly specific -- https://docs.fastly.com/guides/vcl/miscellaneous-VCL-extensions#if

fastly-config.vcl
triblondon: Thought about it, but can't test it on qa if we do that.

fastly-config.vcl
```diff
@@ -57,6 +47,11 @@ sub vcl_deliver {
 	} else if (req.url ~ "^/v\d/polyfill\..*[\?\&]ua=" && req.http.X-Orig-URL && req.http.X-Orig-URL !~ "[\?\&]ua=") {
 		add resp.http.Vary = "User-Agent";
 	}
+
+	if (req.url ~ "[\&\?]rum=1") {
+		add resp.http.Set-Cookie = "FastlyDC=" server.datacenter "; Path=/; HttpOnly; max-age=60";
```
JakeChampion, line 52: My comment disappeared. Should we add `secure` flag here aswell?

fastly-config.vcl
JakeChampion: I think we should get qa on https, Next can't test our qa as their csp policy doesn't allow http resources.

JakeChampion:
LGTM 👍

PR #865, opened 2016-09-08
Serve Fetch polyfill to Edge 14

hannesj:
Currently we are seeing quite a lot of `'fetch' is undefined` errors on our site due to older (supported) versions of EDGE not having fetch. As per [the edge blog](https://blogs.windows.com/msedgedev/2016/05/24/fetch-and-xhr-limitations/#m82TdcSqlraBgqeu.97) you need to have at least build 14342, while there are lots of machines still around with older builds. I'll suggest that we only enable it for EDGE 15 as the user agent library does not provide a more exact version string.

### DIFF
--- a/polyfills/fetch/config.json
+++ b/polyfills/fetch/config.json
@@ -2,7 +2,7 @@
 	"browsers": {
 		"android": "* - 4.4.4",
 		"bb": "10 - *",
-		"ie": "*",
+		"ie": "< 15",
 		"ie_mob": "*",
 		"firefox": "* - 38",
 		"opera": "* - 28",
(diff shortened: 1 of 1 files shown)


JakeChampion:
@hannesj Could you change it from `*` to `< 15`?

hannesj:
@JakeChampion sure, done.

triblondon:
Thanks.  We'll merge this for 3.13.  We're just pushing 3.12 out the door at the moment

samuliheljo:
FYI there is also this:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8475223/

JakeChampion:
@hannesj This is now live on http://qa.polyfill.io - would you be able to report back whether we have fixed your issue?

PR #872, opened 2016-09-14
Fix the validation order

falsandtru:


JakeChampion:
@falsandtru What prompted this change? Is it a spec violation?

falsandtru:
`!(descriptor instanceof Object)` check never works because of `'value' in descriptor`. So that validation is moved to work before using `descriptor`.

JakeChampion:
@falsandtru Could you add a test to cover this change to ensure we don't regress on this?

falsandtru:
Fixed.

JakeChampion:
@falsandtru I refactored your test to use proclaim. It is currently running the suite on CI :-)

PR #873, opened 2016-09-14
Support media queries without closing parentheses

srolel:
Firefox, Chrome, and Edge all support this case, it might be in the spec as well but I haven't checked. Other polyfills support it as well.


JakeChampion:
Hi @Mosho1, would it be possible for you to add a test for this new feature? Thanks for the PR :-)

srolel:
Looks like this is a bit more involved, I'll review and make the additional changes later today, and add tests.

srolel:
I've changed the implementation of the fix and added tests.

JakeChampion:
Running the test suite :-)

srolel:
Now that I think about it, why not use the widespread polyfill at https://github.com/paulirish/matchMedia.js?

JakeChampion:
@Mosho1 We do offer the ability to use polyfills from npm, here is an example of how it is done -- https://github.com/Financial-Times/polyfill-service/pull/805/files

We also have documentation on authoring polyfills -- https://polyfill.io/v2/docs/contributing/authoring-polyfills

triblondon:
I'm going to close this one - it's rather old, and a quick look at the spec does not turn up any suggestion that this parsing behaviour is part of the standard.

Issue #881, opened 2016-09-16
Cannot modify non-writable property 'iterator'


bhandfast:
Recently we've seen lots of error reports from Edge. The error is reported to the console as
SCRIPT5079: Cannot modify non-writable property 'iterator': polyfill.min.js (3,3694)

We load polyfill as:
https://cdn.polyfill.io/v2/polyfill.min.js?features=modernizr:es5array,es6

I did a search in the issues list but I didn't find anything relevant, but I might have overlooked something.

srolel:
I'm seeing this as well. `Symbol.iterator` is required by the `es6` polyfill, Edge does have it. The `Symbol` has `"ie": "< 13"` but `Symbol.iterator` has `"ie": "*"``, and I'm guessing that it is getting required as a dependency of`es6`directly, since`Symbol` isn't included in the Edge polyfill.

srolel:
I submitted a pull request, under the assumption that i was correct in my assertion above. This currently breaks the polyfill and apps that use it for es6 on Edge.

triblondon:
Thanks.  Should be released in 3.13.

bhandfast:
Any ETA when 3.13 will be pushed to the CDN?

JakeChampion:
It will head to qa.polyfill.io tomorrow and if no issues are reported for 7
days then it is promoted to production. If all goes smoothly it will hit
production next Friday.

bhandfast:
Ok, thanks for the info!

JakeChampion:
@bhandfast This is now released onto http://qa.polyfill.io, apologies it took longer than I expected.

PR #882, opened 2016-09-17
Don't serve Symbol polyfills for Edge 13 and above

srolel:
(possibly) solves #881


JakeChampion (approved):
https://kangax.github.io/compat-table/es6/#test-Symbol
LGTM

PR #883, opened 2016-09-19
Missing a space in an exception message.

rjray:
Several cases of a missing space in error messages for exceptions. These occurred where a callback parameter was being tested to ensure that it is a function.


rjray:
Having examined the logs of the failure, it does not appear to be related to my patch.

JakeChampion (approved):

JakeChampion (approved):

PR #884, opened 2016-09-19
Update Array.from to work with all types of iterators

JakeChampion:
This PR enables iterators of all types to work with `Array.from`, whether they be custom iterators, Sets, Maps, iterators passed from a different realm such as a Web Worker, and iterators from old version of Firefox and old versions of Safari.

This PR also enables Strings to be iterated correctly as well, thanks to the work by @falsandtru.

### DIFF
--- a/polyfills/Array/from/polyfill.js
+++ b/polyfills/Array/from/polyfill.js
@@ -1,104 +1,205 @@
+(function(){
+	'use strict';
 
-// Wrapped in IIFE to prevent leaking to global scope.
-(function () {
-	function parseIterable (arraylike) {
+	function toObject(value) {
+		if (value == null) {
+			throw new TypeError('Cannot call method on ' + value);
+		}
+		return Object(value);
+	}
+	function toLength(argument) {
+		var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
+		var number = Number(argument);
+		if (isNaN(number)) {
+			number = 0;
+		}
+		var len = (number >= 0 ? 1 : -1) * Math.floor(Math.abs(number));
+		if (len <= 0) { return 0; } // includes converting -0 to +0
+		if (len > MAX_SAFE_INTEGER) { return MAX_SAFE_INTEGER; }
+		return len;
+	}
+	function isCallable(value) {
+		return typeof value !== 'function' && typeof value !== 'object';
+	};
+	function isString (value) {
+		return typeof value === 'string';
+	}
+
+	var parseIterable = function (iterator) {
 		var done = false;
 		var iterableResponse;
 		var tempArray = [];
 
-		// if the iterable doesn't have next;
-		// it is an iterable if 'next' is a function but it has not been defined on
-		// the object itself.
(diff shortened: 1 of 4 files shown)


JakeChampion:
@falsandtru - would you like to review this? This is the Array.from polyfill from upstream but with modifications to enable it to work with Symbol polyfills.

polyfills/Array/from/polyfill.js
```diff
@@ -1,104 +1,205 @@
...
+			}
+		}());
 
-			// something went wrong return false;
-			return false;
+		if (supportsStrIterator) {
+			iteratorSymbol = '@@iterator';
+		} else {
+			try {
+				if (Function('var s = new Set(); s.add(0); for (var x of s) return x;')() === 0) { // eslint-disable-line no-new-func
```
ruiaraujo, line 100: Be aware that pages with CSP enabled, this would not work?
Can this run on browsers that support CSP?

triblondon: Agree. We should avoid using the `new Function` form if possible. There was an issue some time ago to remove all existing instances.

However, it's ok if the polyfill is only required in browsers that also don't support CSP. This should be marked with a comment.

JakeChampion: Yes, you will only enter this branch if you are Safari < 9

JakeChampion: This is no other way to polyfill this correctly for those browsers without using either `new Function` or `eval`.

JakeChampion: Actually, this will run in any browsers which don't have Symbol nor have an '@@iterator' property (Firefox), so this runs in lots of browsers.

polyfills/Array/from/polyfill.js
```diff
@@ -1,104 +1,205 @@
...
+		var len = (number >= 0 ? 1 : -1) * Math.floor(Math.abs(number));
+		if (len <= 0) { return 0; } // includes converting -0 to +0
+		if (len > MAX_SAFE_INTEGER) { return MAX_SAFE_INTEGER; }
+		return len;
+	}
+	function isCallable(value) {
+		return typeof value === 'function';
+	};
+	function isString (value) {
+		return typeof value === 'string';
```
falsandtru, line 25: This check need to accept `Object('str')` and similar objects. https://github.com/ljharb/is-string

falsandtru:
Can we know which tests are failed in ci (Safari 8) ?

falsandtru:
I guess Safari 8 has 2 iterations that use `Symbol.iterator` polyfill or for-of statement. Probably it is a reason of failures in ci tests. We need to choose one of those 2 iterations dynamically.

polyfills/Array/from/polyfill.js
triblondon: it looks like every function stirng is in a try...catch, so that's something, but it would be worth testing in a browser that fails the array.from detect, and supports CSP, with a CSP header present.

PR #888, opened 2016-09-20
Don't polyfill Array.prototype.includes for chrome 47 or higher -- Fixes #887

### LINKED ISSUE #887: Array.prototype.includes is added even on browsers that have it
I've tried this link https://cdn.polyfill.io/v2/polyfill.js?features=Array.prototype.includes with Chrome 52

```
window.navigator.userAgent
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36"
```

```
!!Array.prototype.includes
true
```

The response is

``` js
/* Polyfill service v3.12.1
 * 
 * UA detected: chrome/52.0.0
 * Features requested: Array.prototype.includes
 * 
 * - Array.prototype.includes, License: CC0 */

(function(undefined) {

// Array.prototype.includes
Array.prototype.includes = function includes(searchElement /*, fromIndex*/ ) {'use strict';
    var O = Object(this);
    var len = parseInt(O.length) || 0;
    if (len === 0) {
        return false;
    }
    var n = parseInt(arguments[1]) || 0;
    var k;
    if (n >= 0) {
        k = n;
    } else {
        k = len + n;
        if (k < 0) {k = 0;}
    }
    var currentElement;
    while (k < len) {
        currentElement = O[k];
        if (searchElement === currentElement ||
           (searchElement !== searchElement && currentElement !== currentElement)) {
            return true;
        }
        k++;
    }
    return false;
};

})
.call('object' === typeof window && window || 'object' === typeof self && self || 'object' === typeof global && global || {});
```

Apparently [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) is available since Chrome 47.

Works fine in Firefox, that link gives an empty response.

JakeChampion:


rowanmanning (approved):
Looks good 👍

PR #890, opened 2016-09-20
Reinstate RUM improvements with fix for gated flag

triblondon:
image

Also added a test to ensure that the `gated` flag causes the output to be different.  It fails before the fix is applied.


triblondon:
I apologise unreservedly for the shameless meme use.

docs/usage.html
```diff
@@ -18,87 +27,89 @@
...
+			<td class='rum-perf-data__nettimeline'><div class="nettimeline nettimeline--dns" style="width:{{percent perf_dns_95 @root.rumPerf.scaleMax}}%" title="DNS lookup: {{perf_dns_95}}ms, {{perf_dns_count}} non-zero samples"></div><div class="nettimeline nettimeline--connect" style="width:{{percent perf_connect_95 @root.rumPerf.scaleMax}}%" title="TCP connect / SSL: {{perf_connect_95}}ms, {{perf_connect_count}} non-zero samples"></div><div class="nettimeline nettimeline--req" style="width:{{percent perf_req_95 @root.rumPerf.scaleMax}}%" title="Request: {{perf_req_95}}ms, {{perf_req_count}} non-zero samples"></div><div class="nettimeline nettimeline--resp" style="width:{{percent perf_resp_95 @root.rumPerf.scaleMax}}%" title="Response: {{perf_resp_95}}ms, {{perf_resp_count}} non-zero samples"></div></td>
```
JakeChampion, line 104: This kind of scares me.

lib/index.js
```diff
@@ -132,17 +133,27 @@ function getPolyfillString(options) {
 					explainerComment.push('');
 				}
 			}
-			return ((!ua.meetsBaseline() || ua.isUnknown()) && options.unknown !== 'polyfill') ? {} : getPolyfills(options);
+			if ((!ua.meetsBaseline() || ua.isUnknown()) && options.unknown !== 'polyfill') {
+				return [{}, []];
+			} else {
+
+				// Turn on the 'always' flag to fetch the complete list of resolved features
+				const optionsForAll = cloneDeep(options);
```
JakeChampion, line 141: 👍

JakeChampion (approved):

docs/usage.html
triblondon: It would be useful to know in what way it scares you.

PR #899, opened 2016-09-30
Serve performance.now polyfill for iOS 8

### LINKED ISSUE #898: performance.now polyfill missing from iOS 8 Safari User Agent
As per [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now), iOS 9 is the first version of Safari with support for performance.now. However, when requesting `https://cdn.polyfill.io/v2/polyfill.js?features=performance.now` from iOS 8, I received the following in return:

```
/* Polyfill service v3.12.1
 * 
 * UA detected: ios_saf/8.0.0
 * Features requested: performance.now
 *  */

```

For now I set the 'always' flag on that feature as a work around, but it would be great to update to properly support iOS < 9.

JakeChampion:
Resolves #898


triblondon (approved):
The original targeting was very specific about skipping iOS8, but there was no mention of why in the PR https://github.com/Financial-Times/polyfill-service/pull/236.  Based on MDN I'm happy with this.

Issue #844, opened 2016-08-22
Incorrectness of Array.from with iterators


falsandtru:
As a note, the following implementation is probably wrong because `parseIterable` detects an iterable response that has a `next` method, not an iterable objects that has a `[Symbol.iterator]` method. In the following code, it should detect an iterable object.

``` js
arrayFromIterable = parseIterable(arraylike);
```

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Array/from/polyfill.js#L72

This problem will be resolved by https://github.com/mathiasbynens/Array.from/pull/27.

falsandtru:
Add: Array.from polyfill still broken with Map/Set polyfills because Array.from checks `value` prop in https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Array/from/polyfill.js#L16 but Map/Set don't provide that prop in https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Map/polyfill.js#L27. It happened with my lib. So #817 is incomplete.

@JakeChampion Can I use the upstream polyfill with this service in a few week, or should I fix it on this repo?

falsandtru:
@triblondon @JakeChampion Could you review #840? I reopened that pr to fix this bug.

triblondon:
So now #840 is closed, what's the status of this problem now?

falsandtru:
This will be fixed by #884. Thanks for your pointing.

JakeChampion:
Just to update this issue, #884 was closed

falsandtru:
Thanks, the problem I found on Android 4.4 is fixed.

PR #901, opened 2016-09-30
Map/Set IteratorResults now always have a value

zetlen:
IteratorResults returned from `map.keys()`, `map.entries()`, and
`map.values()` used to return an object `{done: true}`, with no `value`
property, when they were done. However, that triggered a bug in
`Array.from` implementations that didn't conform to the spec. This bug
exists in native and polyfilled implementations--including [this one](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Array/from/polyfill.js#L16)!

The bug causes Array.from to convert an affected iterator into an empty
array, [demonstrated here](http://jsbin.com/jaliguyaqe/3/edit?html,js,console).

It IS a bug in Array.from and not in this implementation, in fact: the
[IteratorResult](http://www.ecma-international.org/ecma-262/6.0/#sec-iteratorresult-interface) interface specifies that the `value` property
"may be absent from the conforming object", but unfortunately, both
native and polyfilled implementations of Array.from don't respect that
(at least in Chrome, which is a popular browser, I'm led to believe.)

### DIFF
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -24,7 +24,7 @@
 					while (mapInst._keys[nextIdx] === undefMarker) nextIdx++;
 					return {value: getter.call(mapInst, nextIdx++), done: false};
 				} else {
-					return {done:true};
+					return {value: void 0, done:true};
 				}
 			}
 		};
--- a/polyfills/Map/tests.js
+++ b/polyfills/Map/tests.js
@@ -141,6 +141,12 @@ it("exhibits correct iterator behaviour", function () {
 	// new element shows up in iterators that didn't yet finish
 	proclaim.equal(entriesagain.next().value[0], "4");
 	proclaim.equal(entriesagain.next().done, true);
+	// value is present but undefined when done is true, so that Array.from and other noncompliant
+	// interfaces recognize it as a valid iterator
+	var lastResult = entriesagain.next();
+	proclaim.equal(lastResult.done, true);
+	proclaim.ok(lastResult.hasOwnProperty('value'));
+	proclaim.equal(lastResult.value, void 0);
 });
 
 it("implements .forEach()", function () {
--- a/polyfills/Set/polyfill.js
+++ b/polyfills/Set/polyfill.js
@@ -24,7 +24,7 @@
 					while (setInst._values[nextIdx] === undefMarker) nextIdx++;
 					return {value: getter.call(setInst, nextIdx++), done: false};
 				} else {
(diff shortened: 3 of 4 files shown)


zetlen:
(I have access to a SauceLabs instance, so I was able to run a full suite, unlike our friend Circle here.)

falsandtru:
It will be fixes for #816 and #844.

triblondon (approved):
> (at least in Chrome, which is a popular browser, I'm led to believe.)

Kids these days, eh.  Nowt wrong with Lynx if you ask me.

LGTM.

falsandtru:
Can we use this fix in next version 3.13?

JakeChampion:
This is now live on http://qa.polyfill.io - would you be able to report back whether we have fixed your issue?

zetlen:
@JakeChampion Confirmed in IE11--your Array.from and Map polyfills work together like gangbusters. **However, the Map polyfill still does not work with Chrome's native `Array.from`.** I'm not sure why this is, and it's less of an issue because there are very few environments that support one but not the other. Nevertheless I thought I should mention it. Demonstration: http://jsbin.com/lefapefeje/1/edit?html,js,console

(Note that you'll have to accept the invalid SSL certificate for your QA host before the polyfill comes through at all.)

**Edit:** Firefox's native `Array.from` doesn't work with this fake iterator either.

falsandtru:
Seems like there are 2 problems.
1. `Symbol.iterator` is rewritten.
2. Some methods such as `.values()` return a not iterable value (not having a `[Symbol.iterator]()` method).

JakeChampion:
@zetlen Could you make a new issue which decribes the bug you are seeing? 

Thanks for this work by the way :-)

Did you know we offer one month's free premium access to all FT content (normally $50) to anyone who has a pull request merged in this repository? If you want to claim it, there's [one Google form](https://docs.google.com/a/ft.com/forms/d/1aykPiYBwqKpKY6nvGVi43wOU_5epr3XLgE00Vi52SMQ/viewform) to fill out :-)

zetlen:
@JakeChampion The work is its own reward, but I won't look a gift Financial Times in the mouth. Thanks a million! I'll open a separate issue.

zetlen:
@JakeChampion Just so you know, I just tried to produce a failing test for this within your existing framework. There's no browser environment which supports Array.from but not Map, so this is an extreme corner case. Still think it's worth a PR?

JakeChampion:
Hmm, probably not then, thanks for investigating :-)

PR #907, opened 2016-10-03
Change setImmediate detect to use `this` instead of `window`

JakeChampion:
This enables the detect to work in contexts which don't have a `window` object such as Worker contexts or NodeJS.


triblondon (approved):
Also in line with our authoring guidelines.

PR #913, opened 2016-10-03
Update intl and drop support for IE 7&8

JakeChampion:


JakeChampion:
It seems Intl no longer works with IE7/8.

```
Failed tests:
 - ie 8 (Sauce results: https://saucelabs.com/tests/d3d67a36ea7b413fb35bc7fffca19c9a)
    -> Intl
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Intl
 - ie 7 (Sauce results: https://saucelabs.com/tests/6e5ad0d70c5640c1b229aae333536168)
    -> Intl
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Intl
```

I imagine this isn't on purpose and probably we are missing a new polyfill dependency such as `Object.defineProperty`. @caridy, do you happen to know what would be causing the IE7/8 issues?

JakeChampion:
I think I found the Intl ticket for the IE7/8 issues we see on CI -- https://github.com/andyearnshaw/Intl.js/issues/189

caridy:
yeah, I haven't get much help on that! I think no many folks are using the polyfill in ie8, at least no one is complaining! Can someone help to determine what the problem is so we can get this merged?

triblondon:
@caridy FT doesn't support/run JS in IE8, but this service does support those browsers, so we have a similar prioritisation problem.  We can't merge the update knowing that there are service users who depend on us fixing IE8, but we can't invest FT time in fixing Intl issues in that browser.  If no-one comes forward wanting to fix those issues, we can just drop the targeting of IE 7 and 8, so the polyfill will only be delivered in 9+.

caridy:
@triblondon we can try that, and see how it goes. If there are people asking for it, we can revisit.

@JakeChampion my question is still valid here, why is the config pulling `Object.defineProperty`?

JakeChampion:
@caridy it may not require it if the `Intl` polyfill brings it's own `Object.defineProperty` polyfill.

caridy (approved):

rowanmanning (approved):

PR #914, opened 2016-10-04
Polyfill console.info and console.error for IE Mobile <= 9

### LINKED ISSUE #910: IE Mobile 9 lacks support for console.info (and is not being polyfilled)
I wasn't able to find any official reference on which versions of IE Mobile support the console APIs, but I have found that a device with user agent `Mozilla/5.0 (compatible; MSIE 9.0; Windows Phone OS 7.5; Trident/5.0; IEMobile/9.0; LG; LG-C900)` does not have support for the `console.info` function. When requesting it explicitly on Polyfill.io I got back the following:

```
/* Polyfill service v3.12.1
 * 
 * UA detected: ie_mob/9.0.0
 * Features requested: console.info
 *  */
```

I'm assuming `console.error` and most other console functions also don't work on this platform, but I have not yet been able to test this.

rubencodes:
Resolves https://github.com/Financial-Times/polyfill-service/issues/910


polyfills/console/error/config.json
```diff
@@ -4,6 +4,7 @@
 	],
 	"browsers": {
 		"ie": "6 - 11",
+		"ie_mob": "* - 9",
```
triblondon: For this kind of range we prefer `<10`.  This is because 10 is the significant number - the version in which the feature was first supported.  `0 - 9` requires you to remember a) that semver requires spaces around the `-`, which is a gotcha, and b) that 9 is the last version that does not have the feature, not the first version that does.

triblondon (changes requested):
Syntax tweak and please also run through https://polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers, then we're good to go.

rubencodes:
Updated with the syntax tweaks requested. 

I wasn't able to find anything official from MDN, caniuse, or even Microsoft about IE Mobile's support (or lack thereof) for the console APIs, which is probably why it hadn't been included to begin with. However, I was able to confirm myself with a Windows Phone device that IE Mobile 9 does not support console.error nor console.info. I have no way of confirming if IE Mobile 10 does have support, so it is not being included in this polyfill.

I agree to the contribution terms.

JakeChampion (approved):

triblondon (approved):

triblondon:
Think this will have to wait for 3.14 now that we've shipped 3.13 to QA, but we can try and speed up the cycle time for the next release.

JakeChampion:
@rubencodes Would you be able to update this branch with the latest code from `master`? It seems they have fallen out of sync.

rubencodes:
Done!

PR #917, opened 2016-10-04
Remove console logging from sources module

### LINKED ISSUE #915: Remove console log.
Currently every time I start my app using this module I get the following printed to the console.
(from https://github.com/Financial-Times/polyfill-service/blob/4efb6aaf02ce7a94b87abf716c2d3f50efb2e724/lib/sources.js#L23)

``` console
Found 871 polyfills
```

It would be nice if this was hidden behind `NODE_ENV=debug` or better yet if you used the [debug](https://www.npmjs.com/package/debug) module. I don't think the log is very helpful anyways so it might be best to remove it.

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-service/issues/915

### DIFF
--- a/lib/sources.js
+++ b/lib/sources.js
@@ -20,7 +20,6 @@ try {
 		.map(fileName => fileName.replace(/\.json$/i, ''))
 	;
 	polyfillCount = features.length;
-	console.log("Found "+polyfillCount+" polyfills");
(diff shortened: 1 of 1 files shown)


rowanmanning (approved):
Looks good 👍

PR #918, opened 2016-10-04
polyfill symbol in BlackBerry

wheresrhys:
@JakeChampion being deliberately conservative with versions given the compatibility problems we've had with symbol before. FT only needs v10 covered. If others want to invest time in verifying Symbol and related polyfills don't blow up in BB<10 then config can be modified in future PRs

### DIFF
--- a/polyfills/Array/from/config.json
+++ b/polyfills/Array/from/config.json
@@ -19,7 +19,7 @@
 		"android": "*",
 		"ios_saf": "<9",
 		"samsung_mob": "*",
-		"bb": ">=10"
+		"bb": "10 - *"
 	},
 	"dependencies": [
 		"Object.defineProperty"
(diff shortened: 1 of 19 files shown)


JakeChampion:
@wheresrhys Do you have screenshots of the tests passing on BB10?

wheresrhys:
not of the tests, but ft.com is fixed by forcing `Array.from` and `Map` to be bundled on BB10, which is reasonable evidence that this is ok. I'll work on running the tests this afternoon though

polyfills/Array/from/config.json
```diff
@@ -18,7 +18,8 @@
 		"firefox_mob": "<32",
 		"android": "*",
 		"ios_saf": "<9",
-		"samsung_mob": "*"
+		"samsung_mob": "*",
+		"bb": ">=10"
```
JakeChampion: I think this needs to be `10 - *` to be valid semver.

JakeChampion (changes requested):

wheresrhys:
image

JakeChampion (approved):

JakeChampion:
@wheresrhys This is now live on http://qa.polyfill.io - would you be able to report back whether we have fixed your issue?

PR #925, opened 2016-10-10
Add examples for feature detection and async loading

triblondon:
Update our Examples page to include the feature-detection example I wrote for [this twitter thread](https://twitter.com/philwalton/status/780904078716964864)


docs/examples.html
```diff
@@ -49,12 +49,61 @@ <h2 id='flags'>Using flags</h2>
...
+('Promise' in window) || features.push('Promise');
+('IntersectionObserver' in window) || features.push('IntersectionObserver');
+('after' in Element.prototype) || features.push('Element.prototype.after');
+
+// If any features need to be polyfilled, construct
+// a script tag to load the polyfills and append it
+// to the document
+if (features.length) {
+  var s = document.createElement('script');
+  s.src = 'https://polyfill.io/v2/polyfill.min.js?features='+features.join(',')+'&flags=gated,always&ua=chrome/50&callback=main';
```
JakeChampion, line 95: Is the `ua=chrome/50` added to increase the likelihood that the response is in the CDN cache? Would it be better to not include a UA, that way the user doesn't need to change this src URL if we ever remove the UA they've added.

triblondon: It's because if you don't specify a UA, the CDN will do a normalisation and then add your UA to the URL.  So adding one yourself will skip that process and be an optimisation.  If you don't hard code a supported UA, then you would need to add `unknown=polyfill` instead, in case your users have browsers that are not in our supported list.

JakeChampion (approved):

docs/examples.html
JakeChampion: Should we add this as a comment to the code?

triblondon: Done in eef1764

triblondon:
@JakeChampion good to merge but will leave with you since you're managing the 3.13 release from master at the moment.

PR #867, opened 2016-09-10
Serve Element.prototype.closest to Android versions < 5

falsandtru:
image

PR #930, opened 2016-10-12
Serve HTMLPictureElement to android < 52

falsandtru:
http://caniuse.com/#search=picture


triblondon (changes requested):
I see:

image

So it should be `<52` based on that.

falsandtru:
Android 5 or higher use Chrome via WebView. You can see it by hovering the cursor over Android versions on that site. Probably your specified version is a Chrome version. So I think we need to set `<5` when Chrome versions are displayed as Android versions. Also see #867.

triblondon:
Right, but I think that in Android >=5, the webviews identify themselves as `android/<chrome_version_number>` not an android version number.

falsandtru:
I don't know how Android 5 actually behaves because I don't have that one. Can I see actual results in Android 5 like `android/51`?

triblondon:
Yes, I think so, but I don't have a handset handy either.  Can I ask what problem are you trying to solve here?

falsandtru:
At least, this pr will solve confusing of version schemas. I cannot know about which schema I should use. So we should have consistency of version schemas and shouldn't mix the different schemas like Android and Chrome.

triblondon:
The problem is we don't define the version scheme here - we get the version numbers from the ua-parser project, and they continued the android browser sequence with the chrome versions when Google switched to using Chrome.  So I think it's correct for us to jump from 4.4 to the chrome numbers,

falsandtru:
I think then we need to update android versions in all config.json if that android version is 5 or higher. Will  we do that?

JakeChampion:
@falsandtru We can do that. SauceLabs has Android 5.1 available for use, what is it you want to test?

falsandtru:
Thanks, I fixed it.

JakeChampion:
Thanks, rerunning tests.

triblondon (approved):

triblondon:
Looks like CI failed because of the missing Edge variant.  I think we updated those defs on master, so I pulled from master and am rerunning CI

Issue #397, opened 2015-05-09
MutationObserver polyfill


dy:
Hello!
There are some tasks related with handling element mutations, able to be polyfilled in IE8+.
There’s a polyfill-like lib, [mutation-observer](https://github.com/webmodules/mutation-observer), which does this, but there’s no such polyfill in the service.
Is it reasonable to add `MutationObserver` polyfill?
Thank you.

triblondon:
Yes, we'd very much welcome that.

matthewrobb:
Hello, curious if this is still in the pipe?

ghost:
Will MutationObserver be available?

For some reason google cache shows it on the feature list. Was it removed?

sheriff91:
Please add this, I'd love you forever! Is it in v3?

triblondon:
PR is #695, but it has some failing tests.  FT doesn't need it, so we're probably going to close it, but it's close, so if you have an interest, you could pick it up and finish it off?

sheriff91:
I can certainly have a gander. Especially considering this would be incredibly helpful for me. @triblondon Pulled the muOb branch.

triblondon:
@hipkiss91 are you still intending to work on this?

JakeChampion:
This was added in version 3.17 https://github.com/Financial-Times/polyfill-service/releases/tag/v3.17.0

PR #937, opened 2016-10-14
Mutationobserver Updated tests.js (#397)

sheriff91:
Hopefully fixed all code required to move from expect.js to proclaim.js


sheriff91:
Error:  SAUCE_USER_NAME and SAUCE_API_KEY must be set in the environment.

Where are these set? Is this also something you guys do? @JakeChampion @triblondon (Sorry about this.)

JakeChampion:
Yes, only users with push/write access to this repo can run the tests on CI annoyingly. I will do this now :-)

sheriff91:
@JakeChampion Does this mean my modifications have worked? The test shows 2 failures when testing on IE9 but no further information, and it timed out?

polyfills/MutationObserver/tests.js
```diff
@@ -1,18 +1,22 @@
 /* Assertion bridge between web components test suite and expectjs */
 
+
 var expectRecord = function (src, model) {
 	Object.keys(model).forEach(function(key) {
-		expect(src[key]).to.eql(model[key]);
+		// proclaim(src[key]).to.eql(model[key]);
+    proclaim(src[key], model[key]);
```
triblondon, line 7: There are some indentation issues here, @hipkiss91 could you configure your editor to respect our .editorconfig settings?  cc @JakeChampion

polyfills/MutationObserver/tests.js
```diff
@@ -47,15 +51,18 @@ describe('MutationObserver attributes', function() {
     div.setAttribute('a', 'B');
 
     var records = observer.takeRecords();
-    assert.strictEqual(records.length, 2);
+    // assert.strictEqual(records.length, 2);
+    proclaim.strictEqual(records.length, 2);
```
triblondon, line 55: You could remove the old-code comments

triblondon:
@hipkiss91 Jake probably merged this into the MO branch because that's an easy way to get our sauce tests to run on the right fork, and we're comfortable with being trigger-happy on merging into non-master branches.  Looks like following your changes we're failing the ESLint at the moment: https://circleci.com/gh/Financial-Times/polyfill-service/1156.

sheriff91:
@triblondon Thanks for the comments. Apologies on not removing the old code, I will do that now. In regard to the indentation - I use tabs of length 4. What length do you guys prefer? If it is 4, then I imagine I missed a tab, apologies.

triblondon:
@hipkiss91 we use editorconfig (http://editorconfig.org/) so your editor should configure itself automatically, but in short we should be using real tabs everywhere, not spaces.  You can see the inconsistency when you view the diff on Github, assuming your preferred tab size is < 8 spaces.  Some files may be incorrectly indented with spaces - I try and correct them when I can, but the most important thing is that we don't mix two styles in one file :-)

sheriff91:
@triblondon Of course - consistency is king. It should all read as though one person wrote it :) Any thoughts as to why the eslint error might occur? 
Also, once I've fixed it - do I just create another PR?

PR #939, opened 2016-10-18
Fix typo in url

dstarcev:


dstarcev:
It seems that CI fails on PRs made from forks. How do I contribute properly?

JakeChampion (approved):

JakeChampion:
@dstarcev Thanks for the PR! You are correct, forks can not run our CI tests as the tests rely on environment variables to be set for SauceLabs. These environment variables are protected as we would not like to leak them. To enable CI to run on PRs, polyfill.io maintainers first review the code for anything which may leak the API keys, if nothing harmful is found we bring the commits into the upstream repository and push to github which triggers CI to run once again (this time with the environment variables set!).

It is rather annoying to do this, I hope that we can come up with a better solution in the future for this.

PR #942, opened 2016-10-19
Redirect qa to https

JakeChampion:
@triblondon Is there anything else that would need to be changed to force qa to be https?


triblondon (changes requested):
Er, hmm. Doing this will redirect http://qa.polyfill.io -> https://polyfill.io.  I think we need a separate `if` to redirect https for the QA hostname.  Maybe put the two `if`s inside the check for `req.http.Fastly-SSL`?

fastly-config.vcl
```diff
@@ -10,8 +10,13 @@ sub vcl_recv {
...
 	}
 
-	if (!req.http.Fastly-SSL && (req.http.Host == "cdn.polyfill.io" || req.http.Host == "polyfill.io")) {
-		error 751 "Canonicalise";
+	if (!req.http.Fastly-SSL) {
+		if (req.http.Host == "cdn.polyfill.io" || req.http.Host == "polyfill.io") {
+			error 751 "Canonicalise";
+		}
+		if (req.http.Host == "qa.polyfill.io") {
+			error 752 "HTTPS";
```
triblondon: I think these error labels are a bit confusing now.  Maybe call them:

```
error 751 "Redirect to prod HTTPS"
error 752 "Redirect to QA HTTPS"
```

fastly-config.vcl
```diff
@@ -61,12 +66,21 @@ sub vcl_deliver {
...
 	}
+	
+	# Redirect qa to SSL
+	if (obj.status == 752) {
+		set obj.status = 301;
+		set obj.response = "Moved Permanently";
+		set obj.http.Location = "https://qa.polyfill.io" req.url;
+		synthetic {""};
+		return (deliver);
+	}
```
triblondon, line 76: You could make this a bit less repetitive:

```
# Redirect to canonical prod/qa origins
if (obj.status == 751 || obj.status == 752) {
    set obj.status = 301;
    set obj.response = "Moved Permanently";
    set obj.http.Location = "https://" if(obj.status == 752, "qa.", "") "polyfill.io" req.url;
    synthetic {""};
    return (deliver);
  }
```

triblondon (approved):
Suggestions are optional.

PR #944, opened 2016-10-20
Add support for Android for requestAnimationFrame

### LINKED ISSUE #704: requestAnimationFrame not polyfilling in Android stock browser
I've been getting inconsistent results with requestAnimationFrame being detected and loading in certain versions of the Android stock browser as part of the default set.  Specifying the property with |always fixes the issue, but otherwise I get a `Uncaught ReferenceError: requestAnimationFrame is not defined`.

A client reported JS not working properly on two Galaxy S5 devices running versions of Android 5 and 6, but the only place I've had a chance to test the problem/fix is the Galaxy S3 Android 4.1 emulator on Browserstack.  Not sure how useful that information is, but it's all I have to go on at the moment.  I'll post an update if I get any more information - figured it was worth passing my findings along.

ruiaraujo:
Based on the information from canIUse.
Fixes #704.

### DIFF
--- a/polyfills/performance/now/config.json
+++ b/polyfills/performance/now/config.json
@@ -1,5 +1,6 @@
 {
 	"browsers": {
+		"android": "<4",
 		"bb": "<10",
 		"chrome": "<24",
 		"ie": "<10",
--- a/polyfills/requestAnimationFrame/config.json
+++ b/polyfills/requestAnimationFrame/config.json
@@ -9,6 +9,7 @@
 		"modernizr:requestanimationframe"
 	],
 	"browsers": {
+		"android": "<4.4",
 		"bb": "7",
 		"chrome": "1 - 23",
 		"ie": "6 - 9",


triblondon:
Hi @ruiaraujo, can you give us a but more info here?
https://cdn.polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers

Particularly we need to know we're not going to break older versions of Android if they choke on the polyfill code.  If you have devices to hand and can run some tests, that would be very helpful.

ruiaraujo:
I have running with always|gated for a few months now and it seems to be fine.

But I will pick up a jsFiddle and test some old browsers.

Also, it seems I should also update the configs of the dependencies of rAF like performance.now since also has no android targets.

JakeChampion:
I tried to find an Android 3 device, physical or emulated. I have no access to a physical device and browserstack, saucelabs, crossbrowser-testing do not offer android 3 emulators, I couldn't find one to install on my machines either. I then looked at gs.statcounter.com data for android 3 and found 0% usage.

If anyone has an android 3 device, would you be able to visit [this page](https://polyfill.io/test/tests?feature=requestAnimationFrame) and post a screenshot of the result into this issue?

ruiaraujo:
I have visited that page on every Android version from 2.2 to 4.3. The test fail on Android 2.x.

Screenshot of the 3.0 emulator.
captura de ecra 2016-10-30 as 08 50 41

JakeChampion:
Thanks! The polyfill not working for 2.x is okay as our baseline support for android is `>=3`.

JakeChampion (approved):

Issue #941, opened 2016-10-19
Always serve URL to Edge


fredj:
From e54f901 Microsoft Edge no longer receives URL but the [searchParams property](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) is not implemented in this browser.

Would it be possible to revert e54f901?

Tested with (in the debugger console):

```
> u = new URL('https://github.com/Financial-Times/polyfill-service?foo=bar');
url.searchParams === undefined
```

JakeChampion:
@fredj PR for this - #946

fredj:
thanks !

PR #946, opened 2016-10-23
polyfill URL for all IEs and Edge browsers

JakeChampion:
Fixes issue #941

### DIFF
--- a/polyfills/URL/config.json
+++ b/polyfills/URL/config.json
@@ -10,7 +10,7 @@
 	"browsers": {
 		"chrome": "<49",
 		"firefox": "<29",
-		"ie": "9 - 12",
+		"ie": "9 - *",
 		"ie_mob": "*",
 		"safari": "*",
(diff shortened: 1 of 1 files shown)


triblondon:
Does the detect pass in Edge?  We should try to ensure the detect fails where we want the polyfill to take effect.

falsandtru:
I tried to run `'searchParams' in nativeURL` on Edge, then this code returns `false` correctly. And it would be tested by existing tests.

https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/URL/tests.js#L113

triblondon:
Does https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/URL/detect.js return false in Edge?  If so, the proposed change is fine.

triblondon:
Looks like detect fails in Edge, so we're good:

image

triblondon (approved):

PR #948, opened 2016-10-23
Remove detect from fetch polyfill

JakeChampion:
Edge 14 has a broken version of fetch, we target Edge 14 but fail to apply polyfill due to an embedded feature detect in the polyfill. Thanks to @wessberg for spotting this issue.

### DIFF
--- a/polyfills/fetch/polyfill.js
+++ b/polyfills/fetch/polyfill.js
@@ -3,10 +3,6 @@
 (function(self) {
   'use strict';
 
-  if (self.fetch) {
-    return;
-  }
-
   function normalizeName(name) {
     if (typeof name !== 'string') {
       name = String(name);


JakeChampion:
Linking this PR to the commit comment which made it happen -- https://github.com/Financial-Times/polyfill-service/commit/dc8331a7e04c84f2cced7a383ff863d48ccd780f#commitcomment-19533201

triblondon (commented):
This may still not have the desired effect if the user requests the polyfill 'gated' because our detect.js just contains `'fetch' in this`. Is there a better detect we could use (and push upstream into Github's version?

triblondon:
The detect that we would need to be sensitive to the broken implementation in Edge includes actually invoking a fetch.  This obviously has massive overhead, will cause console logs to be printed and could in any case violate CSP.   I tried some experiments with fetching a `data:` URI on the basis that we might get away with that, if it worked, but you can't POST to a data: URI.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8475223/ is the Edge bug.  To be honest I think there might not be anything we can do here.

nolanlawson:
To clarify: we shipped Edge 14 with one confirmed `fetch()` bug (DELETE requests return 0 for content-length and an empty body), as well as some missing features that are technically part of the `iterable` spec. There are some details [in this Reddit thread](https://www.reddit.com/r/webdev/comments/57ii4f/psa_edge_14_ships_a_broken_windowfetch/?st=iuoar07b&sh=15ed03f9); note that [this bug](https://www.reddit.com/r/webdev/comments/57ii4f/psa_edge_14_ships_a_broken_windowfetch/?st=iuoar07b&sh=15ed03f9) is already fixed in 14 (modulo the DELETE case).

Sending Edge <=14 the polyfill is not unreasonable, but we're also discussing possibly servicing the DELETE fix in a pre-15 update, and it's already kind of an edge case IMO. The headers issue is more serious, if folks are already depending on it. Up to you, but just trying to provide backstory! :smiley:

triblondon:
OK, well, all we're doing here is removing an unnecessary built in detect, which transfers the decision on whether to feature detect back to the developer, so I'm happy.

triblondon (approved):

PR #954, opened 2016-10-25
Update test browsers

triblondon:
Use latest available chrome & firefox, add android 5.1


falsandtru:
Thanks for updating! Seems like we can specify `latest`, `latest-1`, and `beta`. Can you use this style?

https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options
https://saucelabs.com/platforms

JakeChampion (approved):

triblondon:
@falsandtru I'm not sure about that.  It might cause passing tests to start failing without any changes on our end, which would start to block releases.  I see the other side too so if @JakeChampion feels strongly about it maybe we can have a chat internally.

Also you might be interested in http://frontend.connpass.com/event/42323/ which we're running next week with the TAG at the Nikkei building. 75% of the talks are in Japanese, and there's translation for the English one.  I think polyfills will come up.

triblondon:
OK, CI is working, but a full compatgen is failing.  I need to do some more work here before merging.

falsandtru:
I see. As a note, TravisCI's `allow_failures` option is helpful in this case, but CircleCI doesn't seem to have that option.

JakeChampion (dismissed):

JakeChampion:
It seems that saucelabs have a specific version of edge 14 which breaks on one of our defineproperty tests. This test passes on my local version and the version on browserstack.

JakeChampion:
Running a full grunt compatgen now with latest code from master merged into this branch.

JakeChampion (approved):

PR #962, opened 2016-10-27
Fix semver for IE/Edge for NodeList.prototype.@@iterator

### LINKED ISSUE #961: Edge does not yet support NodeList.prototype.@@iterator
As of Edge 14 there is still no native support for `NodeList.prototype[Symbol.iterator]`. The semver in the config for that polyfill is `9 - 12`.

tuespetre:
Fixes #961


JakeChampion:
Can confirm it doesn't work on Edge 14. Thanks for fixing this :-)

JakeChampion (approved):

PR #964, opened 2016-10-29
Add Event and performance.now as deps to intersectionobserver

JakeChampion:
performance.now is used on line 525 -- https://github.com/WICG/IntersectionObserver/blob/gh-pages/polyfill/intersection-observer.js#L525

addEventListener is on line 559 -- https://github.com/WICG/IntersectionObserver/blob/gh-pages/polyfill/intersection-observer.js#L559


seanmtracey:
👍

seanmtracey (approved):
👍

PR #965, opened 2016-10-30
Add deprecation note to String.prototype.contains as for Array.prototype.contains

dbaron:
This adds the same note as in Array.prototype.contains, except with the
"Use ... instead" part renamed from Array to String.

Note that this similarity maybe isn't _exactly_ the right thing, since
the link in the note is specifically about Array.prototype.contains
breakage.  That said, there was broken content from
String.prototype.contains documented in
https://bugzilla.mozilla.org/show_bug.cgi?id=789036 , but TC39 didn't
rename the two methods until the later set of broken content from
Array.prototype.contains.  I'm somewhat hesitant to link to bugzilla
from here, however.

I haven't run any tests on this change.  (I'm presuming that it will
make the little (i) icon show up in
https://polyfill.io/v2/docs/features/ , but I haven't tested this.)


dbaron:
And I agree to the contribution terms in https://cdn.polyfill.io/v2/docs/contributing .

JakeChampion:
It might be useful to add the bug ticket for `String.prototype.contains` -- https://bugzilla.mozilla.org/show_bug.cgi?id=789036

triblondon (approved):

PR #970, opened 2016-10-30
Update error messages for Object.defineProperty

JakeChampion:
I noticed our error messages were a bit out dated and that these tests were failing in Canary.


triblondon (approved):
There's something to be said for making it explicit that the error is being thrown by the polyfill, but obviously there's no consistency to that.  I'm happy.

PR #972, opened 2016-10-31
Firefox shipped IO in 52

triblondon:
At TAG meeting, dbaron noted that IO will likely ship in FF52. If it does this update can go through.


triblondon:
Confirmed on MDN
https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Browser_compatibility

JakeChampion (approved):

PR #973, opened 2016-10-31
Add polyfill for HTML5's `document.currentScript`

### LINKED ISSUE #952: Support for HTML5 `document.currentScript`
Can add guaranteed support for HTML's [`document.currentScript`](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) in IE `8-10` (but _**NOT**_ IE 11 😭) by using the techniques from [JamesMGreene/document.currentScript](https://github.com/JamesMGreene/document.currentScript).

If it weren't for the lack of defining getters _**on DOM interfaces**_ in IE `<8`, this can actually be supported all the way down to IE `6`.

I can submit a PR for this soon-ish.

JamesMGreene:
Fixes #952

### DIFF
--- a/polyfills/document.currentScript/config.json
+++ b/polyfills/document.currentScript/config.json
@@ -1,0 +1,12 @@
+{
+	"browsers": {
+		"ie": "8 - 10"
+	},
+	"repo": "https://github.com/JamesMGreene/document.currentScript",
+	"license": "MIT",
+	"docs": "https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript",
+	"spec": "https://html.spec.whatwg.org/multipage/dom.html#dom-document-currentscript",
+	"notes": [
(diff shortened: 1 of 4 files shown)


triblondon:
I happen to be in a TAG meeting with three browser vendors and the google and msft folks we're pretty happy with this.

JamesMGreene:
Funny. :+1:

JamesMGreene:
P.S. If the Microsoft folks would ever consider [releasing a bug fix for IE11](https://github.com/JamesMGreene/document.currentScript#public-service-announcement-psa) that either enables `document.currentScript` or brings back `HTMLScriptElement#readyState`, the world would appreciate it.  :wink:

But I'm guessing that's a no-go since [they passed on doing so in the past and have moved on full-steam ahead to Edge. Understandable but unfortunate. 😕

JakeChampion (approved):

PR #983, opened 2016-11-07
Adding dataset polyfill

### LINKED ISSUE #188: Add dataset polyfill
Requested in https://github.com/jonathantneal/polyfill/issues/48.  Also desired by FT.

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-service/issues/188


polyfills/Element/prototype/dataset/detect.js
```diff
@@ -0,0 +1,6 @@
+(document.documentElement.dataset &&
+	(
+		Object.getOwnPropertyDescriptor(Element.prototype, 'dataset')  &&
+		Object.getOwnPropertyDescriptor(Element.prototype, 'dataset').get
```
triblondon: This is quite a new API to use for a detect.  Is it possible to detect without using ES5?

triblondon (approved):

polyfills/Element/prototype/dataset/detect.js
JakeChampion: We could use `__getter__` if `getOwnPropertyDescriptor` isn't there.
I've included `getOwnPropertyDescriptor` in the dependencies which should mean that it is present for the detect.

triblondon: hmm, I don't know that that's universally true, eg. for RUM beacon code... https://polyfill.io/v2/polyfill.js?rum=1&ua=chrome/50 (see top section).  We will probably want to close that loophole (and figure out if there are any others), or ensure that detects are always ES3 compatible.

One thing that could be done in the RUM case is putting each detect in a try...catch, but that will bloat the code a fair bit.

JakeChampion: According to MDN every browser which has dataset also has `Object.getOwnPropertyDescriptor`.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#Browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor#Browser_compatibility

Shall I update the detect to fallback to `__getter__` instead?

triblondon: Sorry for being slow... what does that correlation tell us?  I'm concerned about browsers that have neither getOwnPropertyDescriptor nor dataset.  The detect for dataset will error in browsers that don't have getOwnPropertyDescriptor, unless a polyfill for getOwnPropertyDescriptor is loaded first.  It will be most of the time, but I am concerned that there are edge cases, eg. RUM, where it is not.

JakeChampion: Agree. The only solution I can think of is to use `getOwnPropertyDescriptor` and fallback to `__getter__` if it doesn't exist. Do you think that will be ok?

triblondon: Fine

JakeChampion:
@triblondon updated the detect :-)

polyfills/Element/prototype/dataset/detect.js
```diff
@@ -0,0 +1,8 @@
+(function(){
+	if (!document.documentElement.dataset) {
+		return false;
+	}
+	var el = document.createElement('div');
+	el.setAttribute("data-a-b", "c");
+	return el.dataset && el.dataset.aB == "c";
```
JakeChampion, line 7: This detect should work in all browser environments.

PR #986, opened 2016-11-15
Use mississippi to handle error cases

JakeChampion:
This PR doesn't resolve the issue around why the destination stream has closed or error'd but it does ensure we don't throw an exception.

When using standard `source.pipe(destination)` the source will not be destroyed if the destination emits `close` or `error`. You are also not able to provide a callback to tell when the pipe has finished.

`mississippi.pipe` does these two things for you, ensuring you handle stream errors 100% of the time (unhandled errors are probably the most common bug in most node streams code).

### DIFF
--- a/service/routes/api.js
+++ b/service/routes/api.js
@@ -6,6 +6,7 @@ const polyfillio = require('../../lib');
 const PolyfillSet = require('../PolyfillSet');
 const metrics = require('../metrics');
 const express = require('express');
+const miss = require('mississippi');
 
 const router = express.Router();  // eslint-disable-line new-cap
 const contentTypes = {".js": 'application/javascript', ".css": 'text/css'};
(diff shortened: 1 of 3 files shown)


service/routes/api.js
```diff
@@ -72,12 +73,20 @@ router.get(/^\/v2\/polyfill(\.\w+)(\.\w+)?/, (req, res) => {
...
-			res.write("\ntypeof "+req.query.callback+"==='function' && "+req.query.callback+"();");
+	const callbackWrapper = miss.through(function () {},
+		function (cb) {
+			if (req.query.callback && req.query.callback.match(/^[\w\.]+$/)) {
+				cb(null, "\ntypeof "+req.query.callback+"==='function' && "+req.query.callback+"();");
+			} else {
+				cb();
+			}
+		}
+	);
```
triblondon, line 84: Do we have to use transform streams in a pipe like this?  It seems to be complicating the logic to push the whole contents of the output through the jsonP-callback-adding stream, then pass the whole lot on again to the response stream.  Previously we were listening for the end of stream event and then writing to the output stream, without seeing any of the stream content, which strikes me as a slightly better pattern, but I don't know a lot about streams.

service/routes/api.js
```diff
@@ -72,12 +73,20 @@ router.get(/^\/v2\/polyfill(\.\w+)(\.\w+)?/, (req, res) => {
...
+
+	miss.pipe(outputStream, callbackWrapper, res, (err) => {
```
triblondon, line 86: Consider renaming callbackWrapper to jsonpWrapper to be more descriptive?

PR #987, opened 2016-11-15
Swap out custom stream lib for external libs

JakeChampion:
This reduces the amount of utility code in our codebase and as a bonus, each external lib has tests which cover their features.


lib/index.js
```diff
@@ -119,139 +122,138 @@ function getPolyfillString(options) {
...
+		if ('all' in options.features) {
+			output.add(streamFromString("\nconsole.log('"+allWarnText+"');\n"));
+		}
+	});
 
-	return options.stream ? output : new Promise(resolve => {
-		const bufs = [];
-		output.on('data', chunk => bufs.push(chunk));
-		output.on('end', () => resolve(Buffer.concat(bufs).toString('UTF-8')));
+	return options.stream ? output : streamToPromise(output).then(bufs => {
```
triblondon: You could use a higher level lib here like stream-to-string: https://www.npmjs.com/package/stream-to-string

service/routes/api.js
```diff
@@ -71,13 +74,18 @@ router.get(/^\/v2\/polyfill(\.\w+)(\.\w+)?/, (req, res) => {
...
-	outputStream.on('end', () => {
-		if (req.query.callback && req.query.callback.match(/^[\w\.]+$/)) {
-			res.write("\ntypeof "+req.query.callback+"==='function' && "+req.query.callback+"();");
+	const outputStream = mergeStream();
+
+	outputStream.add(polyfillio.getPolyfillString(params));
+
+	if (req.query.callback && req.query.callback.match(/^[\w\.]+$/)) {
+		outputStream.add(streamFromString("\ntypeof "+req.query.callback+"==='function' && "+req.query.callback+"();"));
+	}
```
triblondon, line 83: I much prefer this solution to the one in the other PR - the source order still matches the order of output which makes it easier to grok.

test/node/lib/test_index.js
```diff
@@ -138,15 +138,13 @@ describe("polyfillio", () => {
...
 		it('should support streaming output', done => {
-			const ReadableStream = require('stream').Readable;
 			const buf = [];
 			const s = polyfillio.getPolyfillString({
 				features: { default: {} },
 				uaString: 'chrome/30',
 				stream: true,
 				minify: false
 			});
-			assert.instanceOf(s, ReadableStream);
```
triblondon, line 149: Shouldn't we at least assert that it implements ReadableStream?

triblondon (approved):
I'm sad to lose my stream module, but this makes sense and at least I know a lot more about streams now!

JakeChampion:
Thanks for the review @triblondon, useful feedback. I've made the changes you suggested and will merge this one CI passes.

PR #992, opened 2016-11-18
Nicer sauce

triblondon:
Thousands of lines of output made our sauce task hard to follow.  It now looks like this:

image

The clock icon indicates that we have not received data from that browser for that amount of time


triblondon:
Some issues remain to resolve:
- [x] Some browsers regularly timeout before we start receiving results, notably Android
- [ ] Event loop blocking notifications get in the way of the console rewriting, so I disabled in all except prod, which is not ideal
- [x] Yet to fully test what happens when there are failures to report
- [x] Progress meter output fails pretty spectacularly when you try and do a compatgen because the tests don't fit on the console window

triblondon:
Android: turns out this browser typically takes ~45 seconds to start up.  My timeout was simply too short.

tasks/grunt/saucelabs.js
```diff
@@ -10,8 +10,8 @@ const testResultsFile = path.join(testResultsPath, 'results.json');
 const SauceTunnel = require('sauce-tunnel');
 const mkdirp = denodeify(require('mkdirp'));
 const cli = require('cli-color');
-const pollTick = 500;
-const testBrowserTimeout = 40000;
+const pollTick = 1000;
+const testBrowserTimeout = 60000;
```
JakeChampion, line 14: 👍

Gruntfile.js
```diff
@@ -24,18 +24,25 @@ module.exports = function(grunt) {
 			}
 		},
 		"saucelabs": {
+			debug: {
+				options: {
+					urls: { all: 'http://127.0.0.1:3000/test/director?mode=all' },
```
JakeChampion: I think this should grab port from the process.env if it exists as that is what the service will use.

Gruntfile.js
```diff
@@ -24,18 +24,25 @@ module.exports = function(grunt) {
...
+					browsers: browsers.quick,
+					concurrency: 3
+				}
+			},
 			compat: {
 				options: {
 					urls: {
 						all: 'http://127.0.0.1:3000/test/director?mode=all',
```
JakeChampion: same here, not sure why we didn't do this before. Needs to use process.env.PORT if it is set.

JakeChampion:
From CircleCI:
```
 • android/5.1 (targeted): Error: Timed out at 'waiting for results'
```

triblondon:
Dealt with Jake's feedback and my earlier issues.

JakeChampion (approved):

PR #993, opened 2016-11-18
es6 compliant Object.keys polyfill

gnoff:
This PR updates tests to reflect expected es2015 behavior of Object.keys (for example how latest Chrome treats this method)

Note that because IE9+ has es5 compliant Object.keys support it is not getting this polyfill and it is not running these tests at all. I don't know how you intend to make this polyfill available to browsers that support an older API for a function so I did not change which browsers are targeted in tests yet.

* adds new tests for behavior of Functions as well as empty, null, undefined arguments
* adds new failing tests for Arrays, numbers, strings

### DIFF
--- a/polyfills/Object/keys/tests.js
+++ b/polyfills/Object/keys/tests.js
@@ -37,3 +37,43 @@ it('works with objects containing otherwise non-enumerable keys', function () {
 		valueOf: 0
 	}).length, 7);
 });
+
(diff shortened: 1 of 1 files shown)


triblondon:
This is great, thanks - now we just need the solution!

JakeChampion:
This has been added in #1033

PR #1000, opened 2016-11-27
A bit less grunt

triblondon:
Decided to bite the bullet and kill grunt.

WIP; so far I've done a few of the key tasks.  I also took the opportunity to refactor service.js and compattable.js to be promise based.

A few issues to address:

* Options passed from gruntfile to task: Makes sense to use minimist for these (eg. service task), but others are unnecessary
* Logging is less easily coloured and controlled, but we can get more flexibility, eg as we're already doing in the saucelabs task.  It might be worth creating a few useful higher level abstractions to replace things like grunt.verbose.  We still have the issue of differentiating the stdout of the child process from the output of the service task (and subsequent tasks like sauce), which is not handled well under grunt anyway, and maybe this is an opportunity to improve it.
* Most grunt contrib modules can be trivially replaced by shell commands, eg `simplemocha` => `mocha test/node/**/*.js`.  I haven't thought about `watch` yet but maybe nodemon?


triblondon:
Going to pause this to see if @JakeChampion thinks it's worth finishing / is going in the right direction.

JakeChampion:
Seems good, what's the reasoning behind this? If this progresses could we test to see if it still works on Windows OS for development.
Would you have a programmatic way of displaying what tasks are available?

triblondon:
Reasoning:

* No other Origami services use grunt (do they?) so it creates a learning curve
* Grunt isn't really very popular anymore so it might limit community participation
* It adds dependencies. I don't like unnecessary dependencies :-)
* Generally the non-grunt versions of tasks are actually shorter than the grunt version

Displaying tasks: yes, run `npm run`

JakeChampion:
This all sounds great. Origami services have started using `Make` and `Makefiles`, not sure if this project should move in that direction or not, happy either way.

JakeChampion (approved):

triblondon:
Oh, I thought you'd want to discuss this!  But great.  Do the tasks work for you?

PR #1002, opened 2016-11-28
Add Intl polyfill support for QupZilla

lu40:
Add Intl polyfill support for QupZilla versions older than 2.0

As the browser isn't very widespread there's no specific info at MDN, caniuse and Kangax or anywhere else. However I found about the needed Intl polyfill when checking a user-submited issue concering our app (https://github.com/Openki/Openki/issues/596).

This is the infinite loading screen I get when using any version of QupZilla earlier than 2.0 (here 1.8.9-1):

openki_net_qupzilla_1 8 9

And this is how the same page looks starting from 2.0 (here 2.0.2-1):

openki_net_qupzilla_2 0 2

Hereby I agree to the [contribution terms](https://polyfill.io/v2/docs/contributing#contribution-terms).

### DIFF
--- a/polyfills/Intl/config.json
+++ b/polyfills/Intl/config.json
@@ -9,7 +9,8 @@
 		"safari": "<10",
 		"ios_saf": "*",
 		"firefox_mob": "*",
-		"samsung_mob": "*"
+		"samsung_mob": "*",
+		"qupzilla": "<2"
 	},
 	"dependencies": [],
 	"build": {


JakeChampion:
@lu40 This PR will not resolve your issue as we don't have any detection in place for qupzilla.
I don't think we would accept a PR to add qupzilla specific detection. If qupzilla is based on another browser such as Chromium etc then we would accept a PR which aliases qupzilla to a browser we do support. https://polyfill.io/v2/docs/contributing/common-scenarios#new-browser

Another solution to your issue would be to set the polyfill-service to polyfill unknown browsers, this can be done by adding `&unknown=polyfill` to your polyfill.io url

lu40:
@JakeChampion Ok, thanks however for your response! I kind of thought that would have been too easy.. :) As I'm not using QupZilla myself and just tried a quick fix I'm going to close this PR.

PR #1003, opened 2016-11-29
Feature removal: No longer accept port as a flag

JakeChampion:
I couldn't see this documented anywhere or being used anywhere

### DIFF
--- a/bin/polyfill-service
+++ b/bin/polyfill-service
@@ -3,13 +3,10 @@
 
 require('dotenv').config({silent: true});
 
-const minimist = require('minimist');
-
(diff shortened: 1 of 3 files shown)


triblondon (approved):
I'm using minimist in node tasks in the `less-grunt` branch, so it'll conflict if you remove that, might be easier to leave the dep in there.  No objection to removing the CLI flag on the bin file

JakeChampion:
@triblondon I removed it from the `dependencies` section, if you add it to `devDependencies` in your `less-grunt` branch as it is only used for development tasks then I think there won't be a conflict.

PR #1004, opened 2016-11-29
Enable `performance.now` and `Array.from` for Opera Mini

kdzwinel:
Two fixes:

- Opera Mini does not support `performance.now` at all: http://caniuse.com/#feat=high-resolution-time . Note that `requestAnimationFrame` depends on `performance.now` causing raf polyfill to fail on that browser.

- MDN page (and our error tracking solution) indicates that Opera Mini does not support `Array.from`: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from


JakeChampion (approved):

JakeChampion:
Thanks! @kdzwinel

kdzwinel:
Thank you! There are some more polyfills missing for Opera Mini (Map, Set, includes) and I'll create a PR as soon as I'll find time to double check them.

What's the release cycle of polyfill.io? I'd love to know when this fix hits production.

JakeChampion:
I've been ramping up the frequency of release cycles. I expect to cut another release for qa tomorrow, it will go live to production a minimum of 7 days after that.

PR #1007, opened 2016-12-05
Adding feature filter

adamduncan:
Incorporating simple feature filter input component. Addition of `initFeatureFilter` method to ui.js.

Using o-forms 2.0.0, but could possibly do with some design love?

polyfillio-filter


docs/assets/js/ui.js
```diff
@@ -160,9 +160,40 @@ function initNotesToggles() {
 	});
 }
 
+function initFeatureFilter() {
+	var filterInput = document.getElementById('filter-features');
+	var featuresRows;
+
+	if (filterInput) {
+		featuresRows = document.querySelectorAll('[data-feature-name]');
```
triblondon: `tr.feature` is probably more performant?

triblondon: Also I'd consider building a simple object map to allow faster lookups, so you store this:

```
{
   "array.map": [element],
   "fetch": [element],
   ...
}
```

docs/assets/js/ui.js
```diff
@@ -160,9 +160,40 @@ function initNotesToggles() {
...
+		filterInput.addEventListener('keyup', filterFeatures);
```
triblondon: Consider `paste` here as well, for better accessibility?

triblondon (commented):
This looks really nice.  From a design perspective, I'd consider whether you could use a search icon on the left in the same row as the legend/key, and use that to toggle between the legend and the search text box.  That would reduce the amount of vertical space used by the new elements.

Not essential.  @JakeChampion any thoughts?

adamduncan:
Thanks a lot for all the feedback @triblondon. Good shout on the perf/a11y tweaks! I'll get on them and update the PR.

adamduncan:
@triblondon, committed the suggested improvements in 9b4cf6c. A little bit of cross-browser `clipboardData` voodoo required, however the cached lookups were a great shout. Thanks.

Re: search icon/toggle, I wonder whether putting the filter feature behind a click detracts from its immediate usefulness? Definitely agree it needs tightening up from a layout perspective, so happy to take your lead on it.

JakeChampion:
I think this looks great :-)

My opinion on the search bar is to have it visible on page load, I think that if it was hidden on page load then it could be easily missed.

JakeChampion (approved):

triblondon (approved):
This looks fine to me, happy to ship.

adamduncan:
@JakeChampion @triblondon Thanks guys 👍

JakeChampion:
@adamduncan Thanks for the PR! 

Did you know we offer one month's free premium access to all FT content (normally $50) to anyone who has a pull request merged in this repository? If you want to claim it, there's [one Google form](https://docs.google.com/a/ft.com/forms/d/1aykPiYBwqKpKY6nvGVi43wOU_5epr3XLgE00Vi52SMQ/viewform) to fill out :-)

PR #1009, opened 2016-12-07
removing files created at build time and tidying up ignores

JakeChampion:


.eslintignore
```diff
@@ -1,19 +1,21 @@
...
+polyfills/Intl/~locale/**/polyfill.js
+polyfills/JSON/polyfill.js
+polyfills/navigator/sendBeacon/polyfill.js
+polyfills/MutationObserver/polyfill.js
+polyfills/Promise/polyfill.js
+polyfills/setImmediate/polyfill.js
+polyfills/URL/polyfill.js
+polyfills/UserTiming/polyfill.js
+polyfills/WeakSet/polyfill.js
+polyfills/~html5-elements/polyfill.js
```
triblondon, line 26: This reverses a change I made recently. Sorry if I didn't understand what I was doing. Can you explain why we need to specifically exclude the polyfill.js files rather than the whole generated folder? Is it because you want to lint detects and tests for imported polyfills?

rowanmanning (approved):

.eslintignore
```diff
@@ -1,19 +1,27 @@
...
-polyfills/UserTiming
-polyfills/WeakSet
-polyfills/~html5-elements
+# We ignore the polyfill.js files for third-party polyfills
+# because we do not control their implementation.
+
+# We do not ignore the config.json, detect.js or tests.js
+# because we do control their implementation.
+
+polyfills/_ArrayIterator/polyfill.js
```
AdriVanHoudt, line 7: can't you do `polyfills/**/**/polyfill.js` or `polyfills/**/polyfill.js` or something like that?

JakeChampion: That would ignore all polyfills which we implement ourselves in the repository

JakeChampion: That's okay, thanks for taking the time to review :-)

PR #1027, opened 2016-12-29
Add bespoke detects for each Intl locale

JakeChampion:
Detects for locales are currently:
```js
'Intl' in this
```

I propose we switch to this:
```js
'Intl' in this && Intl.Collator.supportedLocalesOf(locale).length === 1 &&
Intl.DateTimeFormat.supportedLocalesOf(locale).length === 1 &&
Intl.NumberFormat.supportedLocalesOf(locale).length === 1
```

where `locale` is the locale we are aiming at detecting support for. This is similar to the approach [yahoo/intl-locales-supported](https://github.com/yahoo/intl-locales-supported/blob/master/index.js) has taken.

### DIFF
--- a/polyfills/Intl/update.task.js
+++ b/polyfills/Intl/update.task.js
@@ -77,7 +77,6 @@ locales.forEach(function (file) {
 	const polyfillOutputPath = path.join(localeOutputPath, 'polyfill.js');
 	const detectOutputPath = path.join(localeOutputPath, 'detect.js');
 	const configOutputPath = path.join(localeOutputPath, 'config.json');
-// locale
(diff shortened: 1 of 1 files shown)


JakeChampion:
@caridy, could I have your opinion on this feature detect code for Intl locales please?

rowanmanning (approved):

caridy:
I'm not sure about this change. Mostly, that code in the docs, and what we created at Yahoo for detection is intended for Node, since some versions of node comes with EN only, or a subset of locales. But I doubt that this applies to the client side. At the moment, client side is all-or-nothing since the polyfill doesn't distinguish between locale supported, it only distinguish between Intl exists or not.

PR #1046, opened 2017-01-10
Enable tests to run on browserstack and saucelabs

JakeChampion:
I've added browserstack as the default and kept the ability to run saucelabs in case we want to revert back.

There seems to be an issue in the android 4.4 emulator on saucelabs where the intersection oberserver tests are failing on a test because of a 1 pixel difference between the expected result and the actual result. These tests pass on the android 4.4 real devices used on browserstack.


rowanmanning (approved):
There are a few whitespace inconsistencies (mostly in the browserstack task), but nothing that should prevent this from being merged. Nice work 👍

tasks/node/saucelabs.js
```diff
@@ -1,366 +1,224 @@
 'use strict';
-
```
JakeChampion, line 2: The original file was renamed to [remotetest.js](https://github.com/Financial-Times/polyfill-service/pull/1046/files#diff-2b56ca1e2aba3b8dc7f088ac145770d2). 

This file now contains the configuration for running the tests on SauceLabs

tasks/node/browserstack.js
```diff
@@ -0,0 +1,218 @@
...
+if (!process.env.BROWSERSTACK_USERNAME || !process.env.BROWSERSTACK_ACCESS_KEY) {
+	throw new Error('BROWSERSTACK_USERNAME and BROWSERSTACK_ACCESS_KEY must be set in the environment to run tests on BrowserStack.  For more information about how to set this up or for alternative methods of testing, see https://polyfill.io/v2/docs/contributing/testing');
+}
+
+module.exports = {
+	creds: {
+		username: process.env.BROWSERSTACK_USERNAME,
+		key: process.env.BROWSERSTACK_ACCESS_KEY
+	},
+	tunnel: function () {
```
JakeChampion, line 14: Each provider has a different way to configure their tunnel, in order to simplify the `remotetest.js` code I've decided to have a standard interface for opening and closing tunnels. It is a basic object which has two methods which return Promises: `openTunnel`, `closeTunnel`.

tasks/node/remotetest.js
```diff
@@ -0,0 +1,340 @@
...
+		'ie/13',
+		'ie/11',
+		'ie/10',
+		'ie/9',
+		'ie/8',
+		'ie/7',
+		'safari/9',
+		'safari/8',
+		'safari/5.1',
+		'android/4.4',
```
JakeChampion, line 70: android 5.1 and ios/9.2 have been removed as browserstack have not added these devices to their automated testing suite yet. I spoke with them on Twitter and this was their response - https://twitter.com/browserstack/status/818835556968833024

tasks/node/remotetest.js
```diff
@@ -0,0 +1,340 @@
...
+const readFile = denodeify(require('graceful-fs').readFile);
+const writeFile = denodeify(require('graceful-fs').writeFile);
+const testResultsPath = path.join(__dirname, '../../test/results');
+const testResultsFile = path.join(testResultsPath, 'results.json');
+const mkdirp = denodeify(require('mkdirp'));
+const argv = require('minimist')(process.argv.slice(2));
+const cli = require('cli-color');
+const pollTick = 1000;
+const testBrowserTimeout = 60000;
+const environment = ['browserstack', 'saucelabs'].filter(x => x in argv)[0] || 'browserstack';
```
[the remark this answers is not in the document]
triblondon: Also is it needed at all or could you simplify to 

```
require('./' + ['browserstack', 'saucelabs'].filter(x => x in argv)[0] || 'browserstack');
```

And it might be slightly less obscure (and more forwards compatible) to name that CLI arg:

```
require('./' + (['browserstack', 'saucelabs'].contains(argv.provider) && argv.provider) || 'browserstack');
```

triblondon (approved):
Marvellous!  Can I have access to the browserstack dashboard?

PR #1047, opened 2017-01-11
use URI Template in API doc

tpluscode:
I find the API documentation unclear and took me a while to understand.

What do you think about using a standard URL template (RFC6570)? I didn't include the `<var></var>` tags. Please edit as you see fit. I just wasn't sure what to outline.

Given such change you could get rid of the `WHERE` column completely. Up to you

### DIFF
--- a/docs/api.html
+++ b/docs/api.html
@@ -16,7 +16,7 @@
 				<li>There are changes to the Node API.  See the README for details.</li>
 			</ul>
 
-			<h2>GET /v2/polyfill<var>:minify</var>.<var>:type</var></h2>
+			<h2>GET /v2/polyfill{.minify}.{type}{?features,excludes,flags,ua,callback,unknown,rum}</h2>
 
 			<p>Fetch a polyfill bundle.</p>
 


triblondon:
Never heard of this RFC, but it sounds like a good idea if it's a common standard.  Do you know of other popular API docs that use this format?

tpluscode:
GitHub API is probably the most prominent example. See the example response in their docs: https://developer.github.com/v3/repos/#response

Also, anyone using [HAL](http://stateless.co/hal_specification.html) is using that format. Although it's not explicitly referenced in the HTML doc, they did mention it in their [RFC draft](https://tools.ietf.org/html/draft-kelly-json-hal-08#section-5.1)

tpluscode:
Ah, there's the documentation bit I was looking for: https://developer.github.com/v3/#hypermedia

triblondon:
Hmm, but their method documentation actually uses basically the same convention that we already use:

image

tpluscode:
True, which kind of surprises me. Such syntax is only good for concrete segments so maybe it's good enough for them. 

In your case URI Template allows

1. being explicit about optional dot-prefixed extension (`{.min}`)
1. explicitly listing query string params (`{?a,b,c}`)

I find this cleaner than having to look at both the template and the table to determine where each option belongs to.

triblondon:
Jake, you're keen on this?  I am mildly against because it means writing the params in two places and they are likely to get out of date.  I don't hear anyone else saying the docs are unclear and I don't see any popular examples of services documenting their APIs like this...  I do appreciate the suggestion but I'm not totally sold tbh.

JakeChampion:
I'm very happy to hear comments about our documentation, thanks for taking the time to give us feedback. I agree with the concern @triblondon has made about it potentially becoming out of date documentation as the parameters would exist in two places.

triblondon:
OK, let's stick with the format we have.  @tpluscode I'm sorry to decline your suggestion and thanks for making it, we do appreciate the input and hope you feel we gave it a fair hearing!

Issue #990, opened 2016-11-17
Facebook in-app browser (on both iOS and Android) not working


andyburke:
We are seeing issues with the polyfill not appropriately filling inside the Facebook in-app browser.

This seems like it may be a regression on #561

Reproduction:

https://m.facebook.com/story.php?story_fbid=10154490930306508&id=739196507

```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <title>FB Browser Window setImmediate repro</title>

    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=default,setImmediate"></script>
</head>

<body>
    <script type="text/javascript">
        setImmediate( function() {
            alert( 'setImmediate worked' );
        } );
    </script>
<body>
```

jacklee0223:
Same thing happens to Object.assign

JakeChampion:
@andyburke @jacklee0223 do you happen to know of a way to do automated testing against facebook's in-app browser? When we fix this I want to ensure we are unlikely to regress again in the future.

andyburke:
@JakeChampion I'm not aware of a way, other than theoretically faking their UA?

Is there any progress on this? Do you have any idea what could be causing it? We're willing to contribute, but would probably need at least a little guidance on how to proceed.

triblondon:
@andyburke I suggest you first find out how we are classifying that browser by sending the browser directly to the polyfill URL (without `.min`), or if FB refuses to directly load and display a script file, by downloading the polyfill.io resource using `fetch` and adding the raw source to the document body, eg something like:

```
<script>
fetch("https://polyfill.io/v2/polyfill.js")
  .then(resp => resp.text())
  .then(text => document.body.innerText = text)
;
</script>
```

It seems likely that we are not recognising the browser, which might be a change we've made to our UA module, but more likely either a change in the useragent module we depend on, or in the actual string Facebook uses as their UA header.

andyburke:
@triblondon I've updated the test url: https://hellofloat.com/fb_browser_polyfill_test/

It results in this (tested on an Android device in FB app):

![screenshot_2017-01-12-20-21-23](https://cloud.githubusercontent.com/assets/273857/21918322/06eac500-d905-11e6-9ffa-d372415eee88.png)

triblondon:
[...]

PR #1050, opened 2017-01-12
Add pre-parse for Vivaldi, fixes #735

### LINKED ISSUE #735: Add support for Vivaldi
I am guessing that all this needs is a user agent alias.

triblondon:


lib/UA.js
```diff
@@ -129,7 +120,17 @@ function UA(uaString) {
...
-		this.ua = useragent.parse(stripiOSWebViewBrowsers(uaString));
+
+		// Chrome and Opera on iOS uses a UIWebView of the underlying platform to render
+		// content, by stripping the CriOS or OPiOS strings the useragent parser will alias the
+		// user agent to ios_saf for the UIWebView, which is closer to the actual renderer
+		uaString = uaString.replace(/((CriOS|OPiOS)\/(\d+)\.(\d+)\.(\d+)\.(\d+)|(FxiOS\/(\d+)\.(\d+)))/, '');
+
+		// Vivaldi browser is recognised by UA module but is actually identical to Chrome, so
+		// the best way to get accurate targeting is to remove the vivaldi token from the UA
+		uaString = uaString.replace(/ vivaldi\/[\d\.]+\d+/i, '');
```
JakeChampion, line 133: Is this possible to do using our function based aliases? https://github.com/Financial-Times/polyfill-service/blob/master/lib/UA.js#L86

triblondon: Good point.  I'd forgotten we could use functions for aliasing.  And it looks like we're already using this for Facebook for iOS, but there's an open issue for Facebook for Android: #990. I'm not really convinced that the function approach works, because it requires that we act on the initial parsed results of the UA string - and in the case of facebook for iOS, that's sort of OK because we can use what the ua module reports as the OS version, as if it's ios_saf.  That's a bit weird though.

Ideally, you want to say 'if this UA is identified as facebook, strip the bit that identifies it as facebook and re-parse it', or even better, turn off the bit of the ua module that matches facebook, in the hope that it would then match chrome or safari instead.

Failing that though, the most reliable option seems to be to strip the Facebook token before we parse it.

triblondon:
@JakeChampion I'd like to go with this solution.  Could you review again and consider my comment in response to your earlier review?  If you're happy, I will change the Facebook approach in the branch for #990 after merging in this PR.

JakeChampion (approved):

JakeChampion:
@triblondon Should we remove the function based aliasing approach?

PR #1055, opened 2017-01-12
Polyfill targeting updates

JakeChampion:
These changes were made based on the data from https://polyfill.io/v2/getRumCompatData


polyfills/Event/hashchange/config.json
```diff
@@ -11,7 +11,10 @@
 	"browsers": {
 		"ie": "6 - 7",
 		"safari": "4",
-		"samsung_mob": "*"
+		"samsung_mob": "*",
+		"chrome": "<52.0.0",
+		"firefox": "<50.0.0",
+		"ios_saf": "< 9.3.0"
```
triblondon: RUM suggests Chrome 54 doesn't have it:

```
"Event.hashchange","chrome","54.0.0","0","128","8","234","No","False negative - add targeting"
```

triblondon: Same for the other hashchange records in RUM - they are false negs, not false positives.

polyfills/Symbol/config.json
```diff
@@ -8,7 +8,7 @@
 		"safari": "<9",
 		"chrome": "< 50",
 		"firefox": "< 40",
-		"ios_saf": "<9",
+		"ios_saf": "<10",
```
triblondon, line 11: This is weird because the RUM suggests it was falsely served to iOS 10 before:

```
"Symbol","ios_saf","10.0.0","100","280","4","493","Yes","False positive - remove targeting"
```

JakeChampion: `<10` should mean that ios 10 does not get the symbol polyfill but ios 9 does

triblondon: Yep, but I don't understand why we were serving it to 10 before. Wonder if this was just people using the `always` flag

JakeChampion: Seems we are serving it to ios saf 10 -- https://polyfill.io/v2/polyfill.js?features=Symbol&ua=ios_saf/10&rum=1

JakeChampion: In production the config is currently set to `*`, that is why ios saf 10 is getting symbol :-)

polyfills/Event/config.json
```diff
@@ -9,7 +9,7 @@
 	"browsers": {
 		"firefox": "6 - 10",
 		"ie": "6 - 11",
-		"ie_mob": "10",
+		"ie_mob": "11",
```
hannesj: should this be <=11

polyfills/Symbol/config.json
triblondon: Ah, OK, that makes sense then

polyfills/Event/config.json
triblondon: If anything, I'd say `<12` because we prefer `<` syntax to `<=`.  I don't see a record of `Event` in `ie_mob` in the compat data, maybe it's fallen out of the 30-day window.

JakeChampion:
@triblondon Could you give this another review?

triblondon (approved):

triblondon:
@JakeChampion any reason not to merge this?

PR #1056, opened 2017-01-12
Add Mutation observer

JakeChampion:


triblondon (approved):
Looks like this polyfill has great test coverage already, so not including tests here is OK.  The licence is not OSI approved and has potential issues (an OSI approved licence like MIT would be better), but I think the risk is low, and in any case we no longer actually distribute the polyfill's code anyway.

JakeChampion:
The WTFPL lets you relicense the work under any other license. What do you
think to relicensing to an OSI approved license?

triblondon:
Since we're not distributing it, I don't think there's actually any opportunity for us to relicense it. I think this is just a low-priority question to raise with legal when you get a chance, but I don't think it stops us merging this.

PR #1062, opened 2017-01-17
add Element polyfills: childElementCount, firstElementChild, lastElem…

ganlanyuan:
…entChild, nextElementSibling, previousElementSibling


polyfills/Element/prototype/childElementCount/config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "browsers": {
+    "firefox": "<3.5",
+    "ie": "6 - 8",
+    "opera": "<10.0",
+    "safari": "<4"
```
triblondon: Can you add targeting for the remaining browser families here?  Does the polyfill work on iOS or Android, for example?  Blackberry, samsung?

[also applies to all the other configs included in this PR]

polyfills/Element/prototype/childElementCount/config.json
```diff
@@ -0,0 +1,12 @@
...
+  },
+  "license": "ISC",
+  "repo": "https://github.com/Alhadis/Fix-IE/blob/master/src/IE8-child-elements.js",
```
triblondon: `repo` links should be to the base of the repo, in this case `https://github.com/Alhadis/Fix-IE`.  I appreciate that the link to the file itself may be more useful, but we should follow the schema here

triblondon (commented):
Thanks for the contribution!  My concern about this is that the polyfill is not transparent - it requires some action on the developer's part to make it work, notably:

https://github.com/Alhadis/Fix-IE#ie8pp

The polyfill is primarily useful for IE8, uses defineProperty to add its stuff, and IE8's defineProperty doesn't work without patching your constructors.

If that turns out to not be an issue, would it be possible to import the polyfill from source rather than copying it?  You can use the `install` section of the polyfill config file for this:

https://polyfill.io/v2/docs/contributing/authoring-polyfills#configuration

Since the author has bundled multiple polyfilled properties in a single file, you would need to import it under one name, and alias the others.

ganlanyuan:
Hey, 
For the IE8PP hack concerning, I think it should be okay. Because the IE8PP hack is only for [Object.defineProperties](https://github.com/Alhadis/Fix-IE/blob/master/src/IE8-defineProperty.js) polyfill, doesn't mater with the polyfills added here.    

Since the source file is a single file, do you think I can add the multiple polyfills as a single file? The folder construction will be something like *Element/ElementTraversal/* or directly *ElementTraversal/*. But I actually don't know how to fetch external file source with Nodejs.   

Can you give a sample beside
```javascript
'use strict';
var path = require('path');
var fs = require('fs');
// ... download/import latest source here
// ... do any required source modification
var newSource = 'test';
fs.writeFileSync(path.join(__dirname, 'polyfill.js'), newSource);
```

polyfills/Element/prototype/childElementCount/config.json
ganlanyuan: The browsers didn't list here support the feature. They don't need the polyfill. In this case, what should I write? 

```javascript
"chrome": "*",
```
means all versions of chrome need the polyfill, right?

polyfills/Element/prototype/childElementCount/config.json
ganlanyuan: Thanks for the tips. I will update this field.

polyfills/Element/prototype/childElementCount/config.json
JakeChampion: That's correct, `"chrome": "*",` will serve the polyfill to all chrome browsers

ganlanyuan: Thanks,
Then what to write if all chrome browsers don't need it?

JakeChampion:
@ganlanyuan I think our documentation needs to surface the simpler system for using third-party polyfills. -- https://polyfill.io/v2/docs/contributing/authoring-polyfills#configuration
>`install`: Object, used to specify how to install the third-party polyfill into the polyfill-service.
module: String, name of the package on NPMJS to install.
>`paths`: Array, paths of the files inside the package to combine together to create the polyfill. If not included, defaults to the packages default entry point.
>`postinstall`: String, path to a javascript module to execute after creating the polyfill. This can be used to create any helper files for the polyfill.

It can be achieved via `config.json` completely :-)
Here is how it would look:
```js
{
  "browsers": {
    "firefox": "<3.5",
    "ie": "6 - 8",
    "opera": "<10.0",
    "safari": "<4"
  },
  "license": "ISC",
  "repo": "https://github.com/Alhadis/Fix-IE/",
  "spec": "https://www.w3.org/TR/ElementTraversal/#attribute-childElementCount",
  "docs": "https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount",
  "install": {
    "module": "fix-ie",
    "paths": ["src/IE8-child-elements.js"]
  }
}
```

ganlanyuan:
Moved `childElementCount`, `firstElementChild`, `lastElementChild`, `previousElementSibling` and `nextElementSibling` to **ElementTraversal** folder.   
Updated `repo` and `install` in config.json.

polyfills/ElementTraversal/config.json
```diff
@@ -0,0 +1,24 @@
...
+  "license": "ISC",
+  "repo": "https://github.com/Alhadis/Fix-IE",
+  "spec": "https://www.w3.org/TR/ElementTraversal",
+  "docs": [
+    "https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount",
+    "https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/firstElementChild",
+    "https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/lastElementChild",
+    "https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/previousElementSibling",
+    "https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling"
+  ],
```
triblondon: Docs is a string property, so you can only link to one of these pages. You could put the rest in `notes`.

polyfills/ElementTraversal/config.json
```diff
@@ -0,0 +1,24 @@
+{
```
triblondon: The name here is a problem because `ElementTraversal` is not a web platform specced global. You can either use a descriptive name prefixed with `~` as we've done for `~html5-elements`, or use one of the real properties exported by the polyfill, eg `Element/protoype/firstElementChild` and list the others as `aliases` in the config.  This configuration option is documented in the contributors guide.

My preference would be for you to do the latter because then I can request any of the included features individually using their real API names in the `features` URL argument, eg `/v2/polyfill.js?features=Element.prototype.firstElementChild`.

polyfills/ElementTraversal/polyfill.js
```diff
@@ -0,0 +1,43 @@
+(function(){
```
triblondon: If you are importing a 3rd party polyfill using the config `install` mechanism, you don't need to commit the original source, it's imported automatically at build time.

triblondon (changes requested):
Thanks, we just need a few tweaks here.  Regarding the utility of this given the defineProperties issue, I haven't tested that but I'll chat to Jake about it.

ganlanyuan:
Hey,
I decided to write these polyfills by myself since they're not that complex.
Please check!

JakeChampion:
Could you add tests for these features please? That way we know for certain
they work as expected :-)

ganlanyuan:
tests added

triblondon:
The features you've targeted to IE8 - do they work in 7 or 6?  Consider using `<9` if the feature was introduced in 9 and the polyfill works in earlier browsers.  Otherwise this looks good.  Will leave to @JakeChampion to push a local branch for CI.

ganlanyuan:
They don't work in IE6-7, because IE6-7 don't support `Object.defineProperty()`. 
I searched a lot, couldn't find a way to fix this issue.

triblondon (approved):

JakeChampion:
@ganlanyuan The tests seem to be failing on CI, would you be able to look into these?

```
Failures:
 - ie/8:
    -> Element.prototype.nextElementSibling
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.nextElementSibling
    -> Element.prototype.previousElementSibling
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.previousElementSibling
    -> Element.prototype.firstElementChild
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.firstElementChild
 - ie/7:
    -> Element.prototype.textContent
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.textContent
```

ganlanyuan:
How can I reproduce the test?

JakeChampion:
@ganlanyuan You can run the application locally using `npm start` and then visiting the urls in the error block posted above. https://github.com/Financial-Times/polyfill-service/pull/1062#issuecomment-303223836

ganlanyuan:
Hey,
when I run `npm start` I got an error says
```
No polyfill sources found.  Run `npm run build` to build them
```
Then I ran `npm run build`, got another error
```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! polyfill-service@3.18.1 build: `npm run clean && node tasks/node/updatesources && node tasks/node/buildsources`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the polyfill-service@3.18.1 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
```

What else I should do?

JakeChampion:
@ganlanyuan I'll need some more information to help with this:
- Which operating system are you using?
- Which version of Node (`node -v`) and npm (`npm -v`)are you using?
- Have you ran `npm install` and was it successful?

Hopefully with that information I can help figure out why the build script is failing.

ganlanyuan:
- Mac v10.12.4
- node v6.10.0, npm v5.0.2
- Ran `sudo npm install` successfully

JakeChampion:
@ganlanyuan I think it is an issue with using npm5 and this project. I believe running the commands separately works:
- `npm run clean`
- `node tasks/node/updatesources`
- `node tasks/node/buildsources`

ganlanyuan:
- `npm run clean` - done
- `node tasks/node/updatesources` - error
```
{ Error: Cannot find module 'web-animations-js/web-animations.min.js'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.resolve (internal/module.js:27:19)
    at polyfillSourcePaths.map.p (/Users/designer01/Documents/ganlanyuan/polyfill-service/tasks/node/updatesources.js:25:27)
    at Array.map (native)
    at installPolyfill (/Users/designer01/Documents/ganlanyuan/polyfill-service/tasks/node/updatesources.js:25:10)
    at Array.forEach (native)
    at glob.then.files (/Users/designer01/Documents/ganlanyuan/polyfill-service/tasks/node/updatesources.js:59:14) code: 'MODULE_NOT_FOUND' }
```

JakeChampion:
@ganlanyuan Thanks for helping debug this issue, if you pull down the changes I've made to your branch, it should be working :-)

JakeChampion (approved):

PR #1063, opened 2017-01-17
Separate Promise.prototype.finally from Promise polyfill

JakeChampion:
The reason for the separation is because modern browsers support `Promise` but not `Promise.prototype.finally`. If we bundle this method with the `Promise` polyfill then users who wanted this method on modern browsers would have to download the whole `Promise` polyfill, which is over 4KB.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2292,9 +2292,9 @@
       "resolved": "https://registry.npmjs.org/xtend/-/xtend-4.0.1.tgz"
     },
     "yaku": {
-      "version": "0.17.6",
-      "from": "yaku@>=0.17.6 <0.18.0",
-      "resolved": "https://registry.npmjs.org/yaku/-/yaku-0.17.6.tgz"
+      "version": "0.16.7",
+      "from": "yaku@0.16.7",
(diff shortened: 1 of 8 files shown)


triblondon (commented):
Can we keep the main promise polyfill as the latest version of yaku?  Otherwise we are stuck and can't upgrade it.  The effect should just be that browsers with no promise implementation at all will get both the full promise polyfill AND a second implementation of `finally` (not ideal but not terrible).

Or, if we want to be fancy about it, we can tweak the browser targeting so that browsers to which we target `Promise` do not also get targeted for `finally`.  A `notes` entry on `finally` would be a good place to document that.

The feature is a stage 0 draft and has no implementations in shipping browsers - I think we need to see at least an intent to implement before we ship this.

JakeChampion:
The proposal is [stage 2](https://github.com/tc39/proposals/commit/a6696379bcec50cea8b3a32496dcaa4d00baf636), I can see the specification document needs updating to reflect that, sorry for the confusion caused.

I'll revert the Yaku change so that we carry on using the latest version (which includes `Promise.prototype.finally`). I'm tempted to not include the `Promise.prototype.finally` tests in the `Promise` test suite as it would show all modern browsers as not supporting Promises in our compat data which we display on the features page of the website.

rowanmanning (approved):

PR #1067, opened 2017-01-19
Update polyfills config for Samsung Internet 5

AdaRoseCannon:
This PR is to update the configuration of the polyfills to account for the recently released Samsung Internet 5.0. This will also be valid for subsequent 5.x releases, such as the 5.2 Beta.

Note: Promise test is failing because of absence of finally but that issue is present in the other browsers set to not receive the polyfill, left it unpolyfilled to match behaviour.

Note: Can not get intersectionObserver tests to pass regardless of presence of polyfill.

### DIFF
--- a/polyfills/Array/from/config.json
+++ b/polyfills/Array/from/config.json
@@ -19,7 +19,7 @@
 		"firefox_mob": "<32",
 		"android": "*",
 		"ios_saf": "<9",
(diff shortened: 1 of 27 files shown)


JakeChampion:
Promise finally issue will be resolved via #1063

JakeChampion (approved):

PR #1069, opened 2017-01-20
Remove Docker to simplify and speed up deployments to Heroku

JakeChampion:
Using the `heroku-postbuild` command in the npm scripts, we can run any tasks after our npm dependencies have been installed but before our application starts up. I'm using this command to run our `build` task, which builds all our third-party polyfills.

The reason for moving away from the Docker approach for deployment is because:
- There have been a few issues with Docker applications on Heroku.
- Pushing images to Heroku Docker registry is very slow and fails frequently to complete.
- Aligns the deployment approach to most other Heroku NodeJS applications.


triblondon (approved):
The reason for doing the docker thing is primarily so that a single build artefact can be tested, deployed to QA and deployed to production, rather than potentially building one thing for testing and then building something slightly different by accident on prod due to differences in environment, dependencies or other non-deterministic elements of the build process.

Pipelines solves this in respect of QA->Prod promotion provided that internally it is not rerunning the build process on prod and the environments are identical (which I guess we don't know as using pipelines makes this heroku's problem?)

Overall I have no objection and any PR with more red than green sounds good to me.

JakeChampion:
When promoting slugs between apps, Heroku will copy the slug verbatim between apps, without changes. Heroku will also not run any scripts when promoting.

Issue #1065, opened 2017-01-18
IE11 mobile is missing Element.prototype.matches


CheloXL:
IE11 mobile needs (just like IE11 desktop) Element.prototype.matches.
Running the following query correctly detects the browser, but the matches polyfill is not included...

`https://cdn.polyfill.io/v2/polyfill.js?ua=Mozilla/5.0%20(Mobile;%20Windows%20Phone%208.1;%20Android%204.0;%20ARM;%20Trident/7.0;%20Touch;%20rv:11.0;%20IEMobile/11.0;%20Microsoft;%20Lumia%20535)%20like%20iPhone%20OS%207_0_3%20Mac%20OS%20X%20AppleWebKit/537%20(KHTML,%20like%20Gecko)%20Mobile%20Safari/537`

triblondon:
We don't target this polyfill to IE mobile at all.  If you can test it and see that it works, feel free to add that browser family here:

https://github.com/Financial-Times/polyfill-service/blob/v3.16.0/polyfills/Element/prototype/matches/config.json#L11-L23

See here for instructions:

https://polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers

CheloXL:
Ah!, Didn't know... since the service correctly detected ie mobile, I supposed it was supported. Will check your instructions, as ie11 mobile is exactly the same as ie11 desktop.

Thanks!

CheloXL:
@triblondon Can I ask for some help on this?
I'm trying to run the tests on windows. After running the npm install, I'm getting several errors on the npm run build because the command "rm" doesn't exists on windows (even under powershell).

Since build doesn't run, neither run dev.

Also, I'm trying to understand why running the page at https://polyfill.io/v2/docs/features/ and selecting the "live test" on my mobile shows matches as "supported with Polyfill service" (that ok, as I noted above ie11mob is the same as ie11), but then, the test at https://polyfill.io/test/tests?feature=Element.prototype.matches&mode=all also passes on my mobile and I expected it to not pass.

triblondon:
@CheloXL we may have broken the build process on Windows.  I'm very sorry about that, the core team are all OSX and Linux and we don't regularly need to build on Windows.  But we'd be very grateful for any fixes you can contribute - eg by replacing the offending shell exec with a cross platform npm module?

[...]

PR #1071, opened 2017-01-26
Ie11 mobile matches

CheloXL:
Hope this time I get this correctly...
wp_ss_20170126_0001
wp_ss_20170126_0002


CheloXL:
Ok.. didn't get it correctly on the first try... now it should be ok.

JakeChampion (approved):

JakeChampion:
@CheloXL This will also serve the polyfill to Edge mobile, is that the intention? If not, the range should be `10 - 11`.

CheloXL:
@JakeChampion Mmm... I don't have Edge mobile to test. I guess since Edge supports matches, it should not be needed. But again, I don't have a way to test. Will change as you suggested and let others test on Edge mobile

JakeChampion (approved):

triblondon:
Fixes #1065

PR #1077, opened 2017-01-31
Adds Set to IOS7

CheloXL:
This fixes the following error on IOS7: "ReferenceError: Can't find variable: Set". On IOS7, the service was servicing Map but not Set to IOS7.
Not sure if this was a bug or not, but on an application I'm using both Map and Set I started to get error reports like the above.
I also adjusted the Safari version just to match what it's being served based on the version number.


CheloXL:
This is so wrong... will commit again.

PR #1078, opened 2017-01-31
Add Set to IOS7

CheloXL:
This fixes the following error on IOS7: "ReferenceError: Can't find variable: Set". On IOS7, the service was servicing Map but not Set to IOS7.
Not sure if this was a bug or not, but on an application I'm using both Map and Set I started to get error reports like the above.
I also adjusted the Safari version just to match what it's being served based on the version number.
Data taken from http://kangax.github.io/compat-table/es6/#test-Set
I don't have any devices with IOS7 to test the commit, but it should work (Map is actually working).


triblondon (approved):

JakeChampion (approved):

asilvas:
Can we bump this PR to the next release if it's holding up `3.17`?

triblondon:
@asilvas we're battling with a CI issue at the moment, so bear with us.

PR #1082, opened 2017-02-04
add polyfill of Object.values

JounQin:


triblondon (changes requested):
The property this adds is enumerable, but the native version is not. Would it be better to use `Object.defineProperty` to create this property with the correct descriptor?

image

triblondon:
@JounQin Are you able to finish off this PR and help us get it merged into the project?

JounQin:
@triblondon 
Sorry, I was a little busy there days, and just forgot to do that.
I've fixed it just now.

JakeChampion (approved):

JakeChampion:
@triblondon, this is ready for another review :-)

polyfills/Object/values/config.json
```diff
@@ -0,0 +1,27 @@
+{
+	"aliases": [
+		"es2017"
```
triblondon, line 3: Is ES2017 technically ES8?  I'm losing track.  We currently have a lot of polyfills aliased to `es6`, but not many attached to the year-based names and none attached to `es7` or `es8`.  We might need some policy here, @JakeChampion - thoughts?

polyfills/Object/values/polyfill.js
```diff
@@ -0,0 +1,14 @@
+(function () {
+	function values(object) {
+		return Object.keys(object).map(function (key) {
+			return object[key];
+		});
+	}
+
+	Object.defineProperty(Object, 'values', {
+		configurable: true,
+		enumerate: false,
```
triblondon: `enumerate` is not a valid object descriptor property.  Should be [`enumerable`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Description).  It is already false by default, but technically required, so we should include it.

polyfills/Object/values/polyfill.js
```diff
@@ -0,0 +1,14 @@
...
+		value: values,
```
triblondon: You could inline the function anonymously here to avoid creating the overhead of an additional symbol.

triblondon (dismissed):
Thanks @JounQin.  Just one bug you need to fix and then I'm happy with it.  Jake and I need to look at aliases more generally but I'm happy for it to not block this PR.

polyfills/Object/values/config.json
JounQin: Actually I just copy the `config.json` from [Object.entries](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object/entries/config.json).

PR #1090, opened 2017-02-20
Update all the dependencies

triblondon:
Ran `npm-check` and updated all deps that had non-breaking updates available.  Also updated major version of dotenv, which no longer offers the `silent` option but performs that behaviour by default.


JakeChampion (approved):

package.json
```diff
@@ -53,63 +53,64 @@
...
-    "graphite": "0.0.6",
-    "handlebars": "^4.0.5",
+    "graphite": "0.0.7",
+    "handlebars": "^4.0.6",
     "html5shiv": "^3.7.3",
-    "intersection-observer": "^0.2.0",
+    "intersection-observer": "^0.2.1",
     "intl": "^1.2.5",
-    "js-polyfills": "^0.1.22",
+    "js-polyfills": "^0.1.32",
```
JakeChampion: This dependency is where our URL polyfill comes from, it looks to be failing in CI with this update.

triblondon: Yep, I saw this.  I wasn't able to run the tests locally as I don't have the browserstack credentials and there are three sets in lastpass (not clear which one to use) and when I found out which one to use, it didn't work.  So it would be helpful to fix that.

I'll look into trying to fix this, but if not I'll roll back that dependency.

triblondon: After some investigation I found the source of the problem and [raised a bug on js-polyfills](https://github.com/inexorabletash/polyfill/issues/125).  @inexorabletash pushed a fix and released a new version, so I've pushed that update to rerun our CI.

PR #1092, opened 2017-02-22
Add unit tests for library code

JakeChampion:


lib/sources.js
```diff
@@ -13,7 +15,7 @@ const metadata = {};
 
 // Discover and index all the available polyfills (synchronously so that the index is available for the first request)
 try {
-	configuredAliases = require(path.join(polyfillsPath, 'aliases.json'));
+	configuredAliases = require('../polyfills/__dist/aliases.json');
```
triblondon, line 24: What's the thinking here?

lib/UA.js
```diff
@@ -145,16 +151,16 @@ function UA(uaString) {
 			if (typeof aliases[this.ua.family] === 'function') {
 				Object.assign(this.ua, aliases[this.ua.family](this.ua));
 
-			// Map to different family, use same version scheme
+				// Map to different family, use same version scheme
```
triblondon: Some weird indentificationing going on here?

test/unit/lib/sources.js
```diff
@@ -0,0 +1,190 @@
...
+		spy.restore();
+		assert.equal(spy.lastCall.args[0]('aliases.json'), false);
+		assert.equal(spy.lastCall.args[0]('example.json'), false);
+	});
+
+	it('catches errors when reading the polyfill directory, logs a friendly error message and exits', () => {
+		const error = new Error;
+		fs.readdirSync.throws(error);
+		assert.doesNotThrow(() => require('../../../lib/sources'));
+		assert.calledWithExactly(consoleMock.log, "No polyfill sources found.  Run `npm run build` to build them");
```
triblondon: Asserting the exact error text seems a tad pedantic - maybe enough to confirm that we log something?

triblondon (commented):
What's the intended relationship between test/unit and test/node?  Seems like some crossover, especially on aliases which looks like the unit version duplicates basically the same tests?

lib/sources.js
JakeChampion: `require` already handles all path related issues

JakeChampion:
@triblondon The existing tests are closer to integration tests than unit tests. There is definitely some cross-over between `test/unit` and `test/node`. With `test/unit` I'm looking to have fully isolated unit tests with which to run coverage reports against where we have purposefully hit a specific code path. Coverage reports will enable us to find out what parts of the code-base are not covered in test cases.

I'm hoping we can split our tests up into unit and integration tests.

triblondon:
@JakeChampion let me know when you want a fresh review of this

.gitignore
```diff
@@ -24,3 +24,5 @@ npm-debug.log
 /test/results
 # Generated by BrowserStack's Tunnelling application
 /local.log
+coverage/
+.nyc_output/
```
triblondon: These are ignored at any point in the project tree.  Did you want to ignore them only at the root?

lib/UA.js
```diff
@@ -79,11 +84,11 @@ const aliases = {
...
 	},
 
-	"chrome mobile ios": "ios_chr",
+	"chrome mobile ios": "ios_chr", // comes back as ios_saf
 
 	"mobile safari": "ios_saf",
-	"iphone": "ios_saf",
-	"iphone simulator": "ios_saf",
+	"iphone": "ios_saf", // can't find a uastring which useragent returns iphone as family
+	"iphone simulator": "ios_saf", // useragent returns either mobile safari or mobile safari uiwebview
```
triblondon, line 91: Should we remove these aliases if they are not working?

lib/index.js
```diff
@@ -49,26 +47,26 @@ function getOptions(opts) {
...
 
+const resolveAliases = createAliasResolver(
+	function aliasFromConfig(featureName) {
+		return sourceslib.getConfigAliasesSync(featureName);
+	},
+	function aliasAll(featureName) {
+		return (featureName === 'all') ? sourceslib.listPolyfillsSync() : undefined;
+	}
+);
+
```
triblondon: Why move this to a larger scope?

test/unit/lib/UA.js
```diff
@@ -0,0 +1,391 @@
...
+			test = UA.normalize("Mozilla/5.0 (iPad; CPU OS 6_1_3 like Mac OS X) AppleWebKit/536.26 (KHTML, like Gecko) Mobile/10B329 [FBAN/FBIOS;FBAV/6.0.2;FBBV/183159;FBDV/iPad3,1;FBMD/iPad;FBSN/iPhone OS;FBSV/6.1.3;FBSS/2; FBCR/;FBID/tablet;FBLC/en_US;FBOP/1]");
+			assert.equal(test, "ios_saf/6.1.0");
+
+			test = UA.normalize("Mozilla/5.0 (iPhone; CPU iPhone OS 6_1_3 like Mac OS X) AppleWebKit/536.26 (KHTML, like Gecko) Mobile/10B329 [FBAN/FBIOS;FBAV/6.0.2;FBBV/183159;FBDV/iPhone4,1;FBMD/iPhone;FBSN/iPhone OS;FBSV/6.1.3;FBSS/2; FBCR/AT&T;FBID/phone;FBLC/en_US;FBOP/1]");
+			assert.equal(test, "ios_saf/6.1.0");
+		});
+	});
+
+	describe(".isUnknown", function () {
+		it("should resolve false for user agents we have a baseline version for", function () {
```
triblondon, line 301: "...should resolve false for user agents that are from known families and above our baseline"

Existing wording a bit confusing.

test/unit/lib/aliases.js
```diff
@@ -0,0 +1,354 @@
...
+					});
+				});
+			});
+
+
+		});
+
+	});
+
+});
```
triblondon: There's a bit of extraneous whitespace in these closing braces.  Also, personally I preferred the more compressed format for the data which make it more likely that you could fit the entire test on the screen at once, but it's not a big point, and I don't really care that much

test/unit/lib/sources.js
triblondon: Still think this is a fragile test

test/unit/lib/sources.js
```diff
@@ -0,0 +1,246 @@
...
+		assert.isFunction(sources.streamPolyfillSource);
+	});
+
+	it('has a getPolyfill method', () => {
+		sources = require('../../../lib/sources');
+		assert.isFunction(sources.getPolyfill);
+	});
+
+	it.skip('loads the polyfill aliases file', () => {
+		// TODO
```
triblondon: TODO?

triblondon (commented):
Also noticed we still have a test suite for lib/index.js in tests/node and also in tests/unit.  If you want to keep the tests/node one as an integration test because it covers more than just that one file, maybe it would be good to rename it endtoend.js or integration.js or something like that.

lib/UA.js
JakeChampion: Yes, will be in a subsequent PR

rowanmanning (approved):
This looks great 👍

PR #1093, opened 2017-02-23
Fixed Object.getOwnPropertySymbols for Object.prototype

### LINKED ISSUE #1058: IE11 bug with Symbols and Enumerables exposed by lodash.omit
[Bountysource](https://www.bountysource.com/issues/40934471-ie11-bug-with-symbols-and-enumerables-exposed-by-lodash-omit?utm_source=40934471&utm_medium=shield&utm_campaign=ISSUE_BADGE)

I've been trying to track this one down for a while now, and though I know where the failure is, it isn't clear as to the best resolution.

Here is an isolated codepen that demonstrates that when poly+lodash+IE11 is combined, will blow up with:

```
"TypeError: Unable to get property '@@__symbol:iterator0.0365778717409362961' of undefined or null reference
   at propertyIsEnumerable (https://cdn.polyfill.io/v2/polyfill.js?features=default,Array.prototype.@@iterator:244:3)
   at Anonymous function (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:6056:9)
   at arrayFilter (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:603:7)
   at Anonymous function (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:6055:7)
   at Anonymous function (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:6070:9)
   at baseGetAllKeys (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:3064:7)
   at getAllKeysIn (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:5903:7)
   at Anonymous function (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:13485:7)
   at apply (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:496:15)
   at Anonymous function (https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js:6600:9)"
```

http://codepen.io/asilvas/pen/VPaOjz

Deps:
* https://cdn.polyfill.io/v2/polyfill.js?features=default,Array.prototype.@@iterator
* https://cdn.jsdelivr.net/lodash/4.17.4/lodash.js

This is the line of the failure: https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Symbol/polyfill.js#L84

Where `this[internalSymbol]['@@' + uid]` assumes that `this[internalSymbol]` is always defined, but is not the case. Of course, might just be able to add a conditional and be done with it, but appears this code behaves this way for a reason. So I'm guessing a resolution is needed elsewhere.

I've also been able to force this error (by overriding native polyfills) in other browsers like Chrome, but the codepen won't do that in its current form.

What appears to be a near identical issue was found before with core-js in reference to lodash: https://github.com/lodash/lodash/issues/2308

Which was supposedly fixed. It's not clear if that fix could somehow help in addressing this.

rhalukongor:
This was causing deep object clones (like lodash.omit) to fail.
Fixes #1058


rhalukongor:
I need some help here: 
Unfortunately I'm out of Browserstack minutes to run the whole suite so I'm testing CircleCI's failed cases (followed link below to the console output) on BrowserStack live. Tried many of them and they all seem to be passing. Clearly I'm doing something wrong but can't figure out what. Any help is appreciated...

EDIT: Then, all of a sudden they all passed. 🤔

JakeChampion (approved):

PR #1094, opened 2017-02-23
Enable service to work correctly when hosted on a path

JakeChampion:
This PR adds the ability for the polyfill-service to be hosted on a path such as `www.example.com/__polyfill`.

The reason for adding this feature is to let users' who self-host the service be able to gain performance benefits by having their requests all come from the same domain, E.G. HTTP/2 multiplexing and opening a minimum amount of TCP connections.


rowanmanning (approved):

triblondon (commented):
Does FT need this?  Not sure it's worth adding this complexity unless we have a business need?  I thought Next FT were proxying to the service from their own VCL so the path change is not an issue?

JakeChampion:
Next FT are moving away from their own proxy. Origami have moved their services to a path on ft.com. Having a version of the Polyfill-Service on a path on ft.com enables us to record logs and do other FT-specific changes to the service without impacting our users on https://polyfill.io

triblondon (approved):
OK, fine with me

JakeChampion:
Can't merge this until https://github.com/Financial-Times/o-in-page-nav/issues/8 is fixed

PR #1098, opened 2017-02-24
Run polyfill service in multiple regions and enable automatic failover if a region is unhealthy

JakeChampion:
I've moved all backend configuration into VCL files so that we can store this in git rather than using the Fastly GUI. 

We can route traffic to the nearest polyfill-service server based on the requesters location. Users' near the US can use the US server, users' nearer the EU will use the EU server. We can also use Fastly's healtcheck system to redirect traffic to a specific region if a server has been marked as unhealthy.


rowanmanning (approved):

wheresrhys:
Out of interest, could shielding be made to work for polyfill service?

triblondon (approved):
I'm slightly wary of creating a top level dir called 'fastly' - would it be better to call it 'deploy', maybe, in case we want to store additional orchestration info together with it in future?  Or maybe 'build-config' or 'config'?

Do we have to have the Heroku apps called totally inconsistent names?! :-)  The `ft-` prefix makes sense to me since the namespace is global. The names should be changable without consequence since they are not publicly advertised.

Otherwise fine.

PR #1100, opened 2017-02-27
Update Opera configurations for ES6 requirements

dlebech:
This PR updates the configuration for some of the ES6 requirements that are still tagged as incompatible with Opera, although they have been available several Opera versions back.

The versions are very conservative version choices based on the compatibility table [here](https://kangax.github.io/compat-table/es6/).


JakeChampion:
@dlebech Thanks for the PR! Are you able to run our test suite against the minimum Opera version each polyfill is now no longer targeting and confirm the native features work as expected? You can do this by visiting https://polyfill.io/v2/docs/features/ and click the "Live test" radio button at the top of the table.

dlebech:
@JakeChampion Neat feature :smiley: I ran Opera 34 in a Windows 7 VM and I actually get "Native" on all the included features. I then installed version 32 which has [feature parity with Chrome 45](http://www.opera.com/docs/history/#32) that seems to have introduced most of `Array` features. Indeed, still native -- across the board.

I have updated the PR so all the minimum versions are 32. I think perhaps some of the other features (like `MIN_SAFE_INTEGER`) were introduced earlier, but if I can avoid going hunting for exact versions, I would appreciate it :smiley: 

What do you think?

JakeChampion:
@dlebech Thanks for doing that!

JakeChampion (approved):

PR #1101, opened 2017-02-28
Add customElements polyfill

alexlafroscia:
Adds a new polyfill for the Custom Elements spec.  As mentioned in #429, this is the "recommended" polyfill when working with frameworks like Polymer and Skate.js, which are layers on top of the "real" API for custom elements.

A few things that I'm not sure of, and would appreciate feedback on from a maintainer:

- I didn't add a ton in the way of tests; what's the expectation? The polyfill itself is really heavily tested, so I really just wanted to make sure that it's actually working in the most basic sense in browsers that need it
- For browsers that don't support native ES6 classes, an additional shim is required. I mentioned it in the "notes" section of the config file, but would it be worth trying to load it as it's own "dependency" polyfill? Or just let people handle it on their own? Likely just let people handle it themselves, but it would be kind of cool to treat the shim as a kind of polyfill that we can help people set up automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "node": ">6.0.0"
   },
   "dependencies": {
-    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
+    "@webcomponents/custom-elements": "github:webcomponents/custom-elements#v1.0.0-alpha.4",
     "Base64": "^1.0.0",
     "array.of": "^0.1.1",
     "audio-context-polyfill": "^1.0.0",
@@ -103,6 +103,7 @@
     "yaku": "^0.17.8"
   },
   "devDependencies": {
+    "babili": "0.0.11",
     "browserstack-local": "^1.3.0",
     "eslint": "^3.11.1",
     "fastly": "Financial-Times/fastly#v2.5.0",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "node": ">6.0.0"
   },
   "dependencies": {
-    "@webcomponents/custom-elements": "github:webcomponents/custom-elements#v1.0.0-alpha.4",
+    "@webcomponents/custom-elements": "github:webcomponents/custom-elements#v1.0.0-rc.1",
     "Base64": "^1.0.0",
     "array.of": "^0.1.1",
     "audio-context-polyfill": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "node": ">6.0.0"
   },
   "dependencies": {
+    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
     "Base64": "^1.0.0",
     "array.of": "^0.1.1",
     "audio-context-polyfill": "^1.0.0",
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ npm-debug.log
 /polyfills/atob/polyfill.js
 /polyfills/AudioContext/polyfill.js
 /polyfills/customElements/polyfill.js
+/polyfills/customElements/~native-shim/polyfill.js
 /polyfills/HTMLPictureElement/polyfill.js
 /polyfills/IntersectionObserver/polyfill.js
 /polyfills/Intl/polyfill.js
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ npm-debug.log
 /polyfills/Array/of/polyfill.js
 /polyfills/atob/polyfill.js
 /polyfills/AudioContext/polyfill.js
+/polyfills/customElements/polyfill.js
 /polyfills/HTMLPictureElement/polyfill.js
 /polyfills/IntersectionObserver/polyfill.js
 /polyfills/Intl/polyfill.js
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2,6 +2,11 @@
   "name": "polyfill-service",
   "version": "3.17.0",
   "dependencies": {
+    "@webcomponents/custom-elements": {
+      "version": "1.0.0-alpha.4",
+      "from": "webcomponents/custom-elements#v1.0.0-rc.1",
+      "resolved": "git://github.com/webcomponents/custom-elements.git#1b2ad174411aab7cccb8231214399bb033dd5d49"
+    },
     "abbrev": {
       "version": "1.1.0",
       "from": "abbrev@>=1.0.0 <2.0.0",
@@ -12,6 +17,45 @@
       "from": "accepts@>=1.3.3 <1.4.0",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz"
     },
+    "acorn": {
+      "version": "4.0.4",
+      "from": "acorn@4.0.4",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.4.tgz"
+    },
+    "acorn-jsx": {
+      "version": "3.0.1",
+      "from": "acorn-jsx@>=3.0.0 <4.0.0",
+      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
+      "dependencies": {
+        "acorn": {
+          "version": "3.3.0",
+          "from": "acorn@>=3.0.4 <4.0.0",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz"
+        }
+      }
+    },
+    "agent-base": {
+      "version": "2.0.1",
+      "from": "agent-base@>=2.0.0 <3.0.0",
+      "resolved": "https://npm.n8s.jp/agent-base/-/agent-base-2.0.1.tgz",
+      "dependencies": {
+        "semver": {
+          "version": "5.0.3",
+          "from": "semver@>=5.0.1 <5.1.0",
+          "resolved": "https://npm.n8s.jp/semver/-/semver-5.0.3.tgz"
+        }
+      }
+    },
+    "ajv": {
+      "version": "4.11.5",
+      "from": "ajv@>=4.7.0 <5.0.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.5.tgz"
+    },
+    "ajv-keywords": {
+      "version": "1.5.1",
+      "from": "ajv-keywords@>=1.0.0 <2.0.0",
+      "resolved": "https://npm.n8s.jp/ajv-keywords/-/ajv-keywords-1.5.1.tgz"
+    },
     "align-text": {
       "version": "0.1.4",
       "from": "align-text@>=0.1.3 <0.2.0",
@@ -21,6 +65,11 @@
       "version": "1.0.1",
       "from": "amdefine@>=0.0.4",
       "resolved": "https://npm.n8s.jp/amdefine/-/amdefine-1.0.1.tgz"
+    },
+    "ansi-escapes": {
+      "version": "1.4.0",
+      "from": "ansi-escapes@>=1.1.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz"
     },
     "ansi-regex": {
       "version": "2.0.0",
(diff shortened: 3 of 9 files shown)


triblondon (commented):
Thanks!  I left a few review comments, and overall:

* Absolutely right that there's no need to port an existing test suite, but I'd like to see more than just a single test.  Could you consider including tests for a few more cases?
* In terms of loading a dependent polyfill, we can do that if we have the polyfill in the collection, using the `dependencies` list in the config.  It's good to think about doing this also for other features that the polyfill relies on.
* The native shim script looks like it's not a standalone polyfill, but an additional patch that works with the customelements one.  We'd handle this with a subfolder of customelements, with a `~` prefix and a descriptive label, in this case perhaps `native-shim`.  A separate config file in the subdirectory would install the shim instead of the main polyfill.  For an example (albeit more complex than you need) see the `Intl` polyfill.

alexlafroscia:
RE: tests, I'll look through the existing ones and pick out a handful that seem relevant. 

RE: the optional native shim..

Would people then choose to add the native shim as a separate feature, that has `customElements` as a dependency? I didn't quite follow what you said about loading the shim instead of the main polyfill.

alexlafroscia:
I'm kind of leaning toward just ignoring the native shim entirely right now, and having people bring it in themselves if needed. People would need the shim if:

- They are using ES6 classes that have been transpiled to ES5 to create their custom elements
- The browser is _not_ using the polyfill for `customElements` -- the polyfill can handle the ES5 classes, apparently

Thoughts on just leaving it out, with a reference in the notes about it (and the link to the full documentation)?

alexlafroscia:
Before this gets pulled in, I need to do a little more research on dependent polyfills. I've only tested this in browsers that have native ShadowDOM support, so I need to try running it against browsers like IE that might have dependent polyfills. The README for the original polyfill doesn't specify anything, but Skate.js published this package:

https://github.com/skatejs/web-components/blob/master/src/index.js

Which apparently bundles all required polyfills together. I might need to add some of those to the dependencies here. I'm trying to get to a point where I can drop the file that they provide, and instead pull in just the needed dependencies from Polyfill.io instead 😄

alexlafroscia:
Alright, so it looks like only `Promise` is actually required by the `custom-elements` polyfill, so I added that in.

I've been mulling over how to handle the "native shim" for a while, and am wondering if we could offer is as an optional package in addition to the custom elements polyfill itself.

If someone were to need the native shim, they would want to run that before the polyfill itself is run. Since most people that are adding this to their page would likely have their polyfills are the first script, this would end up with a load order something like:

```
native customElements shim
polyfill.io with the customElements feature
the rest of their scripts
```

this is kind of unfortuate, because now there's this one script that can't be bundled with the rest of their code.

However, if we could optionally provide both, their load order would be more like:

```
polyfill.io with the customElements+native shim feature
the rest of their scripts
```

which is a lot more friendly, I think. The service would just need to make sure that the native shim code is executed before the polyfill. Thoughts?

The reason I'm unsure about it is that it's really not a polyfill at all, so there's never going to be a browser that does not need the native shim. It's kind of an anti-polyfill; it's only needed by the browsers that are *not* polyfilling `customElements` since the polyfill handles ES5 classes just fine from what I understand. I don't want to put something in here that's outside the scope of what your project is trying to do.

triblondon:
OK, so regarding the native shim, I think I didn't put enough effort into explaining how we could host this.  Basically I'm suggesting you create a subfolder of customElements, and a new config.json file inside it, which pulls in the native shim from the custom elements polyfill repo.  When we run the build process, you would then end up with two registered polyfills, `customElements`, and (maybe) `customElements.~native-shim`.

You could then either list the latter in the `dependencies` array of the main polyfill's config, or leave it to the user to consume it explicitly by adding it to the `features` parameter of their polyfill.io request.  Making it a dependency would ensure that we sort it before the polyfill in the output.  However, whether you need the native shim or not seems unrelated to the browser you're using and more to do with whether you are transpiling your code, right?

I think the options would be:

1. It's not a dependency, and the config for the shim targets all browsers.  The user has to explicitly request it.  Ordering is not guaranteed.
2. Make it a dependency, but target no browsers.  The user has to explicitly request it *and additionally include the `always` flag* in order to get it.  The shim is guaranteed to be before the polyfill in the bundle

alexlafroscia:
>  However, whether you need the native shim or not seems unrelated to the browser you're using and more to do with whether you are transpiling your code, right?

Yes, 100% correct

I think that making it a dependency might be a fine solution. It's not very large, and people can always exclude it if they don't need it -- which is only the case if they're shipping ES6 classes without transpilation, which I'm sure very few people are doing. I'll get that added in as well. 

Thanks for the additional explanation, I have a better understanding of what you're suggesting now.

alexlafroscia:
Is there a way to force the `detect` option? The native shim should _only_ be executed when it's actually needed; you can see how the Skate.js folks solve this here:

https://github.com/skatejs/web-components/blob/43a107a76d266c7bc46ff94d32cafad013726233/webpack.config.js#L13

Basically, they wrap the whole polyfill in an `eval` and then execute it only when the native implementation is available.  We would want to do basically the same thing, so I'm wondering if we can leverage the existing `detect` feature to do this, or if we should instead just bake the check into the polyfill the same way that they do.

Unfortunately, if you run the native shim through Babel, it breaks; you need to leave the native classes in the shim as native classes, so you have to be careful not to let the browser execute it if it doesn't have support for ES6 classes.

alexlafroscia:
So, I've been trying to get things working with what i originally thought was a better approach; always include the shim for `customElements` by adding it as a "dependency" of the

alexlafroscia:
Hmm, so after playing with trying to make the native shim a dependency of `customElements`, I'm not sure that I can get it set up in a way that seems intuitive.  So, I think it actually might make more sense to flip that around and set it up such that the `customElements` polyfill is a dependency of the native shim.

- If the developer wants to author components in ES6, they can just include `customElements`
- If the developer wants to author components in ES5, they can just include `customElements.~native-shim`
  - If the browser supports custom elements natively, it will only download the shim and patch the native implementation
  - If the browser does not support custom elements natively, it will end up downloading the polyfill and the shim; the polyfill supports ES5 classes anyway. There is the additional overheard of getting the shim when it's not necessary, but I think that might be OK

alexlafroscia:
Okay, so with changes that I just pushed, it works as described above.

Note that the `customElements` tests fail in browers without ES6 class support (I tested in IE 11) because it inherently requires ES6.  If you need ES5 class support, you _must_ use the `customElements.~native-shim` settings instead, which passes in IE 11 (as well as Chrome and Safari, which I also tried).

I think keeping them separate is probably the easiest approach; this way it should be clear whether the user should include `customElements` with or without the shim, based on whether they are sending ES6 to the browser or transpiling it.

package.json
```diff
@@ -50,7 +50,7 @@
     "node": ">6.0.0"
   },
   "dependencies": {
-    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
+    "@webcomponents/custom-elements": "github:webcomponents/custom-elements#v1.0.0-alpha.4",
```
alexlafroscia: For now, I'm pulling directly from Github, since "alpha 4" has not been published to NPM. I made an issue requesting that it be published there, which can be found here:

https://github.com/webcomponents/custom-elements/issues/68

triblondon:
This looks great, and I should have suggested the dependency the other way around, sorry that didn't occur to me and well done for recognising it.

You can't force a polyfill to include a detect unless you bake it in.  For the next major version of the service we plan to make inclusion of detects the default bahaviour, but right now you need to add the `flags=gated` param to wrap the polyfills in the detects.  Obviously many of our third party polyfills ship with detects already.

The problem with the `eval()` solution is that it won't work in any page that has an active Content-Security-Policy.  I made some efforts a while back to remove all polyfills that use eval or `new Function("")` but ended up realising that the only polyfills that use those patterns are ones that are required in browsers that don't support CSP.  So we're OK.  But in this case, there are CSP-supporting browsers that you would need native-shim in, right?  If that's the case you may run into trouble.

Pulling a dependency from a git URL is not ideal.  I'd ideally like to wait until the version you need is published before we merge.

Regarding the use of babili, can you run that reasoning past me another way - I'm struggling to grok your comment in the update task.

alexlafroscia:
> But in this case, there are CSP-supporting browsers that you would need native-shim in, right? If that's the case you may run into trouble.

Huh, that might be an issue. Not sure how to avoid that, though... Further explanation and thoughts below.

> Pulling a dependency from a git URL is not ideal. I'd ideally like to wait until the version you need is published before we merge.

That's fine, I'm not in any particular rush. Hopefully there isn't a reason it was released to Github but not NPM so that this isn't left in limbo for too long.

> Regarding the use of babili, can you run that reasoning past me another way - I'm struggling to grok your comment in the update task.

Absolutely. Basically, because the native `customElements` implementation _must_ receive an ES6 class, the native shim can't be compiled from ES6 down to ES5.  Because it must be provided as ES6, Uglify can't minify the file for us (only a non-standard branch supports ES6 right now, as far as I know) so I needed to grab a different tool to minify it (the source only provides an un-minified version). Since `babili` is built for exactly the task of minifying ES6, I grabbed and ran the minification manually to strip out the comments and such.

Needing the native shim to be provided to the browser as ES6 is also the reason for needing `eval()` in the polyfill. We don't want browsers that can't support ES6 to attempt to interpret the file, so we hide the ES6 code in a string and `eval` only if necessary.  Because the native shim is only necessary in instances where `customElements` is provided natively, a browser that can't interpret ES6 would never need the shim.  I didn't think about the CSP issues with `eval` though... 

Maybe we could do the feature-detection server-side and only serve the polyfill to browsers that we know can handle ES6?  Can the browser version matching that the `config.json` uses handle something like "only versions of {X browser} greater than Y" etc.? We would need to flip the normal use case of "only versions of {X browser} less than Y" around, and I don't know if it supports that.

In a case where a browser could not handle ES6, we would still want to provide the `customElements` dependency, since that would handle what the shim is doing as well.  Can the server handle this kind of use case?

- Request `customElements.~native-shim`
- Server determines that the browser can't handle ES6, so it skips the native shim
- Server returns just the `customElements` polyfill

I _think_ that if the server determined that the `native-shim` should be ignored, it would also ignore the dependency of `customElements`, but I might be entirely wrong about that. This is my first time trying to add a polyfill to the service!

The whole setup here is a little strange, but it basically falls into a table like so:

| Developer delivers components as | Developer requests           | Browser supports `customElements` | Browser supports ES6 classes | Server should provide              |
|:---------------------------------|:-----------------------------|:----------------------------------|:-----------------------------|:-----------------------------------|
| ES6 Classes                      | `customElements`             | Yes                               | Yes                          | Nothing                            |
| ES6 Classes                      | `customElements`             | No                                | Yes                          | Just the `customElements` polyfill |
| ES5 Classes (functions or babel) | `customElements.native-shim` | Yes                               | Yes                          | Just the `native-shim`             |
| ES5 Classes (functions or babel) | `customElements.native-shim` | No                                | Yes                          | Just the `customElements` polyfill |
| ES5 Classes (functions or babel) | `customElements.native-shim` | No                                | No                           | Just the `customElements` polyfill |

In cases where the browser should not provide the `native-shim`, it _can_ provided it, as long as it's not evaluated since that will cause the the browser to choke on syntax it doesn't understand. Right now, we're solving that with `eval()` but I know it's not an elegant solution.

JakeChampion:
First of, great work @alexlafroscia!

Hopefully I've understood correctly: 
>Maybe we could do the feature-detection server-side and only serve the polyfill to browsers that we know can handle ES6? Can the browser version matching that the config.json uses handle something like "only versions of {X browser} greater than Y" etc.? We would need to flip the normal use case of "only versions of {X browser} less than Y" around, and I don't know if it supports that.

The browser version matching is based on semver ranges, we can do "less than" ranges, we do this for `Promise.prototype.finally` - https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Promise/prototype/finally/config.json

alexlafroscia:
Sorry for disappearing, I was away last week.

Thanks for the kudos, @JakeChampion!

To confirm, that is exactly what I was referring to. With confirmation that that usage is supported, I'll update this PR to use the browser version to only ship the shim to browsers that can actually use it; this way we can avoid having to `eval` the code at all.

alexlafroscia:
Is there a way to configure a polyfill to match _no_ versions of a browser?

Basically, for any browser that needs the `customElements` polyfill, I want to configure the `customElements.~native-shim` to _not_ be served. If the developer specifies that they need the native shim, some browsers should actually get _only_ the `customElements` polyfill and not the shim. We basically never want to ship both the native shim _and_ the polyfill to a browser; just one or the other.

alexlafroscia:
I think this does fall into the same kind of configuration case that `Promise.prototype.finally` does; you only want to provide the polyfill to browsers that have a native implementation of Promise, since the polyfill already includes `.finally`.

In investigating how `Promise.prototype.finally` works, I tried grabbing the polyfill with just `Promise.prototype.finally` specified as the feature I want in IE 11. Instead of returning me the Promise polyfill (which it would require) it gave me nothing. This seems a little counter-intuitive; shouldn't it have given me the polyfill?  Or do you have to specify `Promise,Promise.prototype.finally` as the feature set you want?

Whatever the mechanics of that are, `customElements.~native-shim` would mirror.  I would _like_ it if, to get the Native Shim working in any browser, you only had to specify it alone, instead of `customElements,customElements.~native-shim` since currently, just asking for the native shim in a browser that does not have native `customElements` support just gives you nothing, instead of giving you the polyfill.

alexlafroscia:
I guess my real question is, because I'm worried I may not be being too clear:

Is there some way the config can be authored so that specifying _just_ `customElements.~native-shim` will ship the custom elements polyfill to the browsers that need it, without the shim, and just the shim to the browsers with native custom elements support, without the polyfill? Or does it have to be done by requesting `customElements,customElements.~native-shim` explicitly?

JakeChampion:
@alexlafroscia Is the table from https://github.com/Financial-Times/polyfill-service/pull/1101#issuecomment-283577394 what you are trying to achieve?

alexlafroscia:
@JakeChampion yes

JakeChampion:
@alexlafroscia I believe it has to be done by requesting `customElements,customElements.~native-shim` explicitly? We can add this in the notes section of each polyfills config.

alexlafroscia:
@JakeChampion yeah, I think that's probably going to be the easiest way to go.

Sorry that this has been put on my back burner, I still really want to get this over the finish line. It seems like the custom elements polyfill finally released a 1.0 version a week or so ago, so I'll try to update my PR to use that and see if we can get this landed.

https://github.com/webcomponents/custom-elements/releases/tag/v1.0.0

JakeChampion:
@alexlafroscia If we advise people to use https://github.com/WebReflection/babel-plugin-transform-builtin-classes, would that mitigate the need for the native shim?

alexlafroscia:
@JakeChampion could be.

I hate to say it, but I'm not sure I'm going to be able to finish this. Work's been keeping me really busy lately, and we ended up deciding to build this polyfill into our bundle rather than using polyfill.io. I just don't have time to finish this up, but hopefully someone else that cares about it can pick up the torch.

bedeoverend:
@alexlafroscia out of interest, what's left to be done on this? It feels like it's really close! Thanks for getting it so far, if I get any time would love to pick it up.

alexlafroscia:
It was pretty much just solidifying the approach to specifying whether just the polyfill was desired, or the polyfill _plus_ the shim. The semantics are a little weird, since the shim would be required when the polyfill isn't, but the polyfill includes the shim... It's weird. [I wrote out a chart a while back for reference.](https://github.com/Financial-Times/polyfill-service/pull/1101#issuecomment-283577394)

Also rebasing the PR to master, since there were some conflicts.

I know it was close, I just know that I won't have time to finish it. It would be really cool to see someone get it over the finish line.

bedeoverend:
Just clarifying the final syntax, the user should request `customElements,customElements.~native-shim` if they want to guarantee CE support w/ transpiled JS? Requesting just `customElements` won't ever pull in `native-shim` and similarly requesting just `customElements.~native-shim` will never pull in the `customElements` polyfill, correct?

FWIW even though there's a bit more going on for the user to consume customElements, I think it's a better approach as it's clearer when it comes to understanding what polyfill / shim you get served.

bedeoverend:
I opened a new PR #1376, I was just wondering if @JakeChampion you'd have any time to review? There's a few questions / changes I'm sure I'll need to make, but really keen to land this one :)

PR #1103, opened 2017-03-01
Update Array.prototype.values config for Firefox

### LINKED ISSUE #1081: Array.prototype.values not polyfilled in FF
`Array.prototype.values` is only supported in the Nightly Firefox build and should be polyfilled which is not happening right now.

See also [ES6 compat table](http://kangax.github.io/compat-table/es6/#test-Array.prototype_methods_Array.prototype.values)

Siddharth11:
Fixes #1081


triblondon:
Can you document the compat info here?  Some quick research suggests this was shipped in stable and then withdrawn due to [a compat issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1299593).  In that case we should consider a) targeting only the versions that did not ship the native impl (multi-range semver statement), and b) adding a `notes` entry to the config to point to the bugzilla entry.

triblondon:
https://polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers has a checklist of stuff to cover off in the PR.

triblondon:
@Siddharth11 will you be able to help us finish this?

JakeChampion (approved):

PR #1104, opened 2017-03-06
Upgrade to whatwg-fetch@2.0.3

### LINKED ISSUE #1091: Fetch polyfill failing in Microsoft Edge
I'm seeing trouble with calls to `fetch` that have `headers` that are arrays of arrays (which looks like if [follows the spec](https://fetch.spec.whatwg.org/#typedefdef-headersinit)).

For example, this request:

```js
fetch('https://example.com', {
  method: 'GET',
  headers: [
    ['accept', 'application/json'],
    ['content-type','application/json']
  ],
  mode: 'cors',
  credentials: 'same-origin'
});
```

is eventually reaching [this code](https://github.com/Financial-Times/polyfill-service/blob/e09871767dcebe671975d3c1c59c1452db5be235/polyfills/fetch/polyfill.js#L383-L385):
```js
request.headers.forEach(function(value, name) {
  xhr.setRequestHeader(name, value);
});
```

Since `headers` is still an array of arrays, this is essentially:
```js
xhr.setRequestHeader(0, ['accept', 'application/json']);
xhr.setRequestHeader(1, ['content-type','application/json']);
```

Calls to `fetch` are coming from `stream-http`, where it looks like https://github.com/jhiesey/stream-http/commit/11d14fc642f4878e9dc8e2329a62e7a8d18895fe might be a relevant change.  I've filed jhiesey/stream-http#71 in hopes of getting more info on that change.  But it looks to me like this might be an issue with the polyfill.

tschaub:
I've manually copied in `fetch.js` from the `whatwg-fetch@2.0.3` release and then applied the change from d7d42f82adcdd1c879e8645b52945a339c26b0a7.  I'm uncertain if there is another way to update this polyfill (though I have read the [updating doc](https://cdn.polyfill.io/v2/docs/contributing/common-scenarios#update-polyfill)).  Please let me know if you'd like to handle this a different way.

This fixes #1091.


triblondon:
Because we are modifying the polyfill, automating this is a bit tricky.  What you *could* do is create a patch file for the changes we need to make (ie, removing the detect), then write an `update.task.js` in the polyfill folder to apply the patch after any update.  Then we'd update the source automatically and apply the patch each time.  If the patch failed to apply you'd abort the task, and we'd have to revert to an earlier version of the dependency or update the patch.

If you fancy doing this, it would certainly be appreciated.  Here's the documentation on writing update tasks:

https://cdn.polyfill.io/v2/docs/contributing/authoring-polyfills#update-tasks

colthreepv:
Also interested in this PR.
I was tempted to contribute, but the guideline prompts to create a synchronous script, is that even possible to create a synchronous script using node HTTP methods? (to retrieve latest source)

triblondon:
@colthreepv you don't need to download the source.  Looking at the docs, they are misleading on this point and I'll update them.  Basically, you want to do this:

1. Put the whatwg-fetch module in our `dependencies`, or update its semver range, or just update the shrinkwrapped version - as appropriate
2. Put an `install` section into the `fetch` config.json file, to pull out the right file from the dependency.  This will create a `polyfill.js` file in the appropriate polyfill directory
3. Now create a synchronous `update.task.js` to transform the polyfill.js file as you need - in this case to remove the simplistic detect.  I'd prefer to do this by having an update.patch file in the polyfill directory alongside the update.task.js file, and use the task to just apply the patch to the polyfil.js file.  If the patch does not apply, exit with a non-zero exit code to signal the failure.

colthreepv:
Thanks for taking the time to put this checklist, It is feasible this way!

JakeChampion:
This was landed in #1133

PR #1127, opened 2017-03-16
Set the Host header in order to let Heroku's router know which app to route to

JakeChampion:


rowanmanning (approved):

sjparkinson:
Looks 👌 . The only comment I'd make is if both regions are unhealthy ( 🙏  ) we end up sending traffic to the opposite regions which seems a bit funny.

JakeChampion:
@sjparkinson Completely agree, I'm not sure what a better approach would be for that scenario.

PR #1132, opened 2017-03-17
Use brotli compression if browser supports it

JakeChampion:


onishiweb (approved):

triblondon:
The VCL used here comes from a Fastly doc page, I think. If so a comment with a URL would be very useful.

PR #1144, opened 2017-03-26
Stop tests route from blocking the event loop with sync fs methods

JakeChampion:
This might be easier to review by ignoring whitespace and shrinking the size of tabs --> https://github.com/Financial-Times/polyfill-service/pull/1144/files?w=1&ts=2

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,7 +23,7 @@ function listAllPolyfills() {
 }
 
 function describePolyfill(featureName) {
-	return sourceslib.getPolyfill(featureName);
+	return sourceslib.getPolyfillMetaSync(featureName);
 }
 
 function getOptions(opts) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,7 +23,7 @@ function listAllPolyfills() {
 }
 
 function describePolyfill(featureName) {
-	return sourceslib.getPolyfillMetaSync(featureName);
+	return Promise.resolve(sourceslib.getPolyfillMetaSync(featureName));
 }
 
 function getOptions(opts) {
--- a/service/RumReport.js
+++ b/service/RumReport.js
@@ -40,7 +40,6 @@ function Perf(rawopts) {
 	;
 
 	const sqlQuery = `SELECT ${sqlFields} FROM requests WHERE req_time BETWEEN (CURDATE() - INTERVAL ${options.period} DAY) AND CURDATE() AND data_center IS NOT NULL AND perf_req IS NOT NULL LIMIT 1000000`;
-	console.log(options.minSample);
 
 	const dataPromise = MySQL.createConnection(process.env.RUM_MYSQL_DSN)
 		.then(conn => {
--- a/service/RumReport.js
+++ b/service/RumReport.js
@@ -40,7 +40,7 @@ function Perf(rawopts) {
 	;
 
 	const sqlQuery = `SELECT ${sqlFields} FROM requests WHERE req_time BETWEEN (CURDATE() - INTERVAL ${options.period} DAY) AND CURDATE() AND data_center IS NOT NULL AND perf_req IS NOT NULL LIMIT 1000000`;
-console.log(options.minSample);
+	console.log(options.minSample);
 
 	const dataPromise = MySQL.createConnection(process.env.RUM_MYSQL_DSN)
 		.then(conn => {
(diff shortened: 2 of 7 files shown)


lib/index.js
```diff
@@ -23,7 +23,7 @@ function listAllPolyfills() {
 }
 
 function describePolyfill(featureName) {
-	return sourceslib.getPolyfill(featureName);
+	return sourceslib.getPolyfillMetaSync(featureName);
```
triblondon: Is this a breaking change?  Previously returned a promise?

service/RumReport.js
```diff
@@ -40,7 +40,7 @@ function Perf(rawopts) {
 	;
 
 	const sqlQuery = `SELECT ${sqlFields} FROM requests WHERE req_time BETWEEN (CURDATE() - INTERVAL ${options.period} DAY) AND CURDATE() AND data_center IS NOT NULL AND perf_req IS NOT NULL LIMIT 1000000`;
-console.log(options.minSample);
+	console.log(options.minSample);
```
triblondon: Hm, this is probably accidental debug that we should remove

service/RumReport.js
```diff
@@ -151,9 +151,8 @@ function Compat() {
 		})
 		.then(results => {
 			return dbconn.end().then(() => {
-				return Promise.all(results[0].map(rec => {
-					return polyfillio.describePolyfill(rec.feature_name)
-					.then(polyfill => {
+				return results[0].map(rec => {
+					const polyfill = polyfillio.describePolyfill(rec.feature_name);
```
triblondon: Should this be using the sources lib directly rather than the external API, I wonder?

service/routes/test.js
```diff
@@ -40,30 +42,32 @@ function createEndpoint(type, polyfillio) {
...
+							const baseDir = path.join(__dirname, '../../polyfills');
+							const testFile = path.join(baseDir, config.baseDir, '/tests.js');
+							return readFile(testFile)
+								.then(tests => {
+									return {
+										feature: featureName,
+										detect: config.detectSource ? config.detectSource : false,
+										tests
+									};
+								});
```
triblondon: Move semicolon to next line to avoid jumping two indent levels?

tasks/node/buildsources.js
```diff
@@ -233,14 +239,15 @@ Promise.resolve()
         .map(absolute => new Polyfill(absolute, path.relative(src, absolute)))
         .filter(polyfill => polyfill.hasConfigFile)
         .map(polyfill => polyfill.loadConfig()
-            .then(() => polyfill.loadSources())
+			.then(() => polyfill.loadSources())
+			.then(() => polyfill.updateConfig())
```
triblondon: Are you sure your indentation is right here?

service/routes/test.js
```diff
@@ -40,30 +42,32 @@ function createEndpoint(type, polyfillio) {
...
 				return Promise.all(featuresList.map(featureName => {
-					return polyfillio.describePolyfill(featureName).then(config => ({[featureName]: config}) );
-				})).then(featureObjs => Object.assign({}, ...featureObjs));
+					return Promise.resolve()
+						.then(() => {
+
+						const config = polyfillio.describePolyfill(featureName);
+
+						const isTestable = config.isTestable;
+						const isPublic = config.isPublic;
```
triblondon: Maybe no need to reassign these anymore?  It now feels inconsistent with the treatment of `config.hasTests`

tasks/node/buildsources.js
```diff
@@ -115,13 +119,15 @@ class Polyfill {
             .then(data => {
                 this.config = JSON.parse(data);
                 this.config.detectSource = '';
-                this.config.baseDir = this.path.relative;
+				this.config.baseDir = this.path.relative;
```
rowanmanning: A few tab/space indent inconsistencies in this file

lib/index.js
JakeChampion: Yes, removing `getPolyfill` from `lib/sources.js` is also a breaking change.

tasks/node/buildsources.js
JakeChampion: I am sure, seems the files was in spaces :-(

tasks/node/buildsources.js
JakeChampion: Seems the files was using spaces for indentation

lib/index.js
JakeChampion: I made it a non-breaking change now

rowanmanning (approved):
Looks great 👍

PR #1145, opened 2017-03-26
Cache content on CDN for a year, soft purge from CDN when deploying

JakeChampion:
As the content does not change during the lifetime of a deployment, it makes sense to cache the content on the CDN for a really long time and to mark the content as stale whenever a new version of the service is deployed.

This is achieved in this PR by assigning the long surrogate-control headers for all resources (caching on CDN for a long time) and assigning the same surrogate-key to every request (a way to mark each object in the CDN in order to label the objects stale). We already run a purge script whenever deploying to QA or promoting to production. The purge script has been updated to do a soft-purge of the surrogate-key.

### DIFF
--- a/lib/sources.js
+++ b/lib/sources.js
@@ -51,7 +51,7 @@ exports.streamPolyfillSource = function (featureName, type) {
 };
 
 // TODO: deprecate this
-exports.getPolyfill = function(featureName) {
+exports.getPolyfill = function (featureName) {
 	if (features.indexOf(featureName) !== -1) {
 		const getSources = [
 			readFile(path.join(polyfillsPath, featureName, 'raw.js'), 'utf-8'),
--- a/lib/sources.js
+++ b/lib/sources.js
@@ -7,12 +7,6 @@ const readFile = denodeify(fs.readFile);
 const process = require('process');
 const console = require('console');
 
-const streamFromString = require('from2-string');
-const streamToString = require('stream-to-string');
-const cache = require('lru-cache')({
-	max: 5000
-});
-
 const polyfillsPath = path.join(__dirname, '../polyfills/__dist');
 
 let features;
@@ -53,18 +47,7 @@ exports.getConfigAliasesSync = function(featureName) {
 };
 
 exports.streamPolyfillSource = function (featureName, type) {
-	const key = featureName + type;
-	const hit = cache.get(key);
-	if (hit) {
-		return streamFromString(hit);
-	} else {
-		const polyfill = fs.createReadStream(path.join(polyfillsPath, featureName, type + '.js'), { encoding: 'UTF-8' });
-
-		streamToString(polyfill)
-			.then(polyfillString => cache.set(key, polyfillString));
-
-		return polyfill;
-	}
+	return fs.createReadStream(path.join(polyfillsPath, featureName, type + '.js'), { encoding: 'UTF-8' });
 };
 
 // TODO: deprecate this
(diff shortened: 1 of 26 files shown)


JakeChampion:
@phamann - What do you think about this idea?

lib/sources.js
```diff
@@ -53,18 +47,7 @@ exports.getConfigAliasesSync = function(featureName) {
 };
 
 exports.streamPolyfillSource = function (featureName, type) {
-	const key = featureName + type;
```
JakeChampion, line 56: _(This caused a lot of needless garbage-collection pausing due to large polyfill strings in memory.)_

This was added because we had lots of requests hitting our server, which was causing a lot of filesystem access. With the long surrogate-control (cdn caching), we should no longer need to have an in-memory cache of polyfills.

npm-shrinkwrap.json
```diff
@@ -12,11 +12,51 @@
...
+          "from": "acorn@>=3.0.4 <4.0.0",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz",
+          "dev": true
+        }
+      }
+    },
+    "agent-base": {
+      "version": "2.0.1",
+      "from": "agent-base@>=2.0.0 <3.0.0",
+      "resolved": "https://npm.n8s.jp/agent-base/-/agent-base-2.0.1.tgz",
```
JakeChampion: 🤔 I have never used this npm registry...

leggsimon:
Looks ok to me

leggsimon (dismissed):
Looks ok to me

triblondon (commented):
Slightly neater solution might be to use `s-maxage`, then you can include it in the default `Cache-control` header, but when you override the default in the various special case route handlers, you don't have to explicitly set surrogate control to the same thing as cache control.

Another option is to not set a surrogate control header by default, and add it only in route handlers where you want to specifically vary the cache TTL on the Edge (which I think is only in one route handler)

Side note: It's pretty hilarious to look at how many times we've flip flopped on having an in-memory cache of polyfill sources.

tasks/node/purge.js
```diff
@@ -4,67 +4,46 @@ require('dotenv').config();
 
 const argv = require('minimist')(process.argv.slice(2));
 
-const DRY_RUN = argv.dryRun;
 const PRODUCTION = argv.env === 'prod';
```
triblondon: Wondering whether we should be using NODE_ENV here

tasks/node/purge.js
```diff
@@ -4,67 +4,46 @@ require('dotenv').config();
...
+	FASTLY_SERVICE_ID = process.env.FASTLY_SERVICE_ID_QA;
 }
+
+const FastlyPurge = require('fastly-purge');
+
+const fastly = new FastlyPurge(FASTLY_API_KEY, {
+	softPurge: true
+});
+
+const key = 'polyfill-service';
```
triblondon: Maybe hoist this def to the top and name it as a constant, eg `CDN_SURROGATE_KEY`.  Also since this needs to be in sync between this task and the lib code, should it be an env var?

triblondon:
Sorry, I submitted my review before I'd finished commenting :-)

rowanmanning (dismissed):

tasks/node/purge.js
JakeChampion: Good idea!

rowanmanning (approved):

PR #1146, opened 2017-03-27
Prevent Event being applied in web workers, fixes #1110

### LINKED ISSUE #1110: Event polyfill fails in a worker
The `Event` polyfill assumes that [`window` is defined](https://github.com/Financial-Times/polyfill-service/blob/5f374f29fdb7e1ca26bb4f043cedcb7a29b1730c/polyfills/Event/polyfill.js#L35), so it does not work in a worker.

In case it is useful for others searching for the same, this means that the `Promise` polyfill (and anything else that depends on `Event`) does not work in a worker.

triblondon:
Potential solution for #1110, but we don't have a test case for workers, so @tschaub if you have a suggestion for how best to set up a test please let me know.

### DIFF
--- a/polyfills/Event/polyfill.js
+++ b/polyfills/Event/polyfill.js
@@ -20,7 +20,7 @@
 	// This polyfill depends on availability of `document` so will not run in a worker
 	// However, we asssume there are no browsers with worker support that lack proper
 	// support for `Event` within the worker
-	if (!document || !window) return;
+	if (typeof document === 'undefined' || typeof window === 'undefined') return;
 
 	function indexOf(array, element) {
 		var
--- a/polyfills/Event/polyfill.js
+++ b/polyfills/Event/polyfill.js
@@ -17,6 +17,11 @@
 		textinput: 1
 	};
 
+	// This polyfill depends on availability of `document` so will not run in a worker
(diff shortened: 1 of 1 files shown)


tschaub:
Thanks @triblondon.

Here is a simple smoke test: https://bl.ocks.org/tschaub/raw/00cead930df134336fde43f0dc965f52/

This is not adapted for the test setup in this repo, but it could provide clues on a starting point.  Sources here: https://gist.github.com/tschaub/00cead930df134336fde43f0dc965f52

JakeChampion (approved):

polyfills/Event/polyfill.js
```diff
@@ -17,6 +17,11 @@
 		textinput: 1
 	};
 
+	// This polyfill depends on availability of `document` so will not run in a worker
+	// However, we asssume there are no browsers with worker support that lack proper
+	// support for `Event` within the worker
+	if (!document || !window) return;
```
tschaub: This will throw a `ReferenceError` where `document` or `window` are not defined.

Instead, this could be written as `typeof document === 'undefined' || typeof window === 'undefined'`.

JakeChampion (approved):

PR #1157, opened 2017-03-30
Optimize the size tracking operations in the Map/Set polyfills

### LINKED ISSUE #1153: `Map.prototype.size` is O(n)
Why is the `size` accessor O(n)?  It can easily be O(1) as you explicitly know when `set` or `delete` is called.  Also please don't make `calcSize` O(n) in set/delete as you can just use a simple increment or decrement as necessary.

dcherman:
The size will be tracked on a per operation basis rather than being
calculated in a getter which changes this from an O(n) operation to O(1)

fixes #1153

### DIFF
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -30,16 +30,6 @@
 		};
 	}
 
-	function calcSize(mapInst) {
-		var size = 0;
-		for (var i=0, s=mapInst._keys.length; i<s; i++) {
-			if (mapInst._keys[i] !== undefMarker) size++;
-		}
-		return size;
-	}
-
-	var ACCESSOR_SUPPORT = true;
-
 	function hasProtoMethod(instance, method){
 		return typeof instance[method] === 'function';
 	}
@@ -47,6 +37,7 @@
 	var Map = function(data) {
 		this._keys = [];
 		this._values = [];
+		this.size = this._size = 0;
 		// If `data` is iterable (indicated by presence of a forEach method), pre-populate the map
 		if (data && hasProtoMethod(data, 'forEach')){
 			// Fastpath: If `data` is a Map, shortcircuit all following the checks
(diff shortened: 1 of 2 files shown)


JakeChampion:
Thanks for the contribution @dcherman, this will definitely help on some large Sets and Maps!

Would you be able to add the email address used for this commit to your account, then reading our [Contributor License Agreement](https://cla-assistant.io/Financial-Times/polyfill-service?pullRequest=1157) and signing it if you are happy with the terms.

dcherman:
@JakeChampion Will do; waiting to hear back from our legal dept before I can officially sign.  I've read over it and don't expect any roadblocks though, so that should hopefully happen shortly.

dcherman:
@JakeChampion Sorry for the delay, I've both signed the CLA and added that email address to my account

JakeChampion (approved):

leggsimon (approved):

JakeChampion:
@dcherman This is failing on CI, would you be able to look into this at all?

```
Failures:
 - ie/8:
    -> Map
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Map
    -> Set
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Set
 - ie/7:
    -> Map
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Map
    -> Set
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Set
```

JakeChampion (changes requested):
Currently failing on IE7 and IE8

dcherman:
@JakeChampion Ah, I see the problem.  Need to set `size` in the constructor as well since `Object.defineProperty` isn't supported in IE8, so it was never defined until an operation occurred.  Pushing a fix in a moment.

JakeChampion (approved):

JakeChampion:
Hi @dcherman , this is now available for testing on https://qa.polyfill.io/ 👍

PR #1158, opened 2017-03-31
Polyfill prepend for DocumentFragment

ctjhoa:
According to:
- https://dom.spec.whatwg.org/#dom-parentnode-prepend
- https://dom.spec.whatwg.org/#interface-parentnode
- https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend

Prepend is a method for ParentNode yet ParentNode is an interface
which is implemented by Document, DocumentFragment and Element.
So prepend should be polyfilled for DocumentFragment too.
Polyfill provided by MozillaDeveloperNetwork confirms it.

### DIFF
--- a/polyfills/DocumentFragment/prototype/prepend/config.json
+++ b/polyfills/DocumentFragment/prototype/prepend/config.json
@@ -1,0 +1,30 @@
+{
+	"aliases": [
+		"default-3.4",
+		"default-3.5",
+		"default-3.6",
+		"default"
+	],
+	"browsers": {
+		"android": "*",
+		"bb": "*",
+		"chrome": "<54",
+		"firefox": "<49",
+		"ios_chr": "*",
+		"ios_saf": "<10",
+		"ie": "8 - *",
+		"ie_mob": "10 - *",
+		"opera": "*",
+		"op_mini": "*",
+		"safari": "<10",
+		"firefox_mob": "<49",
+		"samsung_mob": "*"
+	},
+	"dependencies": [
+		"Document",
+		"DocumentFragment",
+		"Element",
+		"_mutation"
+	],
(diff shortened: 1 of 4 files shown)


JakeChampion (approved):

JakeChampion:
@ctjhoa Would it be possible to add a test for this change at all?

ctjhoa:
@JakeChampion I've wrote some tests. Tell me if I missed something.

JakeChampion (approved):

JakeChampion (changes requested):
Tests failing on IE7 and IE8

ctjhoa:
@JakeChampion I've done some changes, now IE8 is OK but I'm not sure about IE7 because I cannot find a IE7 javascript debugger. Can you relaunch circle ci please?

ctjhoa:
PS: This trick comes from https://github.com/WebReflection/dom4/blob/master/src/dom4.js#L286

JakeChampion:
I reran the tests on circle @ctjhoa, it looks like IE7 is still having issues with the polyfill and/or the tests.

```
Failures:
 - ie/7:
    -> Element.prototype.prepend
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.prepend
```

JakeChampion:
@ctjhoa We now have a polyfill for just DocumentFragment and separate polyfills for it's methods. Could you update this PR to add a separate polyfill for this prepend method? If you don't have the time, I can look into this.

https://github.com/Financial-Times/polyfill-service/tree/master/polyfills/DocumentFragment/prototype/append

ctjhoa:
@JakeChampion Thanks for the update. I have updated my work keep me in touch.

JakeChampion (approved):

PR #1178, opened 2017-04-24
Ensure Maps and Sets have correct constructor references, fixes #1161

### LINKED ISSUE #1161: Set polyfill doesn't work with Array.from in IE11
Hi all,
Using the included [Set polyfill](https://github.com/Financial-Times/polyfill-service/tree/master/polyfills/Set), I have the following:

`const testSet = [...new Set([1, 2, 1])];`

In Chrome and Edge (for example), the output is what I expect it to be:

screen shot 2017-04-04 at 1 38 10 pm

But in IE11 (where the polyfill kicks-in), this is what I'm shown:

screen shot 2017-04-04 at 1 38 53 pm

It looks like it's creating an object within the array. I've run the code directly from your source as well and am getting the same results. Any idea what's going on?

triblondon:
Our `Array.from` polyfill uses a check for `thing.constructor.name === 'Set'` to identify sets, but our `Set` polyfill doesn't set that property.


JakeChampion:
We should fix of Array.from polyfill as it is not always safe to rely on the constructor property of an object.

polyfills/Map/polyfill.js
```diff
@@ -141,6 +141,9 @@
 
 	Map.length = 0;
 
+	Map.prototype.constructor = Map;
```
JakeChampion: I think the constructor property is added automatically upon construction of an object.
```js
function Test() {} // => undefined
a = new Test // => Test {}
a.constructor // => function Test() {}
```

polyfills/Map/polyfill.js
```diff
@@ -141,6 +141,9 @@
...
+	Map.name = "Map";
```
JakeChampion: The value of this comes from `Function.prototype.name`, which I think we already have a polyfill for.

JakeChampion (dismissed):

JakeChampion:
@triblondon This is failing on CI, would you be able to look at this?

```
Failures:
 - ie/8:
    -> IntersectionObserver
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=IntersectionObserver
 - safari/8:
    -> WeakMap
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=WeakMap
    -> Map
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Map
    -> Set
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Set
 - android/4.4:
    -> Map
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Map
    -> Set
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Set
    -> WeakMap
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=WeakMap
```

triblondon:
Hmm, ok, will do.

kanzelm3:
@triblondon  Any ETA on this?

triblondon:
I'm not actually sure what the right solution is here.   Is it:

1. Current solution (subject to investigating the CI failures - which I'm struggling with because I not longer have access to a paid browserstack or sauce labs account)
2. Change Array.from to use a different method to test for a Set
3. Investigate possible issue with Function.prototype.name polyfill in IE11?

JakeChampion:
@triblondon, @falsandtru and myself did look to add full spec compliance to [another Array.from polyfill](https://github.com/mathiasbynens/Array.from) and hope to bring that polyfill into the polyfill-service. We can probably take another look at this effort.

polyfills/Map/polyfill.js
JakeChampion: This code already exists on line 139

polyfills/Set/polyfill.js
```diff
@@ -105,6 +105,9 @@
 
 	Set.length = 0;
 
+	Set.prototype.constructor = Set;
```
JakeChampion: This code already exists on line 104.

JakeChampion:
@triblondon I updated our test runner to report which test has failed:
```
- safari/8:
    -> Map has valid constructor
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Map
        == Map
    -> Set has valid constructor
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Set
        == Set
    -> WeakMap has valid constructor
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=WeakMap
        == WeakMap
 - android/4.4:
    -> Map has valid constructor
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Map
        == Map
    -> Set has valid constructor
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Set
        == Set
    -> WeakMap has valid constructor
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=WeakMap
        == WeakMap
```

JakeChampion (approved):

PR #1190, opened 2017-05-03
Default value of deep parameter should be false (Fixes #1189)

### LINKED ISSUE #1189: Default value of `cloneNode` argument should be false
According to MDN, the default value for the `deep` parameter was `true` in Firefox up to `v29`.

https://developer.mozilla.org/en/docs/Web/API/Node/cloneNode

JoyceBabu:
According to MDN, the default value for the deep parameter was true in Firefox up to v29.

https://developer.mozilla.org/en/docs/Web/API/Node/cloneNode

Fixes #1189

### DIFF
--- a/polyfills/Element/prototype/cloneNode/detect.js
+++ b/polyfills/Element/prototype/cloneNode/detect.js
@@ -3,7 +3,9 @@
 	test.type = "radio";
 	test.checked = true;
 	div.appendChild(test);
-	var result = test.cloneNode();
-	var result2 = div.cloneNode();
-	return !!result.checked && (result2.childNodes.length === 0);
+	var result = test.cloneNode(false), result2;
+	try {
+		result2 = div.cloneNode();
+	} catch (e) {}
+	return !!result.checked && (!result2 || result2.childNodes.length === 0);
 })()
--- a/polyfills/Element/prototype/cloneNode/detect.js
+++ b/polyfills/Element/prototype/cloneNode/detect.js
@@ -1,6 +1,8 @@
 'document' in this && "cloneNode" in document.documentElement && (function() {
-	var test = document.createElement('input');
+	var div = document.createElement('div'), test = document.createElement('input');
 	test.checked = true;
+	div.appendChild(test);
 	var result = test.cloneNode();
-	return !!result.checked;
+	var result2 = div.cloneNode();
+	return !!result.checked && (result2.childNodes.length === 0);
 })()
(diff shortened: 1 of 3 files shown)


polyfills/Element/prototype/cloneNode/detect.js
```diff
@@ -1,6 +1,8 @@
 'document' in this && "cloneNode" in document.documentElement && (function() {
-	var test = document.createElement('input');
+	var div = document.createElement('div'), test = document.createElement('input');
 	test.checked = true;
+	div.appendChild(test);
 	var result = test.cloneNode();
-	return !!result.checked;
+	var result2 = div.cloneNode();
```
JakeChampion: Where do we use result2?

JoyceBabu: Oops! It should be `return !!result.checked && (result2.childNodes.length === 0);` not `return !!result.checked && (result.childNodes.length === 0);`

JoyceBabu: Should I amend and force push or fix it is a separate commit? I don't have much experience with Github.

JakeChampion: Feel free to amend and push. No need to force push as I will squash all commits down to a single commit when I am merging :-)

JoyceBabu: I have pushed the updated code.

Just to be sure, I downloaded Firefox 28 and tested the detection code and polyfill. It is working as expected.

JoyceBabu:
The `deep` parameter is mandatory in UC Mini. I have pushed a new commit with a try/catch statement to prevent the detection code from failing on UC Mini.

With  #1189 and #1200 applied, the default set is no longer failing to load on UC Mini.

JakeChampion:
@JoyceBabu Could you post a screenshot of this working on UC Browser at all?

JakeChampion:
@JoyceBabu Could you post a screenshot of this working on UC Browser at all?

JoyceBabu:
@JakeChampion I would like to clarify that UC Mini does not allow overriding the `Element.prototype.cloneNode` function. The deep parameter is still required. The above pull request ensures that the detection code does not throw an error preventing other polyfills from loading.

I have uploaded two test pages with the current and new polyfills

http://www.joycebabu.com/test/uc-clonenode-test/current.htm
http://www.joycebabu.com/test/uc-clonenode-test/new.htm

Both the scripts have the following code at the end of the polyfill script

    document.getElementById('parent').innerHTML = 'Replaced text';

The text gets replaced when using the updated polyfill, but not with the original polyfill (with `deep` parameter fix).

screenshot_20170523-112051
screenshot_20170523-112040

JakeChampion (approved):

rowanmanning (approved):

PR #1201, opened 2017-05-11
Element polyfill fails on UCBrowser Mini (Fixes #1200)

### LINKED ISSUE #1200: Element polyfill fails on UCBrowser Mini
The `Element` polyfill fails on UCBrowser Mini because

1. `document` node cannot have more than one child
2. The polyfill uses `attachEvent`. The event `onpropertychange` proprietary to IE. Irrespective of that, using `addEventListener` will prevent it from failing completely on other browsers that require `Element` polyfill.
3. `createElement('*')` is used, which is not valid.

JoyceBabu:


JoyceBabu:
The reason I have checked for the presence of `attachEvent`, instead of using `addEventListener` is to prevent cyclic dependency `Element <=> Event`.

polyfills/Element/polyfill.js
```diff
@@ -10,12 +10,15 @@
...
+	var vbody = null;
+	if (!document.body) {
+		vbody = document.appendChild(document.createElement('body'));
+	}
+	var frame = (document.body||vbody).appendChild(document.createElement('iframe'));
 
 	// use sandboxed iframe to replicate Element functionality
 	var frameDocument = frame.contentWindow.document;
-	var prototype = Element.prototype = frameDocument.appendChild(frameDocument.createElement('*'));
+	var prototype = Element.prototype = (frameDocument.body||frameDocument.appendChild(frameDocument.createElement('body'))).appendChild(frameDocument.createElement('_'));
```
JoyceBabu, line 21: If all we need is a node, can we just use the `body` node, rather than creating a new custom element under it?

JakeChampion:
@JoyceBabu Can you post a screenshot of this working on UC Browser at all?

JoyceBabu:
Either I was stupid and added the `always` flag without the `gated` flag while testing this, or UC Mini updated their server javascript engine to add the `Element` interface.

This fix is no longer required on UC Mini.

JakeChampion:
It is always nice to hear about a browser no longer requiring a polyfill :-)

PR #1227, opened 2017-06-14
Fix Map polyfill clear method

irmantaszenkus:
Array is not a primitive so assigning the same object of type 'Array' to _keys and _values is a bug. It causes 'set' to push both keys and values to the same array.


JakeChampion (approved):

irmantaszenkus:
When could I expect for this fix to be merged and released?

JakeChampion:
@Irmiz As soon as it passes the tests on CI :-)

PR #1230, opened 2017-06-17
Removed unused dependencies (Fixes #1229)

### LINKED ISSUE #1229: setImmediate dependancy for Promise
`setImmediate` is listed as a dependency for the `Promise` polyfill. But the current Promise polyfill does not use `setImmediate`.

https://github.com/ysmood/yaku/blob/b5187ec39a90befed234ad27c980d7a36744ef49/src/yaku.core.js#L40-L42

Is it okay to remove the dependancy?

JoyceBabu:
The `Promise` polyfill is IE5+ compatible (according to project description) and does not depend on any other polyfill.

### DIFF
--- a/polyfills/Promise/config.json
+++ b/polyfills/Promise/config.json
@@ -22,11 +22,6 @@
 		"safari": "* - 7",
 		"firefox_mob": "6 - 28"
 	},
-	"dependencies": [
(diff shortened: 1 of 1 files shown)


JakeChampion:
Merged the change in #1234

PR #1239, opened 2017-07-21
Add polyfill for KeyboardEvent.prototype.key

Garbanas:
Since KeyboardEvent.which, KeyboardEvent.keyIdentifier, KeyboardEvent.keyCode are all marked as deprecated or non-standard, add a polyfill for the new (draft) KeyboardEvent.key property.

See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

### DIFF
--- a/polyfills/KeyboardEvent/prototype/key/config.json
+++ b/polyfills/KeyboardEvent/prototype/key/config.json
@@ -1,0 +1,29 @@
+{
+	"aliases": [
+	
+	],
+	"browsers": {
+		"android": "*",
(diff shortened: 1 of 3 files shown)


JakeChampion:
Hi @Garbanas, would it be possible to add some tests for this polyfill at all?

Garbanas:
Hey @JakeChampion, testing would only work in some browsers since I'd have to trigger a custom KeyboardEvent. I'll add some tests later and see how that works out.

PR #1241, opened 2017-07-21
[fix] add `append(...args)` to fragments too

WebReflection:
Specifications make `append` available for document fragments too.

It's easy to fail a feature detection via `'append' in document` assuming `append` can be then used via `document.createDocumentFragment().append(...)` too.

This change fixes the issue and avoid conflicts with other libraries.

### DIFF
--- a/polyfills/Element/prototype/append/config.json
+++ b/polyfills/Element/prototype/append/config.json
@@ -22,6 +22,7 @@
 	},
 	"dependencies": [
 		"Document",
+		"DocumentFragment",
 		"Element",
 		"_mutation"
 	],
--- a/polyfills/Element/prototype/append/polyfill.js
+++ b/polyfills/Element/prototype/append/polyfill.js
@@ -1,3 +1,3 @@
-Document.prototype.append = Element.prototype.append = function append() {
+Document.prototype.append = Element.prototype.append = DocumentFragment.prototype.append = function append() {
 	this.appendChild(_mutation(arguments));
 };


JakeChampion (approved):

JakeChampion:
Hi @WebReflection, thanks for the PR! Would you be able to add a test for this at all?

WebReflection:
Do you want me to test that in JavaScript the `Constructor.prototype.method = function () {}` works?

What would be the purpose of the test at all?

JakeChampion:
To ensure it works correctly across all the browsers we are serving the polyfill to as well as to ensure that we do not regress back to not supportinf document fragments. The tests are located in [this file](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Element/prototype/append/tests.js).

JakeChampion:
It looks like this is failing on some browsers:
```
 - ie/8:
    -> Element.prototype.append
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.append
 - ie/7:
    -> Element.prototype.append
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.append
```

WebReflection:
I don't understand.

You have `Document.prototype.append = ` inside the `Element` polyfill and zero test about it, but you want me to test that my change is covered?

Where is the append test for Document then?

The `append` is part of [ParentNode](https://dom.spec.whatwg.org/#interface-parentnode) interface and there's no such thing in this repository.

All I can do that makes sense in terms of regression is chaging this line:
https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Element/prototype/append/tests.js#L11

And use `element = document.createDocumentFragment();` instead so that [this line](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Element/prototype/append/tests.js#L14) covers `.append` for `DocumentFragment.prototype` and [the rest](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Element/prototype/append/tests.js#L30-L35) covers `.append` for `Element.prototype`.

Would that work?

If not, there's a need for a refactoring about `ParentNode`, to change tests for `Document.prototype.append` too (which I've no idea why is there in the first place) and eventually copy and paste the same done for `Document` on a `DocumentFragment` folder and re-do same `Element` test in there but I'm afraid I won't have time to do this, I've PR'd this to patch your broken polyfill.

WebReflection:
> It looks like this is failing on some browsers:

Because in those cases you need a polyfill upfront. There is no global `DocumentFragment` so you have to create one via `this.DocumentFragment = document.createDocumentFragment().constructor`.

But that means, like I'v esaid before, a whole new folder for `DocumentFragment` and I'm not familiar with your repo and unfortunately I won't have time to fix all these.

Can you help here improving the missing DocumentFragment folder? I'm sure it'll take nothing for you, then I can re-file a PR on top of that folder with the `prototype/append` logic.

In such case you can close this and please let me know when the rest has been fixed.

Thank you.

JakeChampion:
Yes, that makes perfect sense @WebReflection and I am happy to do that work.

JakeChampion (changes requested):
This needs a DocumentFragment polyfill to exist before this can be merged

JakeChampion:
@WebReflection I got it working in IE8 using the DocumentFragment polyfill. The tests for append are still failing in IE7, are you able to look into this at all?

```
- ie/7:
    -> Element.prototype.append
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.append
```
https://circleci.com/gh/Financial-Times/polyfill-service/2190

WebReflection:
The poly for DocumentFragment looks ok, but I have no idea what's the issue with IE7.

I've also no idea how to test IE7 because not even Microsoft releases VM targeting IE7.

However, I believe IE7 was broken before this PR and I wonder, if that's the case, why it's now a concern.

Out of curiosity, since you have more data than me: who on earth is still using IE7 and why would you patch it instead of letting it die?

WebReflection:
So, I am going to close this pull request and eventually create a new one.

PR #1246, opened 2017-07-29
using existing polyfills for Object(.values .entries)

caub:


.npmignore
```diff
@@ -6,3 +6,4 @@
 /node-*
 /service-process.pid
 .DS_Store
+package-lock.json
```
JakeChampion: Please could you not ignore the package lock file, this is a very important file to keep for this project as it ensures we pull in the exact versions of third party polyfills that we have tested on BrowserStack

JakeChampion (changes requested):
Need to recommit package lock file and not ignore it

JakeChampion:
@caub I'm not sure what using third party polyfills gain us for these features as the polyfill code for them is small. Could you explain the reasoning behind this please

caub:
Just was just a suggestion, this polyfill is the main one for Object.entries/values, by the person who made those proposals, it's well supported probably like the one you did, but it saves time and worry to rely on it probably

if you don't want it's ok, it makes sense too

I'm less and less convinced it's needed too (it's just a bit inconsistent how you added Object.values as a getter but not Object.entries

JakeChampion:
I'm going to close this PR as we already have polyfills for these features, thanks for taking the time to create it and sorry it didn't get merged.

PR #1248, opened 2017-07-31
Function.name is available since at least chrome 15

### LINKED ISSUE #1052: Function.name polyfill crashes on Android 4.4
Since it was included in the es6 alias it broke my website in .

The user agent includes Chrome 30 which causes it to be parsed as Chrome.

christophehurpeau:
should fix #1052


JakeChampion (approved):

p0wl:
thanks!

PR #1249, opened 2017-07-31
Remove the note about `searchParams` not being iterable

caub:
https://github.com/inexorabletash/polyfill/blob/master/polyfill.js#L5514 has .forEach, etc..


JakeChampion:
Hi @caub, would you be able to sign our CLA?

PR #1253, opened 2017-08-08
Use typeof callback === 'function' instead of instanceof

### LINKED ISSUE #1245: IE11 Array.prototype.find anonymous function is not a function TypeError
In IE11 we're seeing a type error `function(t) { return t.cardId === other.id } is not a function` when trying to call `[].find()`. But in the console `function(t) { return... } instanceof Function` is true. I'm not sure what's going on and why it would work in one place but not the other in IE11. This is an issue on all Array.prototype functions that take a callback and do a type check. 

I believe the simple solution would be to use a `typeof callback === 'function'` check instead of `instanceof Function` check.

aaronjensen:
In certain cases on IE11, instanceof Function does not return true for
functions.

Fixes #1245

### DIFF
--- a/polyfills/Array/from/polyfill.js
+++ b/polyfills/Array/from/polyfill.js
@@ -53,7 +53,7 @@
 			}
 
 			// handle maps that are not functions
-			if (1 in arguments && !(arguments[1] instanceof Function)) {
+			if (1 in arguments && typeof arguments[1] !== 'function') {
 				throw new TypeError(arguments[1] + ' is not a function');
 			}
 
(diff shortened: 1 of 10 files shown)


aaronjensen:
Thanks for merging this @JakeChampion. Do you know if polyfill.io is automatically updated? It looks like it's still serving the old version unless i'm mistaken. Thanks!

JakeChampion:
Hi @aaronjensen, thanks for the question. This PR was merged into release 3.21 which goes to production on Thursday. It is currently on our [QA service](https://qa.polyfill.io/v2/docs/) if you would like to test it out.

aaronjensen:
Thanks!

PR #1258, opened 2017-08-11
No versions of IE support toBlob

turadg:
IE11 doesn't and that's the last IE version ever.

---
This change is [Reviewable](https://reviewable.io/reviews/financial-times/polyfill-service/1258)

### DIFF
--- a/polyfills/HTMLCanvasElement/prototype/toBlob/config.json
+++ b/polyfills/HTMLCanvasElement/prototype/toBlob/config.json
@@ -6,7 +6,7 @@
 		"chrome": "<50",
 		"firefox": "<19",
 		"safari": "*",
-		"ie": "*",
+		"ie": "10-11",
 		"opera": "*"
 	},
 	"dependencies": [
--- a/polyfills/HTMLCanvasElement/prototype/toBlob/config.json
+++ b/polyfills/HTMLCanvasElement/prototype/toBlob/config.json
@@ -6,7 +6,7 @@
 		"chrome": "<50",
 		"firefox": "<19",
 		"safari": "*",
-		"ie": "10",
+		"ie": "*",
 		"opera": "*"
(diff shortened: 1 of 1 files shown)


turadg:
@JakeChampion is there anything this needs before it can be reviewed?

JakeChampion:
IE11 supports the basic usage but not the quality parameter. The polyfill supports the quality parameter, serving this to IE11 makes sense.

Thanks for spotting this @turadg!

EDIT: IE11 has it prefixed with `ms`.

JakeChampion (approved):

JakeChampion:
@turadg Can you edit the IE version range to be 10 - 11. We bundle Microsoft Edge into our IE browser detection, with the IE version range set to `*`, it will serve the polyfill to Microsoft Edge as well. We also haven't got this polyfill working in IE 7,8 or 9 which is why it only listed 10 in the version range.

JakeChampion (changes requested):
Need to change the range to 10-11 to avoid targeting MS Edge or IE versions which the polyfill doesn't yet work on

turadg:
Cool. Like that?

JakeChampion (approved):

JakeChampion:
Yes, thanks @turadg!

amarinov:
@JakeChampion @turadg I don't think Edge supports it:

image

And the polyfill **does work** for Edge:

image

JakeChampion:
@amarinov Thanks for the comment and showing that the polyfill would infact make the tests pass. It looks like we have an issue open for Edge already -- https://github.com/Financial-Times/polyfill-service/issues/1437

amarinov:
And btw it is supported in Safari 11.1 and tests are passed:

image

amarinov:
Made a PR about it: #1704

Issue #1259, opened 2017-08-15
Array.from polyfill still does not work correctly when passed a set in IE11


mcshaz:
this is a continuation on from issue #1161. 

The tests and fix provided in the fix for issue #1161 check the map or set for the 'entries' or 'values' methods respectively on the prototypes, but IE11 does not have these methods on the map and set prototypes. Similarly, it does not seem in my testing that these methods are being polyfilled in IE11.

The test passes in IE11 only because the test is bypassed - the  existence of the 'entries' and 'values' methods are checked before running the test and bypassed if not present. 

2 potential options I can see:

- check for the forEach method on the prototypes of map and set, and then build the array using this rather than looking for an iterator.
- polyfill in the entries and values methods to map and set prototypes prior to running Array.from

I can submit a pull request for failing tests to look at the issue more closely.

Thanks

mcshaz:
some approaches to testing and a simple fix are here https://github.com/Financial-Times/polyfill-service/compare/master...mcshaz:master. 2 commits - the first with the tests that fail in IE11 and the second with the code changed to use forEach rather than entries/values for map & set instances.

mcshaz:
I think this issue should probably be merged with #1209  although I don't believe the Set polyfill is used because IE11 has a Set constructor, but the object produced doesn't adhere to the iterable pattern - the code in the pull request related to this solution is a workaround using IE11s native set which does have a foreach method which does iterate through undefined values.

JakeChampion:
@mcshaz As the PR related to this issue ( #1260 ) has been merged, can we close this issue?

PR #1260, opened 2017-08-16
Array.from working with IE11 native Set - issue #1259

mcshaz:


polyfills/Array/from/tests.js
```diff
@@ -49,42 +49,49 @@ describe('returns an array with', function () {
...
 
-				// Test map iterable
-				proclaim.deepEqual(Array.from(mapIterator), [2,4]);
+				if (typeof mapIterator.next === 'function') {
 
-				it('can convert from Map', function () {
-					proclaim.deepEqual(Array.from(map), [[1,2],[3,4]]);
-				});
+					// Test map iterable
+					proclaim.deepEqual(Array.from(mapIterator), [2,4]);
```
JakeChampion: How come this test is now using the map iterator instead of the map itself? Could we have a comment which explains the reasoning please

mcshaz: The map and the map iterator are both tested as before (and a 3rd test is added for the "map" function) - this is simply the way the dif makes it look. Prior to the tests (and this is exactly the same for sets) the test only tested using the 'map' variable if the Map.prototype contained a 'entries' method (although it erroneously actually used the 'values' method i.e. NOT the method whos existance was tested for). Now _all_ browsers which have a window.Map object are tested against the map itself, and only those who can produce a map iterator via the values function are tested against the map iterator.

Before changes to tests:

	if ('Map' in window && 'entries' in Map.prototype) {

		map = new Map();
		map.set(1,2);
		map.set(3,4);
		mapIterator = map.values();

		if (typeof mapIterator.next === 'function') {

			// Test map iterable
			proclaim.deepEqual(Array.from(mapIterator), [2,4]);

			it('can convert from Map', function () {
				proclaim.deepEqual(Array.from(map), [[1,2],[3,4]]);
			});
		}
	}

after changes to test:

	if ('Map' in window) {
		map = new Map();
		map.set(1,2);
		map.set(3,4);
		if (map.values) {
			mapIterator = map.values();

			if (typeof mapIterator.next === 'function') {

				// Test map iterable
				proclaim.deepEqual(Array.from(mapIterator), [2,4]);
			}
		}
		
		it('can convert from Map', function () {
			proclaim.deepEqual(Array.from(map), [[1,2],[3,4]]);
			proclaim.deepEqual(Array.from(map, returnArgs), [[[1,2],0],[[3,4],1]]);
		});
	}

mcshaz:
I have neatened things a little further - hopefully the changes are now a little more obvious.

polyfills/Array/from/polyfill.js
```diff
@@ -12,22 +11,18 @@
...
-				) {
+				if (typeof iterableResponse.done === 'boolean') {
 					if (iterableResponse.done === true) {
 						done = true;
 						break;
-
-					// handle the case where the done value is not Boolean
-					} else if (iterableResponse.done !== false) {
+					}
+					//was using hasownProperty but changed as 'value' property might be inherited through prototype chain and could still be a valid iterable response
```
JakeChampion, line 25: Are you sure about this? I think the Iterator Result Object does not have a prototype and always has a `done` property but may not have a `value` property -- https://tc39.github.io/ecma262/#table-55

JakeChampion:
@mcshaz The tests for this change are failing on a few browsers:
```
Failures:
	http://127.0.0.1:3000/test/tests?mode=targeted&feature=Array.from
 - ie/10:
    -> Array.from passes a feature detect
       false == true
    -> Array.from returns an array with arrays
       'Map' is undefined
    -> Array.from returns an array with fills holes in arrays
       'Map' is undefined
    -> Array.from returns an array with objects
       'Map' is undefined
    -> Array.from returns an array with strings
       'Map' is undefined
    -> Array.from returns an array with numbers
       'Map' is undefined
    -> Array.from returns an array with regular expressions
       'Map' is undefined
    -> Array.from returns an array with objects with in-range lengths
       'Map' is undefined
    -> Array.from returns an array with objects with out-of-range lengths
       'Map' is undefined
    -> Array.from returns an array with mapping functions
       'Map' is undefined
    -> Array.from returns an array with this as an object
       'Map' is undefined
    -> Array.from returns an array with this as 42
       'Map' is undefined
    -> Array.from returns an array with this as false
       'Map' is undefined
 - ie/9:
    -> Array.from passes a feature detect
       false == true
    -> Array.from returns an array with arrays
       'Map' is undefined
    -> Array.from returns an array with fills holes in arrays
       'Map' is undefined
    -> Array.from returns an array with objects
       'Map' is undefined
    -> Array.from returns an array with strings
       'Map' is undefined
    -> Array.from returns an array with numbers
       'Map' is undefined
    -> Array.from returns an array with regular expressions
       'Map' is undefined
    -> Array.from returns an array with objects with in-range lengths
       'Map' is undefined
    -> Array.from returns an array with objects with out-of-range lengths
       'Map' is undefined
    -> Array.from returns an array with mapping functions
       'Map' is undefined
    -> Array.from returns an array with this as an object
       'Map' is undefined
    -> Array.from returns an array with this as 42
       'Map' is undefined
    -> Array.from returns an array with this as false
       'Map' is undefined
 - ie/8:
    -> Array.from passes a feature detect
       false == true
    -> Array.from returns an array with arrays
       'Map' is undefined
    -> Array.from returns an array with fills holes in arrays
       'Map' is undefined
    -> Array.from returns an array with objects
       'Map' is undefined
    -> Array.from returns an array with strings
       'Map' is undefined
    -> Array.from returns an array with numbers
       'Map' is undefined
    -> Array.from returns an array with regular expressions
       'Map' is undefined
    -> Array.from returns an array with objects with in-range lengths
       'Map' is undefined
    -> Array.from returns an array with objects with out-of-range lengths
       'Map' is undefined
    -> Array.from returns an array with mapping functions
       'Map' is undefined
    -> Array.from returns an array with this as an object
       'Map' is undefined
    -> Array.from returns an array with this as 42
       'Map' is undefined
    -> Array.from returns an array with this as false
       'Map' is undefined
 - ie/7:
    -> Array.from passes a feature detect
       false == true
    -> Array.from returns an array with arrays
       'Map' is undefined
    -> Array.from returns an array with fills holes in arrays
       'Map' is undefined
    -> Array.from returns an array with objects
       'Map' is undefined
    -> Array.from returns an array with strings
       'Map' is undefined
    -> Array.from returns an array with numbers
       'Map' is undefined
    -> Array.from returns an array with regular expressions
       'Map' is undefined
    -> Array.from returns an array with objects with in-range lengths
       'Map' is undefined
    -> Array.from returns an array with objects with out-of-range lengths
       'Map' is undefined
    -> Array.from returns an array with mapping functions
       'Map' is undefined
    -> Array.from returns an array with this as an object
       'Map' is undefined
    -> Array.from returns an array with this as 42
       'Map' is undefined
    -> Array.from returns an array with this as false
       'Map' is undefined
 - android/4.4:
    -> Array.from passes a feature detect
       false == true
    -> Array.from returns an array with arrays
       Map is not defined
    -> Array.from returns an array with fills holes in arrays
       Map is not defined
    -> Array.from returns an array with objects
       Map is not defined
    -> Array.from returns an array with strings
       Map is not defined
    -> Array.from returns an array with numbers
       Map is not defined
    -> Array.from returns an array with regular expressions
       Map is not defined
    -> Array.from returns an array with objects with in-range lengths
       Map is not defined
    -> Array.from returns an array with objects with out-of-range lengths
       Map is not defined
    -> Array.from returns an array with mapping functions
       Map is not defined
    -> Array.from returns an array with this as an object
       Map is not defined
    -> Array.from returns an array with this as 42
       Map is not defined
    -> Array.from returns an array with this as false
       Map is not defined
```

mcshaz:
Thank you for running the tests @JakeChampion I think it should be working for these browsers now, although in truth I only have emulation to rely on rather than virtual machines. 

On the comment about the iterator obtained by calling next() - the code only looks for a value on the iterator if `done === false`, if `done ===true` the function stops iterating, and if done is non-boolean the whole function returns false (does not implement iterator). 

As for the link - I am no expert on the spec, but I am unsure whether I can interpret "The IteratorResult interface includes the properties listed in Table 58:" as a _direct_ property or whether an _inherited_ property is OK - you have far more experience with the spec than I and I don't feel at all strongly whether to test 'done' in iterator or iterator.hasOwnProperty('done')
I think it should be working for these browsers now, although in truth I only have emulation to rely on rather than virtual machines. 

As for the link - I am no expert on the spec, but I am unsure whether I can interpret "The IteratorResult interface includes the properties listed in Table 58:" as a _direct_ property or whether an _inherited_ property is OK - you have far more experience with the spec than I and I don't feel at all strongly whether to test `'done' in iterator` or `iterator.hasOwnProperty('done')`

JakeChampion (approved):

PR #1262, opened 2017-08-16
String.prototype.padEnd.length should be 1

JakeChampion:
2nd argument is optional which means it should not count towards function's length property.

Our live test suite shows this test failing in latest Chrome Stable -- https://polyfill.io/test/tests?mode=targeted&grep=String%5C.prototype%5C.padEnd


alicebartlett:
Looks good to me :+1:

PR #1270, opened 2017-08-18
For `Element.prototype.before` and `Element.prototype.after` - Let viableNextSibling be context object’s first following sibling not in nodes, and null otherwise.

caub:
https://dom.spec.whatwg.org/#dom-childnode-after point 3.

possibly .before and .replaceWith too

### DIFF
--- a/polyfills/Element/prototype/after/config.json
+++ b/polyfills/Element/prototype/after/config.json
@@ -23,6 +23,7 @@
 	"dependencies": [
 		"Document",
 		"Element",
+		"Array.prototype.indexOf",
 		"_mutation"
 	],
 	"spec": "http://dom.spec.whatwg.org/#dom-childnode-after",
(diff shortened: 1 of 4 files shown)


JakeChampion:
@caub Thanks for this! Would you be able to add tests to cover this at all?

JakeChampion:
@caub It looks like this PR has some tests failing on CI:
```
- ie/8:
    -> Element.prototype.after can insert elements after itself
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.after
       Object doesn't support this property or method
 - ie/8:
    -> Element.prototype.before can insert elements before itself
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.before
       Object doesn't support this property or method
 - ie/7:
    -> Element.prototype.after can insert elements after itself
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.after
       Object doesn't support this property or method
 - ie/7:
    -> Element.prototype.before can insert elements before itself
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.before
       Object doesn't support this property or method
```

caub:
ok, added Array.prototype.indexOf in dependencies

for replaceWith I think there are more checks to do, will see

JakeChampion (approved):

PR #1269, opened 2017-08-18
non-node are accepted and converted as Text

caub:

PR #1272, opened 2017-08-22
[fix] add `append(...args)` to fragments too

WebReflection:
Specifications make append available for document fragments too.

It's easy to fail a feature detection via 'append' in document
assuming append can be then used via document.createDocumentFragment().append(...) too.

This change fixes the issue and avoid conflicts with other libraries.

### DIFF
--- a/polyfills/DocumentFragment/config.json
+++ b/polyfills/DocumentFragment/config.json
@@ -5,7 +5,7 @@
 		"firefox": "<4",
 		"ios_chr": "<4",
 		"ios_saf": "<4",
-		"ie": "8-9",
+		"ie": "8 - 9",
 		"ie_mob": "<9",
 		"opera": "<11",
 		"op_mini": "<11",
--- a/polyfills/DocumentFragment/config.json
+++ b/polyfills/DocumentFragment/config.json
@@ -5,7 +5,7 @@
 		"firefox": "<4",
 		"ios_chr": "<4",
 		"ios_saf": "<4",
-		"ie": "<9",
+		"ie": "8-9",
 		"ie_mob": "<9",
 		"opera": "<11",
 		"op_mini": "<11",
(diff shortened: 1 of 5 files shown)


caub:
isn't it redundant with https://github.com/Financial-Times/polyfill-service/pull/1269 ?

WebReflection:
This is DocumentFragment append, which doesn't inherit from Element.

JakeChampion:
I made the tests output some more information:
```
Failures:
 - ie/7:
    -> DocumentFragment.prototype.append passes a feature detect
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=DocumentFragment.prototype.append
       'DocumentFragment.prototype' is null or not an object
 - ie/7:
    -> DocumentFragment.prototype.append has correct instance
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=DocumentFragment.prototype.append
       undefined instanceof 
function Function() {
    [native code]
}

 - ie/7:
    -> DocumentFragment.prototype.append has correct name
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=DocumentFragment.prototype.append
       Function expected
 - ie/7:
    -> DocumentFragment.prototype.append has correct argument length
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=DocumentFragment.prototype.append
       'fragment.append.length' is null or not an object
 - ie/7:
    -> DocumentFragment.prototype.append can append elements
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=DocumentFragment.prototype.append
       Object doesn't support this property or method

```

JakeChampion (approved):

JakeChampion:
@WebReflection Here are some stats on a months worth of origin traffic for IE 7/8/9. In that month we had ~5 million hits to origin and ~2.5% of those were IE 7/8/9.

screen shot 2017-09-01 at 18 04 00

WebReflection:
2.5 but only 1/5 of that looks like IE7 ... I don't think the world should put any effort to support less than IE9 in 2017, nothing would work there anyway, unless you write code like it's 1999 and if you do, I'm sorry to hear that, no polyfill can save you.

Thanks for sharing the graph though, interesting and terrifying at the same time.

PR #1274, opened 2017-08-24
Polyfill Array.prototype.@@iterator on Safari 9

matthieusieben:
Got several error reports in our tracker from iOS users with the following UA.
```
Mozilla/5.0 (iPhone; CPU iPhone OS 9_2_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13D15 Safari/601.1
```
http://caniuse.com/#search=iterator

### DIFF
--- a/polyfills/Array/prototype/@@iterator/config.json
+++ b/polyfills/Array/prototype/@@iterator/config.json
@@ -8,9 +8,9 @@
 		"ie_mob": "9 - 12",
 		"firefox": "<38",
 		"chrome": "<49",
-		"safari": "<9",
+		"safari": "<10",
 		"android": "<5.1",
-		"ios_saf": "<9",
+		"ios_saf": "<10",
 		"opera": "<25",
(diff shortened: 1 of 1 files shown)


JakeChampion:
@matthieusieben [Kangax's ES6 table](https://kangax.github.io/compat-table/es6/#test-Array.prototype_methods_Array.prototype[Symbol.iterator]_a_href=_https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/@@iterator_title=_MDN_documentation_img_src=_../mdn.png_alt=_MDN_(Mozilla_Development_Network)_logo_width=_15_height=_13_/_/a_nbsp;) shows that iOS 9 does support `Array.prototype[Symbol.iterator]`.

JakeChampion:
I've tested Array iterator support on iOS 9 and it works natively

PR #1277, opened 2017-08-29
To avoid an error where console is undefined, check that console exists before checking a console method exists

JakeChampion:
I've updated all the console method detection scripts to first check that console actually exists. If console does not exist the detection script will return false.


rowanmanning (approved):
Looks good 👍

PR #1282, opened 2017-09-01
Set the age header to 0 before sending the request back to the browser

### LINKED ISSUE #1276: Polyfill always fetched from source with 'age' header value higher than 'max-age' header value.
When requesting `https://cdn.polyfill.io/v2/polyfill.min.js?features=es5,es6,es7,default-3.6,Object.values&flags=gated&version=latest&unknown=ignore` I see the following response headers:

```
accept-ranges:bytes
access-control-allow-origin:*
age:1099341
cache-control:public, s-maxage=31536000, max-age=604800, stale-while-revalidate=604800, stale-if-error=604800
content-encoding:br
content-length:978
content-type:application/javascript;charset=utf-8
date:Fri, 25 Aug 2017 15:37:54 GMT
server:Cowboy
status:200
strict-transport-security:max-age=31536000; includeSubdomains; preload
timing-allow-origin:*
vary:User-Agent
vary:Accept-Encoding
via:1.1 vegur
via:1.1 varnish
x-cache:HIT
x-cache-hits:1
x-content-type-options:nosniff
x-frame-options:sameorigin
x-served-by:cache-ams4135-AMS
x-timer:S1503675475.984820,VS0,VE0
x-xss-protection:1; mode=block
```

Because the `age` value is higher than the `max-age` value the response is never stored in browser cache and always fetched from the CDN.

Would it be possible to omit the `age` header?

---

update

Have done some more research on this. The 'age' header seems to be set by fastly. This seems an incorrect configuration to me as the current cache mechanic for polyfill.io is to store indefinitely in CDN and purge on update. Serving "stale", as indicated by the response headers can't be intentional.

JakeChampion:
Fixes #1276 

This should be safe as it is after the `restart` call inside of vcl_deliver.

### DIFF
--- a/vcl/main.vcl
+++ b/vcl/main.vcl
@@ -122,6 +122,8 @@ sub vcl_deliver {
 		set resp.http.Debug-Host = req.http.Host;
 		set resp.http.Debug-X-Original-Host = req.http.X-Original-Host;
 	}
+	
+	unset resp.http.Age;
 
 	return(deliver);
 }
--- a/vcl/main.vcl
+++ b/vcl/main.vcl
@@ -123,7 +123,7 @@ sub vcl_deliver {
 		set resp.http.Debug-X-Original-Host = req.http.X-Original-Host;
 	}
 	
-	unset resp.http.Age;
+	set resp.http.Age = 0;
 
 	return(deliver);
 }


sjparkinson:
Related https://community.fastly.com/t/cache-control-max-age-and-age-headers/1018.

sjparkinson:
Reading https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html, I don't think it's wise to remove the `Age` header.

sjparkinson (approved):

PR #1286, opened 2017-09-06
Move to CircleCI 2

JakeChampion:


.circleci/do-exclusively.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
JakeChampion, line 1: This ensures that only one build is running this task at a time. We need this because we use all allowed BrowserStack browsers when running tests against one PR, if we are running tests against two PRs BrowserStack will error saying we are using too many browsers concurrently.

PR #1290, opened 2017-09-07
revert 90cb1ab7e6cf118692e9046700aa4d60de7a40f5

JakeChampion:
This is no longer used by FT.com


triblondon:
Doesn't this make it super hard for anyone to run their own polyfill service?

JakeChampion:
@triblondon I've added back the host override options, `process.env.HOST`, falling back to host header of the request or falling back to `https://polyfill.io` if that doesn't exist. Do you think that is okay?

triblondon (approved):

PR #1293, opened 2017-09-08
Support Objects which have Symbol.iterator defined

### LINKED ISSUE #1209: Polyfill fails in IE11 when using Array.from of a Map
Normally, applying `Array.from` to a JavaScript `Map` results in an array of `[key, value]` tuples. However, when using polyfill.io with default features in IE11, it results in an array of `undefined`.

Here is a repro:

```javascript
var fooMap = new Map()
  .set("key1", "value1")
  .set("key2", "value2");

var fooKeys = Array.from(fooMap.keys());
console.log("first key", fooKeys[0]); // works fine

var fooPairs = Array.from(fooMap);
console.log("first pair", fooPairs[0]); // in IE11, results in an array of undefined instead of key/value tuples
```

Please see this [CodePen](https://codepen.io/devuxer/pen/PmgYjY?editors=1111) for a live demo of the problem.

### LINKED ISSUE #1251: Array.from does not try to use iterator protocol
In Chrome (without Polyfill), the following snippet produces the expected result:
```js
var o = {};
o[Symbol.iterator] = function() {
  let i = 0;
  return ({
    next: function() {
      i++;
      return { done: i > 5, value: i};
    }
  });
};
Array.from(o);
```

```js
(5) [1, 2, 3, 4, 5]
```

The same in IE11 with https://cdn.polyfill.io/v2/polyfill.min.js?features=es6 loaded returns an empty array:
```
[object Array]   []
   length        0
```

I discovered this actually with MobX observables, for which Array.from also produces an empty array.

JakeChampion:
Fixes #1251 & fixes #1209

I need to get this working in IE8 and IE7


polyfills/Array/from/config.json
```diff
@@ -24,7 +24,12 @@
 		"bb": "10 - *"
 	},
 	"dependencies": [
-		"Object.defineProperty"
+		"_ArrayIterator",
```
JakeChampion:
```
Array.from
├─┬_ArrayIterator
│ ├─┬ _Iterator
│ │ ├── Symbol
│ │ │ ├── Object.keys -- *Doesn't work on IE7 or IE8*
│ │ │ └── Object.freeze -- *We don't have a polyfill for this*
```

polyfills/Array/from/polyfill.js
```diff
@@ -1,106 +1,306 @@
...
+
 	//same as instanceOf, but works through iframes
 	function appWideInstanceOf(obj, typeName){
 		return Object.prototype.toString.call(obj) === '[object ' + typeName + ']'
 	}
-	function parseIterable (arraylike) {
-		//worth considering the performance bypass in the line below
-		//if (appWideInstanceOf(arraylike,'Array')){ return arraylike.slice(); }
+
+	function ToInteger(value) {
```
rowanmanning, line 10: Why's `To` capitalized?

rowanmanning (approved):
Code looks good from a quick run through, just one small comment

75lb:
Is this live yet? When i request `Array.from` in IE11 like so:

```html
<script src="https://cdn.polyfill.io/v2/polyfill.js?features=Array.from"></script>
```

The polyfill code returned looks old, i.e. it doesn't support default iterators (`@@iterator`) as mentioned in #1251.

```js
/* Polyfill service v3.21.3
 * For detailed credits and licence information see https://github.com/financial-times/polyfill-service.
 * 
 * UA detected: ie/11.0.0
 * Features requested: Array.from
 * 
 * - Array.from, License: CC0 */

(function(undefined) {

// Array.from
// Wrapped in IIFE to prevent leaking to global scope.
(function () {
	//same as instanceOf, but works through iframes
	function appWideInstanceOf(obj, typeName){
		return Object.prototype.toString.call(obj) === '[object ' + typeName + ']'
	}
	function parseIterable (arraylike) {
		//worth considering the performance bypass in the line below
		//if (appWideInstanceOf(arraylike,'Array')){ return arraylike.slice(); }
		var done = false;
		var iterableResponse;
		var tempArray = [];

		// if the iterable doesn't have next;
		// it is an iterable if 'next' is a function but it has not been defined on
		// the object itself.
		if (typeof arraylike.next === 'function') {
			while (!done) {
				iterableResponse = arraylike.next();
				if (typeof iterableResponse.done === 'boolean') {
					if (iterableResponse.done === true) {
						done = true;
						break;
					}
					//was using hasownProperty but changed as 'value' property might be inherited through prototype chain and could still be a valid iterable response
					if ('value' in iterableResponse){ 
						tempArray.push(iterableResponse.value);
					} else {
						break;
					}
				} else {
					// it does not conform to the iterable pattern
					break;
				}
			}
		}

		if (done) {
			return tempArray;
		} else {
			// something went wrong return false;
			return false;
		}
	}
	
	function iterateForEach(arraylike, asKeyValArrays) {
		if (typeof arraylike.forEach !== 'function') {
			return false;
		}
		var tempArray = [];
		var addEl = asKeyValArrays
			? function (val, key) { tempArray.push([key, val]); } 
			: function (val) { tempArray.push(val); };
		arraylike.forEach(addEl);
		return tempArray;
	}

	Object.defineProperty(Array, 'from', {
		configurable: true,
		value: function from(source) {
			// handle non-objects
			if (source === undefined || source === null) {
				throw new TypeError(source + ' is not an object');
			}

			// handle maps that are not functions
			if (1 in arguments && typeof arguments[1] !== 'function') {
				throw new TypeError(arguments[1] + ' is not a function');
			}
			
			var arraylike = typeof source === 'string' ? source.split('') : Object(source);
			var map = arguments[1];
			var scope = arguments[2];
			var array = [];
			var index = -1;
			var length = Math.min(Math.max(Number(arraylike.length) || 0, 0), 9007199254740991);
			var value;
			// variables for rebuilding array from iterator
			var arrayFromIterable = parseIterable(arraylike);

			//if it is a Map or a Set then handle them appropriately
			if (!arrayFromIterable) {
				if (appWideInstanceOf(arraylike,'Map')) {
					arrayFromIterable = iterateForEach(arraylike, true);
				} else if (appWideInstanceOf(arraylike,'Set')) {
					arrayFromIterable = iterateForEach(arraylike);
				}
			}

			if (arrayFromIterable) {
				arraylike = arrayFromIterable;
				length = arraylike.length;
			}

			while (++index < length) {
					value = arraylike[index];

					array[index] = map ? map.call(scope, value, index) : value;
			}

			array.length = length;

			return array;
		},
		writable: true
	});
}());
})
.call('object' === typeof window && window || 'object' === typeof self && self || 'object' === typeof global && global || {});
```

JakeChampion:
3.22 is not yet live, please follow the [twitter account] (https://twitter.com/polyfillio) or this repo for release announcements 👍

75lb:
will do, thanks 👍

devuxer:
FYI, this did *not* fix #1209.

PR #1294, opened 2017-09-08
Opera Version Updates

fooloomanzoo:
- adding missing Versions for Opera according to chromium-versions
- correcting wrong Versions of Opera


JakeChampion:
@fooloomanzoo Have you tested those browser versions? Ideally we would like to see a screenshot of the tests passing on the minimum versions of the browsers you have set to serve the polyfills for.

fooloomanzoo:
@JakeChampion Since I personally doesn't have a premium plan on any online browser test suite provider, this is hard to do for me to test for all older releases. I made only the very obvious changes. Opera implements `chromium` and `webkit` directly and without changes by version number, so that a feature that is implemented  for example in `chromium 54` is implemented in Opera 41 according to their [release history](https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Opera_2017). Some of the noted version numbers in the polyfill configurations were also larger than Opera's latest version number. These are just very obvious changes.

JakeChampion:
@fooloomanzoo If opera is exactly the same as chrome, we can alias it which would mean we know longer need to keep updating config files for opera

christophehurpeau:
Since opera 15, opera is based on chromium and blink: http://www.opera.com/docs/changelogs/unified/1500/

fooloomanzoo:
@JakeChampion Opera definitely references in its UserAgent the chromium version it is using.

fooloomanzoo:
Or in other `chromium` based browser like **Vivaldi**

image

christophehurpeau:
also usefull: https://dev.opera.com/blog/opera-user-agent-strings-opera-15-and-beyond/

JakeChampion:
@fooloomanzoo @christophehurpeau I've made a PR to alias many versions of opera to their chrome equivalent. What this means is we would not need to target opera explicitly  https://github.com/Financial-Times/polyfill-service/pull/1296

fooloomanzoo:
@JakeChampion Would it be necessary to remove or to update the version numbers in the Polyfill configurations, do they superimpose the alias version numbers?

JakeChampion:
@fooloomanzoo It is not necessary to do anything with the configurations after that PR is merged :-)

Issue #1156, opened 2017-03-30
Regenerator runtime stop working using _ArrayIterator polyfill


forceuser:
After this PR https://github.com/facebook/regenerator/pull/264 regenerator-runtiume throws error
`Uncaught TypeError: Cannot assign to read only property '__symbol:toStringTag0.42153681488707667' of [object Object]` using _ArrayIterator polyfill and Symbol polyfill

regenerator mistakenly determines the prototype of the iterator as a native

code from regenerator-runtime/runtime.js version 0.9.7

```js
var getProto = Object.getPrototypeOf;
var NativeIteratorPrototype = getProto && getProto(getProto(values([])));
if (NativeIteratorPrototype &&
      NativeIteratorPrototype !== Op &&
      hasOwn.call(NativeIteratorPrototype, iteratorSymbol)) {
    // This environment has a native %IteratorPrototype%; use it instead
    // of the polyfill.
    IteratorPrototype = NativeIteratorPrototype;
}
```

tested on chrome build 33.0.1739.0 (240917) linux
polyfill request: https://cdn.polyfill.io/v2/polyfill.js?features=default,es6,es7,caniuse:console-basic&flags=gated&unknown=polyfill

i decided to start issue here but maby it is better to submit it to regenerator repo?
for now i rollback to regenerator-runtime v0.9.5

JakeChampion:
Hi @forceuser & @christophehurpeau, do you have some code I can run to reproduce the issue at all?

christophehurpeau:
@JakeChampion here a gist 
https://gist.github.com/christophehurpeau/dae28d1e6a8c6babb7cf5458649eed69

you can try with IE9 for example using rawgit :
https://rawgit.com/christophehurpeau/dae28d1e6a8c6babb7cf5458649eed69/raw/f5a507050099cd3c0f64b27dbae06f38e8e4e42d/polyfillio-regenerator-runtime.html

<img width="492" alt="screen shot 2017-09-07 at 10 38 33" src="https://user-images.githubusercontent.com/302891/30153993-b880fa14-93b8-11e7-870c-4e4daa1d036c.png">

christophehurpeau:
@JakeChampion here a gist 
https://gist.github.com/christophehurpeau/dae28d1e6a8c6babb7cf5458649eed69

you can try using rawgit :
[...]

PR #1295, opened 2017-09-08
Set Symbol.toStringTag to be writable to make this polyfill spec compliant

JakeChampion:
and to work with generator polyfills such the the regenerator library by Facebook #1156


christophehurpeau:
@JakeChampion do you know when this will be testable on the QA ?

JakeChampion:
If all goes well, Monday 👍

JakeChampion:
This is live on https://polyfill.io

PR #1298, opened 2017-09-11
Upgrade to newest version of polyfill

dotbear:
Should solve a lot of our bugsnags


dotbear:
My bad - totally blanked this was the wrong repo :snooze:

PR #1300, opened 2017-09-11
Fix: Do not call callback for Map.prototype.forEach if no map entries exist

### LINKED ISSUE #1299: Map.prototype.forEach fires the callback for an empty Map
This bug in the polyfill impacts React 16 users, and we are shipping React 16 very soon.

Specifically, `Map.prototype.forEach` [attempts to skip deleted items](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Map/polyfill.js#L24) but then [always calls the callback](https://github.com/Financial-Times/polyfill-service/blob/c14494506bc2c95c70195d54d77110995b6cdf93/polyfills/Map/polyfill.js#L25). The problem is that it may be the case that *all* items were deleted, in which case it should have skipped the callback. In other words, [this check](https://github.com/Financial-Times/polyfill-service/blob/c14494506bc2c95c70195d54d77110995b6cdf93/polyfills/Map/polyfill.js#L22) should be done just before calling the callback as well.

Specific reproducing case:

```js
var x = new Map();
x.set(42, 'hi');
x.delete(42);
x.forEach(function() {
  throw new Error('Should never get here!');
});
```

This breaks with the provided `Map` polyfill but works with the native implementation.

JakeChampion:
Fixes https://github.com/Financial-Times/polyfill-service/issues/1299

### DIFF
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -19,7 +19,7 @@
 		var done = false;
 		return {
 			next: function() {
-				if (nextIdx === mapInst._keys.length) done = true;
+				if (!mapInst.size || nextIdx === mapInst._keys.length) done = true;
 				if (!done) {
 					while (mapInst._keys[nextIdx] === undefMarker) nextIdx++;
 					return {value: getter.call(mapInst, nextIdx++), done: false};
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -19,9 +19,20 @@
 		var done = false;
 		return {
 			next: function() {
-				if (!mapInst.size || nextIdx === mapInst._keys.length) done = true;
+				if (!mapInst.size || nextIdx === mapInst._keys.length) {
+					done = true;
+				}
 				if (!done) {
-					while (mapInst._keys[nextIdx] === undefMarker) nextIdx++;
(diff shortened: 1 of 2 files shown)


gaearon:
Does this work for case like

```js
x.set(42, 'hi')
x.set(43, 'bye')
x.delete(43)
```

(edit: typo)

?

My intuition is it would have the same problem, calling the callback extra time at the very end because of the deleted item. `size` wouldn't be `0` but it would still fail.

JakeChampion:
@gaearon That should work already but just in case I have added a test with your code.

gaearon:
Apologies for dumb question but could you help me understand why it also works? Since the new check is for `.size` I would expect the same bug to manifest itself if the last item is marked with undef sentinel, but the map is not empty.

notlee (approved):

gaearon:
I think I understand now. If it's the last one, you decrement `size`. If it's not, then `next()` will skip the missing items next time.

JakeChampion:
@gaearon There was still a bug in there, thanks for getting me to look at it even more thoroughly 👍

JakeChampion:
<details><summary>
Screenshots of tests passing in IE 7 and IE 8:</summary>

Screenshot of tests passing in IE7

Screenshot of tests passing in IE8
</details>

PR #1328, opened 2017-09-14
Reduce variance in QA/prod code paths

triblondon:
This change reduces code repetition and simplifies the logic, but  most importantly ensures that:

* Both QA and prod VCL envs will run every line of code
* You can access prod origin env via QA VCL env

This makes testing of VCL prior to deploy much more robust.


vcl/main.vcl
```diff
@@ -37,45 +48,19 @@ sub vcl_recv {
...
-				set req.backend = origami_polyfill_service_eu;
-				set req.http.Host = "ft-polyfill-service.herokuapp.com";
-
-				if (!req.backend.healthy) {
-					set req.backend = origami_polyfill_service_us;
-					set req.http.Host = "ft-polyfill-service-us.herokuapp.com";
-				}
-			}
+		# Set origin geography - US servers or EU servers
+		set var.geo = (client.geo.continent_code ~ "(NA|SA|OC|AS)", "us", "eu");
```
JakeChampion: This looks to be missing the `if` before the statement

vcl/main.vcl
```diff
@@ -37,45 +48,19 @@ sub vcl_recv {
...
+		set req.backend = if (var.geo == "us", origami_polyfill_service_us, origami_polyfill_service_eu);
+		
+		# Swap to the other grography if the primary one is down
+		if (!req.backend.healthy) {
+			set var.geo = if (var.geo == "us", "eu", "us");
+			set req.backend = if (var.geo == "us", origami_polyfill_service_us, origami_polyfill_service_eu);
 		}
+		
+		# Set origin environment - by default match VCL environment, but allow override via header for testing
+		set var.env = if (req.http.X-Origin-Env, req.http.X-Origin-Env, if(req.http.Host == "qa.polyfill.io", "qa", "prod");
```
JakeChampion: What do you think about ensuring X-Origin-Env is either `prod` or `qa` before using it?

triblondon: Could do.  It's safe not to since it's an own-goal if you set it to something invalid.  It only affects you.

vcl/main.vcl
```diff
@@ -95,6 +80,22 @@ sub vcl_recv {
...
+sub vcl_fetch {
+
+	# Deliver stale if possible when unexpected errors are received from origin
+	# This includes 404s because Heroku will deliver a 404 if you hit the 'no app
+	# found' page.
+	if ((beresp.status >= 500 && beresp.status < 600) ||  beresp.status == 404) {
+		if (stale.exists) {
+			return(deliver_stale);
+		}
+	}
```
triblondon, line 103: This is the approach we advocate in https://docs.fastly.com/guides/performance-tuning/serving-stale-content.  Another approach might be to change `beresp.status` from 404 to 503, which should trigger varnish's default stale-if-error logic.  You could experiment with that if you want.

sjparkinson (approved):
Very neat, nice to see an example of using `table`.

PR #1331, opened 2017-09-18
Add privacy-policy and T&Cs pages

JakeChampion:
The [privacy policy](https://polyfill.io/v2/docs/privacy-policy) outlines what data we collect about requests to https://polyfill.io and what we do with that data. 

The [terms of service](https://polyfill.io/v2/docs/terms) document describes what you can expect from us when you use https://polyfill.io on your site, and the actions you as a user might take that would cause us to revoke your access to https://polyfill.io.

### DIFF
--- a/docs/nav.html
+++ b/docs/nav.html
@@ -31,5 +31,11 @@
 				</ul>
 				{{/ifPrefix}}
 			</li>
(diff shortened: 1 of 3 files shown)


alicebartlett:
:+1:

PR #1345, opened 2017-09-27
Fail the feature detect for localstorage if we are in safari private browser

JakeChampion:


polyfills/localStorage/detect.js
```diff
@@ -1 +1,10 @@
...
+'localStorage' in this && (function() {
+	try {
+		// Needed for Safari private browsing.
+		this.localStorage.setItem('storage_test', 1);
+		this.localStorage.removeItem('storage_test');
+		return true;
+	} catch (e) {
+		return false;
+	}
+}())
```
dcherman: The last thing I'd suggest here is using `.call(this)`, that way you're accessing `localStorage` the same way throughout the entire feature check.

```js
'localStorage' in this && (function() {
	try {
		// Needed for Safari private browsing.
		this.localStorage.setItem('storage_test', 1);
		this.localStorage.removeItem('storage_test');
		return true;
	} catch (e) {
		return false;
	}
}).call(this);
```

JakeChampion: Good point! I've amended this is a subsequent commit

rowanmanning (approved):

PR #1349, opened 2017-09-28
Enable `URL` polyfill for all versions of Opera Mini.

### LINKED ISSUE #1348: Polyfill URL is not served in Opera Mini Browser
According to http://caniuse.com/#feat=url URL is not support on any version of Opera Mini.

Our error tracking tool shows errors for following versions of Opera Mini browser: 7.6.40234, 29.0.2254, 7.5.33361, 14.0.2065 with polyifill.io script included (https://cdn.polyfill.io/v2/polyfill.js?features=default&unknown=polyfill&flags=gated)

you can see it's not included when you test it here: https://cdn.polyfill.io/v2/polyfill.js?features=default&unknown=polyfill&flags=gated&ua=Opera/9.80%20(Android;%20Opera%20Mini/22.0.2254/69.190;%20U;%20en)%20Presto/2.12.423%20Version/12.16

It throws this => `Uncaught exception: ReferenceError: Undefined variable: URL`

matzimowski:
http://caniuse.com/#feat=url

closes #1348

### DIFF
--- a/polyfills/URL/config.json
+++ b/polyfills/URL/config.json
@@ -15,6 +15,7 @@
 		"safari": "*",
 		"firefox_mob": "<29",
 		"opera": "<36",
+		"op_mini": "*",
 		"android": "*",
 		"samsung_mob": "<5",
 		"ios_saf": "*"


matzimowski:
screen shot 2017-09-28 at 11 07 52

I ran tests and unfortunately they failed.
I checked polyfill repo for updates, but unfortunately there are none (https://github.com/inexorabletash/polyfill/blob/master/url.js).

JakeChampion:
Thanks for the issue and PR @matzimowski :-)

It looks like the bug in the URL polyfill would be coming from https://github.com/inexorabletash/polyfill/blob/master/url.js#L276-L302.

I'm not sure when I'll have time to look into the issue with using the base parameter in the URL constructor. Are you able to debug the issue at all?

PR #1362, opened 2017-10-16
Separate out IntersectionObserverEntry to serve a smaller polyfill to Edge 15

JakeChampion:
Edge 15 has IntersectionObserver but has an incorrect IntersectionObserverEntry as it is lacking `isIntersecting` - https://github.com/WICG/IntersectionObserver/issues/211

This PR fixes the dependency relationship between IntersectionObserver and IntersectionObserverEntry, serving the IntersectionObserverEntry polyfill to browsers which need it to have a working IntersectionObserver implementation.


triblondon:
A few questions about this:

* It looks like we already had a polyfill specific to the isintersecting problem in Edge, and this change adds a full implementation of `IntersectionObserverEntry` separate from `IntersectionObserver` and serves it to lots of browsers.  I can't see how this change relates to the isIntersecting issue...
* Can you provide some info on the circle config file?

JakeChampion:
This looks to have been a bad merge that I did.

IntersectionObserverEntry is currently not a dependency for IntersectionObserver, which means users need to specifically request it to ensure they support Edge 15.

I thought if we make IntersectionObserverEntry a dependency of IntersectionObserver, we should probably place the whole of IntersectionObserverEntry into the polyfill. That does take away the somewhat elegant solution of serving a 4 line polyfill for Edge 15. If you think we should only have the Edge 15 fix in the IntersectionObserverEntry polyfill, I'm happy to do that.

URL for Edge 15 and IntersectionObserver: https://polyfill.io/v2/polyfill.js?features=IntersectionObserver&ua=ie/15

URL for Edge 15 and IntersectionObserverEntry: https://polyfill.io/v2/polyfill.js?features=IntersectionObserverEntry&ua=ie/15

triblondon (approved):
Ok, sounds fine

PR #1375, opened 2017-10-30
update js-polyfill to get opera mini url fix

JakeChampion:
Fixes bad implementation spotted in #1349


JakeChampion:
@matzimowski it took a bit of time but we got the url polyfill working in opera mini 👍

PR #1376, opened 2017-11-02
Custom Elements polyfill

bedeoverend:
This builds off @alexflorisca PR #1101 to support `customElements`. As per the discussion over there, it adds in two new features, `customElements` and `customElements.~native-shim`.

The `customElements` polyfill will be loaded in browsers that don't support Custom Elements, and the `native-shim` will be loaded only on browsers that _do_ Custom Elements, as it's used to allow ES5 classes with the native Custom Element Registry. So if you're using ES5 code and want to work on all browsers, you'd request `features=customElements,customElements.~native-shim`

Differences from #1101:

- I've updated the browser support in `native-shim` to be only the browsers that support CEs natively
- I've updated the browser support in `customElements` to Opera 41 (as per this [ChromeStatus](https://www.chromestatus.com/feature/4696261944934400) page) and iOS Mobile Safari 10.1
- Re-written the tests for `customElements` so that the tests passed on IE, and removed tests that included Shadow DOM

Few questions:

- The [official support](https://github.com/webcomponents/webcomponentsjs#browser-support) for the CE polyfill is IE11+, Safari 9+ and all evergreens. Given that, should we update the browser config to target only explicitly supported browsers? It does note that support for older browsers may be possible with older polyfills (`MutationObserver`, `Map`, `Set` etc.) though it doesn't specify exactly which polyfills / how to make it work, so it would be difficult to guarantee. 
- To get the `customElements` tests working on IE11, I had to rewrite them using ES5 Functions instead of ES6 classes. Is it an issue that it doesn't test ES6 classes anymore?
- The `customElements` polyfill has feature detection inbuilt, which can be overridden, is it worth overriding this in a `postinstall` task, in case the user has used an `always` flag?

Let me know if there's anything else needed! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ npm-debug.log
 /polyfills/Array/of/polyfill.js
 /polyfills/atob/polyfill.js
 /polyfills/AudioContext/polyfill.js
+/polyfills/customElements/polyfill.js
 /polyfills/EventSource/polyfill.js
 /polyfills/HTMLPictureElement/polyfill.js
 /polyfills/Intl/polyfill.js
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ npm-debug.log
 /polyfills/atob/polyfill.js
 /polyfills/AudioContext/polyfill.js
 /polyfills/customElements/polyfill.js
+/polyfills/customElements/~native-shim/polyfill.js
 /polyfills/EventSource/polyfill.js
 /polyfills/HTMLPictureElement/polyfill.js
 /polyfills/Intl/polyfill.js
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "node": "6.11.1 - 8"
   },
   "dependencies": {
-    "@webcomponents/custom-elements": "github:webcomponents/custom-elements#v1.0.0-alpha.4",
+    "@webcomponents/custom-elements": "github:webcomponents/custom-elements#v1.0.0-rc.1",
     "Base64": "^1.0.0",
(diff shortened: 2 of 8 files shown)


bedeoverend:
I've updated the config to not try polyfill <=IE10 and Safari <=8, by using ranges in the config, as I figure we can't guarantee them working, so best to make that clear.

kapouer:
I'd advise to use https://www.npmjs.com/package/document-register-element, it has proven much more stable and backward-compatible here.
But the situation might have changed with recent developments...

Also for documentation:
https://hackernoon.com/extending-built-in-elements-9dce404b75b4
https://github.com/WebReflection/built-in-element

Having a Custom Elements v1 + built-in elements polyfill makes a lot of sense now.

0xR:
This PR seems old, but useful. Can we get an update on this? Is this going to be merged or not?

Is this project abandoned / looking for maintainers?

JakeChampion:
@0xR Hi, this project is not abandoned. We are actively working on a v3 and will not be deploying anything to the live version (v2) until v3 is finished. We are always happy to have people contribute polyfills. CustomElements is a tough one to polyfill, which is why it has yet to be merged.

robinwhittleton:
Now that v3 is out (or at least the v2 docs redirect to v3, which I guess is the same thing?), is there a plan to revisit this functionality?

JakeChampion:
Yes, if someone is willing to contribute the work I will gladly review and test it out :-)

jfbrennan:
Would really love to see Custom Elements polyfill added. It's badly needed now days as these have finally hit mainstream use

PR #1378, opened 2017-11-03
updates HTMLPictureElement config

benib:
ie 13, safari 9.1 and ios_saf 9.3 have shipped support for the `Picture` Element. This updates the polyfill config accordingly.

according to http://caniuse.com/#search=srcset there is a bug in Edge 13/14/15 that shows distorted images if `srcset` (that is also covered by this polyfill afaik) is used. Probably the polyfill cannot help in this case, so I guess it does no harm to not deliver it to Edge 13/14/15.

### DIFF
--- a/polyfills/HTMLPictureElement/config.json
+++ b/polyfills/HTMLPictureElement/config.json
@@ -1,12 +1,12 @@
 {
 	"browsers": {
-		"ie": "7 - *",
+		"ie": "7 - 12",
 		"ie_mob": "10 - *",
 		"firefox": "4 - 37",
 		"opera": "11.6 - 29",
-		"safari": "6 - *",
+		"safari": "6 - 9",
 		"chrome": "* - 38",
-		"ios_saf": "*",
+		"ios_saf": "* - 9.2",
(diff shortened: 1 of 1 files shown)


JakeChampion:
@benib, would you be able to test whether the polyfill solves the bug in Edge?

benib:
@JakeChampion I do not have a Edge 15 here to test. And the virtual machines on modern.ie have Edge 16 only (which has the bug already fixed and was released and rolled out mid October). So I am not really able to test whether the polyfill solves the bug.
From my understanding of https://github.com/scottjehl/picturefill/blob/master/src/picturefill.js the polyfill actually tests for srcset support, so it would not do something in Edge 13/14/15.
Also from reading through the bug report https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7778808/ it looks like it's pretty hard to reproduce.

I am not sure how to move on with this information.

JakeChampion (approved):

JakeChampion:
Thanks for linking to the bug report :-)
Your understanding of the polyfill is correct, as far as I can tell 👍

PR #1380, opened 2017-11-06
Add getAttributeNames polyfill

JSteunou:
Hi,

hope I did it right, I tried to polyfill this method https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttributeNames which is very convenient. So much convenient I though it was in every browsers since years... *sigh*

@JakeChampion could you guide me for the config and stuff? I tried to mimic others polyfills. Thanks.


polyfills/Element/prototype/getAttributeNames/config.json
```diff
@@ -0,0 +1,31 @@
+{
+    "aliases": [
+        "default-3.3",
```
JakeChampion: Right now I don't think we want to add this polyfill to any of the default sets.

In the future we may add it into the default set 👍

polyfills/Element/prototype/getAttributeNames/polyfill.js
```diff
@@ -0,0 +1,5 @@
+Element.prototype.getAttributeNames = function() {
+    return Array.from(this.attributes).map(function(a) {
```
JakeChampion: Array.from is a big polyfill, we can avoid the dependency on it by relying only on `Element` and `Array.prototype.map` like so:
```js
Element.prototype.getAttributeNames = function() {
  return Array.prototype.map.call(this.attributes, function(a) {
    return a.name;
  });
};
```

JakeChampion:
Thanks for the pull request @JSteunou 👍  This is looking good, I've done an initial review, I will run through the config.json at a later date. The way I usually create the config.json browser list is by checking the MDN page's browser support table.

JSteunou:
Yep, I did check support on MDN too, but I'm still in the dark for android, bb, opera mobile & samsung mobile

JakeChampion:
@JSteunou I will run some tests on those browsers and update the config in this PR 👍

JakeChampion:
@JSteunou The tests currently fail in IE7. You should be able to replicate this by starting the application via `npm start` and visiting the urls below in IE7.
```
- ie/7:
    -> Element.prototype.getAttributeNames should return an empty array when no attributes
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       85 == 0
    -> Element.prototype.getAttributeNames should return the list of all attribute names
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       87 == 2
    -> Element.prototype.getAttributeNames should return the list of all attribute names including id
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       85 == 1
    -> Element.prototype.getAttributeNames should return the list of all attribute names including data attributes
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       87 == 2
```

JSteunou:
Ho my... IE7 ! This far ?! I will try to find one

JSteunou:
Ok I think I find why, IE7 does not support `Element.prototype.setAttribute`. Are you sure to support IE7, there is even no polyfill for `Element.prototype.setAttribute` and Windows XP last browser is IE8.

I can rewrite my tests for this,  `setAttribute` is only use for the tests but meh... IE7...

JSteunou:
@JakeChampion friendly ping ;)

should I ignore IE7 support or add a `setAttribute` polyfill?

JakeChampion:
@JSteunou I've updated the tests to work in IE7, they will fail though as the polyfill doesn't work in IE7 because in IE7 the `attributes` property on Element objects contains 85 items for empty elements. E.G. `document.createElement('<div></div>').attributes.length === 85`

JakeChampion:
Failing tests on IE7:
```
Failures:
 - ie/7:
    -> Element.prototype.getAttributeNames should return an empty array when no attributes
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       85 == 0
 - ie/7:
    -> Element.prototype.getAttributeNames should return the list of all attribute names
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       87 == 2
 - ie/7:
    -> Element.prototype.getAttributeNames should return the list of all attribute names including id
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       85 == 1
 - ie/7:
    -> Element.prototype.getAttributeNames should return the list of all attribute names including data attributes
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.getAttributeNames
       87 == 2
```

JSteunou:
Damn... even after your changes, still the same error?

JakeChampion:
@JSteunou That error is because IE7 lists all options in the Element object attributes property instead of only the explicitly set attributes. It is a bug in IE7 that we would need to work around.

PR #1385, opened 2017-11-13
Add `iterable` support for `Map` and `Set` iterators.

simono:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

In IE11 `new Map().values()[Symbol.iterator]` is `undefined` without this patch. This means you can't use Map values in `for…of` loops.

### DIFF
--- a/polyfills/Map/polyfill.js
+++ b/polyfills/Map/polyfill.js
@@ -121,12 +121,20 @@
 		this.size = this._size = 0;
 	};
 	Map.prototype['values'] = function() {
-		return makeIterator(this, function(i) { return this._values[i]; });
+		var iterator = makeIterator(this, function(i) { return this._values[i]; });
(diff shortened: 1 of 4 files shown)


JakeChampion:
@simono Thanks for the PR! Would you be able to add some tests for the added feature to ensure we do not break/remove it in the future?

simono:
@JakeChampion sure! I'll update the PR in the next days.

simono:
@JakeChampion I've added tests. They fail in IE11 on master and pass on my branch.

JakeChampion:
@simono Thanks, running the tests on CI now

JakeChampion (approved):

simono:
@JakeChampion Thanks for the fast merge!

JakeChampion:
Thank you for the great PR 👍

PR #1386, opened 2017-11-13
Add EventTarget polyfill

JSteunou:
In order to add support for `passive` and `once` options. See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener


JakeChampion:
This codebase has moved to https://github.com/Financial-Times/polyfill-library, feel free to open a pull-request on that codebase and I will review it 👍

Issue #1367, opened 2017-10-25
IntersectionObserver Polyfill Out-of-Date


worsnupd:
Looks like there have been changes with the [official IntersectionObserver polyfill](https://github.com/w3c/IntersectionObserver/blob/master/polyfill/intersection-observer.js) that are not-yet reflected in Polyfill.io.

mathieuhays:
The browser support for Firefox should also be updated. (https://polyfill.io/v2/docs/features/#IntersectionObserver)

Native support for the IntersectionObserver has been enabled by default from version 55.
(https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Browser_compatibility)

JSteunou:
@mathieuhays already updated since last month, see https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/IntersectionObserver/config.json

PR #1387, opened 2017-11-13
Update to latest IntersectionObserver polyfill #1367

JSteunou:
keep IntersectionObserverEntry separate for Edge15

### DIFF
--- a/polyfills/IntersectionObserver/polyfill.js
+++ b/polyfills/IntersectionObserver/polyfill.js
@@ -1,17 +1,10 @@
 /**
  * Copyright 2016 Google Inc. All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * Licensed under the W3C SOFTWARE AND DOCUMENT NOTICE AND LICENSE.
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *  https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


JSteunou:
@JakeChampion could review this one please? I would love to see it in the next release ;)

polyfills/IntersectionObserver/polyfill.js
```diff
@@ -48,11 +41,7 @@ function IntersectionObserverEntry(entry) {
   this.rootBounds = entry.rootBounds;
   this.boundingClientRect = entry.boundingClientRect;
   this.intersectionRect = entry.intersectionRect || getEmptyRect();
-  try {
-    this.isIntersecting = !!entry.intersectionRect;
-  } catch (err) {
-    // This means we are using the IntersectionObserverEntry polyfill which has only defined a getter
```
JakeChampion: We need this part because we the IntersectionObserverEntry polyfill is only a getter https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/IntersectionObserverEntry/polyfill.js

JakeChampion (changes requested):
Currently this will throw an error if used alongside the IntersectionObserverEntry polyfill. Need to re-add some code which I have commented in-line.

polyfills/IntersectionObserver/polyfill.js
JSteunou: Ok I can add it back

JakeChampion:
@JSteunou I think we may need to also update our tests based on the ones from the upstream repo

JSteunou:
Good call!

JSteunou:
Thank you, I couldn't see nothing with that enormous diff

JSteunou:
@JakeChampion This polyfill now match last official release (0.5.0) If you have some time to review it

polyfills/IntersectionObserver/polyfill.js
JSteunou: The blocking review is still here even with the rollback oO 
https://github.com/Financial-Times/polyfill-service/pull/1387/commits/d4eac9e05256110f93cbee66d150a2e32a3897c1

PR #1389, opened 2017-11-14
Update engines to node 9

paradoxxxzero:
Node 9 is out since 10/31 and is working fine for me, but the package restriction breaks builds. (I have to force npm to `--ignore-engines`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "license": "CC0-1.0",
   "engines": {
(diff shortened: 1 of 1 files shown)


JakeChampion:
@paradoxxxzero Thanks for the PR, would you be able to sign our CLA? If not, we are not able to merge the pull-request

paradoxxxzero:
Not that I think this kind of PR really need a CLA, I did it.

PR #1393, opened 2017-11-20
Make Array.prototype.includes non-enumerable, fixes #1333

### LINKED ISSUE #1333: Polyfill Array.prototype.includes breaks for...in (IE 11)
When I use the Array.prototype.includes polyfill in IE11, any `for...in` loop over an array will return the element "includes".
See [here](http://jsbin.com/rubeniselo/1/edit?html,console) for a JS Bin (run in IE11).

I believe the problem lies in the polyfill implementation, which does `Array.prototype.includes = ...` instead of `Object.defineProperty(Array.prototype, 'includes', ...)`. When I use the [polyfill on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Polyfill), that uses `Object.defineProperty`, everything works as intended.

JakeChampion:


polyfills/Array/from/tests.js
```diff
@@ -7,7 +7,7 @@ it('has correct instance', function () {
 
 it('has correct name', function () {
 	function nameOf(fn) {
-		return Function.prototype.toString.call(fn).match(/function\s*([^\s]*)\(/)[1];
+		return Function.prototype.toString.call(fn).match(/function\s*([^\s]*)\s*\(/)[1];
```
JakeChampion, line 10: These helper functions were not fully correct as they did not take into consideration the fact that there can be any number of spaces between the function name and the opening parenthesis.

polyfills/Array/prototype/includes/tests.js
```diff
@@ -16,6 +29,10 @@ it('has correct argument length', function () {
 	proclaim.equal(Array.prototype.includes.length, 1);
 });
 
+ifSupportsDescriptorsIt('is not enumerable', function () {
```
JakeChampion, line 32: We have to use this because in environments where descriptors are not supported, you can not create a non-enumerable property.

polyfills/Array/prototype/includes/tests.js
```diff
@@ -1,13 +1,26 @@
 /* eslint-env mocha, browser */
 /* global proclaim */
 
+var supportsDescriptors = Object.defineProperty && (function () {
+	try {
+			var obj = {};
```
rowanmanning: These are tabbed in a bit too much. Shouldn't stop merging but it doesn't match the style of the rest of the code

rowanmanning (approved):
Looks good, one small comment but nothing major

PR #1396, opened 2017-11-22
Complex test case for setPrototypeOf

shauns:
Note that this method should modify its first argument


rowanmanning (approved):

shauns:
Thanks for approving. Worth mentioning that the unsupported use case is what babel uses for inheritance

JakeChampion:
@shauns Thanks, could you open a separate issue up which mentions this potential incompatibility with babel code please? If you could also create a small test case on https://babeljs.io/repl/ that would be really helpful :-)

Issue #1399, opened 2017-11-25
Object.defineProperty() being replaced in modern browsers with calls to deprecated methods that are sometimes undefined


mscalora:
I have run into a compatibility issue on a third party site that is using your polyfill.js and our plug-in's use of Polymer/webcomponents-lite.

Specifically, your implementation of `Object.defineProperty()` fails to use the native function but overrides the behaviour to invoke `__defineGetter__()`  and `__defineSetter__()` methods for plain-old-js-objects (POJSOs) but not DOM, Element instances, etc.

It appears your implementation of Object.defineProperty() naively assumes POJSOs have a `__proto__` with `__defineGetter__()`  and `__defineSetter__()` Object methods. It fails when this is not the case when a POJSO does not inherit from Object as:

```var obj = Object.create(null, {});```

I don't know why Polymer/webcomponents-lite uses this type of POJSOs but they do in certain complex initialization cases and your `Object.defineProperty()` fails with `__defineGetter__()`  and `__defineSetter__()` undefined exceptions while the native Object.defineProperty() has no problems. Chrome and Firefox's native Object.defineProperty() works perfectly on this type of POJSO.

Firstly I find it odd that your polyfill replaces `Object.defineProperty()` in the very latest Chrome and Firefox browsers, are these implementations deficient in some respect? I see no support in the standard for the idea that `__defineGetter__()`  and `__defineSetter__()` `Object` methods should be preferred on POJSOs nor do I see any reason to unconditionally replace `Object.defineProperty()` with `__defineGetter__()`  and `__defineSetter__()` Object methods.

At the very least, when `object.__defineGetter__`  or `object.__defineSetter__` are undefined, the `Object.defineProperty()` should be used rather than inducing exceptions in latest versions of these evergreen browsers.

Mozilla documentation suggests that `__defineGetter__()`  and `__defineSetter__()` are deprecated but I didn't find explicit mention of that in the standards documents.  [MDN Wed Docs: __defineGetter__](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__)

I'm going to submit a pull request but given my inexperience in the world of polyfills and lack of understanding around the rational for some of the design and implementation decisions I'm somewhat unsure of the correct approach to resolve the issue I'm investigating.

Mike Scalora
Chief Architect - The World Table

Refs:

[ECMAScript 5.1 (ECMA-262)
[...]

PR #1400, opened 2017-11-25
Use native Object.defineProperty for POJSOs when needed

mscalora:
Plain old JS objects without __defineGetter__ or __defineSetter__ should use the native Object.defineProperty() function to prevent undefined function exception.

See: [issue #1399](https://github.com/Financial-Times/polyfill-service/issues/1399) for more detail.


JakeChampion:
This has been fixed via #1408

PR #1402, opened 2017-11-28
Re-enable RUM, using BigQuery

triblondon:
This PR re-enables RUM logging for participating users.

### Library changes

* Output RUM logging even if the bundle would otherwise be totally empty (it's worth doing feature tests in this scenario, so we can check we're not omitting to polyfill where we need to)
* Limit the number of tests per response to 10, choosing a random set from the set of features requested (this will get cached, but given the number of different batches and bundles and cache servers, the variety should still be sufficient).  I added the `shuffle-array` module for this because it's much more efficient than a random sort

### RUM client changes

* Round the perf numbers to whole integers, as sub-ms precision is really unnecessary
* Report detect results using `detect{n}={featureName}_{result}` syntax, because in VCL we can't enumerate unknown query string keys.  This is also why the number of tests per report needs to be limited to 10.

### VCL changes

* Moved S3 logging into VCL so it is versioned (S3 endpoint needs to remain defined in Fastly UI but automatic logging to it deactivated with a request condition)
* Added logging of RUM data to BigQuery
    * Generate a random `rumRequestID`, current time plus 10 char string = 10^18 possible values each second.
    * Sanitise IP and hostname.  IP is the 32 bits a 16 bit offset from the end of the address, which means the first 16 bits of an IPv4, and the 32 bits of an IPv6 between bit 80 and bit 112.  Referer is reduced to just a hostname, and `www.` prefixes are removed.
    * Individual detect results are logged to a separate BigQuery table, one row per result.  VCL cannot iterate so it's the same block of code repeated 10 times.
    * The main RUM log string as JSON is echoed back to the reporting client in a header.  This helps debug RUM reports, and should not have security implications since it provides only data the user has themselves supplied, or which they could themselves determine (aside from the request ID)

### External changes pre-deploy

* Set up a private edge dictionary for the hash salt and populate it.  The value doesn't matter, as long as it is complex, but it should then remain constant. [DONE in prod and QA]

### External changes post-deploy

* Reinstate the `RUM_BEACON_HOST` env var in Heroku


vcl/main.vcl
```diff
@@ -161,8 +161,115 @@ sub vcl_recv {
...
+			{""country":""} cstr_escape(geoip.country_code) {"","}
+			{""data_center":""} cstr_escape(if(req.http.Cookie:FastlyDC, req.http.Cookie:FastlyDC, server.datacenter)) {"","}
+			{""refer_domain_hash":""} cstr_escape(regsub(digest.hash_sha256(var.hashSalt var.safeRef), "^(.{15}).*?$", "\1")) {"","}
+			{""client_ip_hash":""} cstr_escape(regsub(digest.hash_sha256(var.hashSalt var.safeIP), "^(.{15}).*?$", "\1")) {"","}
+			{""ua_family":""} cstr_escape(regsub(urldecode(req.http.Normalized-User-Agent), "^(\w+)\/.*$", "\1")) {"","}
+
+			# Integers
+			# CAREFUL: These variables are concatenated into a raw JSON string.
+			# Ensure they are valid numbers
+			{""perf_dns":"} regsub(req.url.qs, "^(.*\&)?dns=(\d+).*?$", "\2") ","
```
JakeChampion: Should we use `std.atoi` here?

>Takes a string (which represents an integer) as an argument and returns its value.Example:if (std.atoi(req.http.X-Decimal) == 42) { set req.http.X-TheAnswer = "Found"; }
https://docs.fastly.com/guides/vcl/miscellaneous-VCL-extensions

JakeChampion (approved):

vcl/main.vcl
triblondon: No reason to cast it to an int when we actually need a string anyway.  But there's a hole here: if the regex doesn't match you end up dumping the entire querystring into the output.  I'm refactoring it.

JakeChampion (approved):

Issue #748, opened 2016-07-04
Intl not served for Blackberry 10


aflatter:
The Intl polyfill is not served for Blackberry 10, but it's necessary according to [caniuse](http://caniuse.com/#search=intl).

```
$ curl -A "Mozilla/5.0 (BB10; Kbd) AppleWebKit/537.35+ (KHTML, like Gecko) Version/10.3.2.2876 Mobile Safari/537.35+" "https://cdn.polyfill.io/v2/polyfill.js?features=Intl.~locale.de,Intl.~locale.en&unknown=polyfill"
/* Polyfill service v3.7.1
 * For detailed credits and licence information see http://github.com/financial-times/polyfill-service.
 *
 * UA detected: bb/10.3.0
 * Features requested: Intl.~locale.de,Intl.~locale.en
 *  */
```

triblondon:
No problem.  See https://cdn.polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers for how to fix this.

triblondon:
We don't have anyone stepping up to test this, and FT devs are not supporting blackberry devices with script, so closing.  If anyone wants this, please see #945 and do the testing to complete this work.

PR #1418, opened 2017-12-08
Serve Intl polyfill for Opera Mini and Blackberry

radev:
#748 Happens for us (we receive Sentry reports of errors including user agent)
So, I propose adding Intl polyfill for Blackberry and Opera Mini.

I have checked https://caniuse.com/#search=intl and it's not supported on all Blackberry and Opera Mini versions (I don't have BB device to verify, but I can confirm Opera Mini).

### DIFF
--- a/polyfills/Intl/config.json
+++ b/polyfills/Intl/config.json
@@ -1,10 +1,12 @@
 {
 	"browsers": {
 		"android": "<=4.3",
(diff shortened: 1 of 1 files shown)


JakeChampion:
@radev Have you ran the test suite on those browsers?

PR #1420, opened 2017-12-12
Removed zlib from dependencies

tomdid:
Removed zlib package from dependencies. 
zlib package is deprecated and comes with node.js by default since v.0.10 (https://nodejs.org/docs/latest-v0.10.x/api/zlib.html)

Also installing deprecated zlib package has caused deployment issues in our  environment:
```
...
> zlib@1.0.5 preinstall /home/centos/jenkins_home/workspace/Build_FrontEnd_GameApp/node_modules/.staging/zlib-31869530
> node-waf clean || true; node-waf configure build
>sh: node-waf: command not found
...
```


JakeChampion (approved):

JakeChampion:
@tomdid thanks for the pr, before we can merge it we require the CLA to be signed. Are you able to sign our CLA?

PR #1421, opened 2017-12-12
Enable `MutationObserver` polyfill for all versions of Opera Mini

ctjhoa:
https://caniuse.com/#feat=mutationobserver

### DIFF
--- a/polyfills/MutationObserver/config.json
+++ b/polyfills/MutationObserver/config.json
@@ -9,6 +9,7 @@
 		"firefox_mob": "<14",
 		"ie_mob": "<11",
 		"op_mob": "<15",
+		"op_mini": "*",
 		"ios_saf": "<7"
 	},
 	"repo": "https://github.com/megawac/MutationObserver.js",


JakeChampion:
Hi @ctjhoa, have you had a look through our contribution documentation with regards to serving a polyfill to a new browser? https://polyfill.io/v2/docs/contributing/common-scenarios#adding-browsers

We would be really grateful if you could take a screenshot of the browser running the test suite for this polyfill. The url to run the tests with the polyfills is https://polyfill.io/test/tests?feature=MutationObserver&mode=all

ctjhoa:
I'm very confused about this bug. Here is what I've got from my monitoring system:

2018-01-09-085450_3839x1994_scrot

Here is my polyfill.io script
```
<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=default,matchMedia,MutationObserver"></script>
```
Using [caniuse test](https://tests.caniuse.com/?feat=mutationobserver&ua=omini5), MutationObserver should work without prefix (but the test is still considered as failed)

In addition, I have tested this url https://polyfill.io/test/tests?feature=MutationObserver&mode=all with a real opera mini with both "extreme" data savings mode enabled and disabled and the test is still successful (since caniuse seems to be based on that https://github.com/Fyrd/caniuse/issues/3883).

Do you have experienced such weird cases in the past? I'm little bit lost...

Thanks

PR #1425, opened 2017-12-14
Fix Element.prototype.closest for elements inside inline SVGs

### LINKED ISSUE #1424: Element.prototype.closest does not work on any SVG node
This was covered in part by #1279, but not only `<svg>` elements themselves cannot access their parent element by `.parentElement` in IE but also all elements belonging to the SVG namespace.

See it in this [jsFiddle](https://jsfiddle.net/0s3dn4jc/2/).

I'll attach a PR to fix this. :)

loilo:
Fixes #1424:

All elements belonging to the SVG namespace

* do not have a `parentElement` property in IE.
* are instances of the `SVGElement`.

This PR extends the fix of #1285 to not only `<svg>` elements but all elements in the SVG namespace.

### DIFF
--- a/polyfills/Element/prototype/closest/polyfill.js
+++ b/polyfills/Element/prototype/closest/polyfill.js
@@ -3,7 +3,7 @@ Element.prototype.closest = function closest(selector) {
 
 	while (node) {
 		if (node.matches(selector)) return node;
-		else node = node.tagName === 'svg' ? node.parentNode : node.parentElement;
+		else node = node instanceof SVGElement ? node.parentNode : node.parentElement;
 	}
 
 	return null;
--- a/polyfills/Element/prototype/closest/polyfill.js
+++ b/polyfills/Element/prototype/closest/polyfill.js
@@ -3,7 +3,7 @@ Element.prototype.closest = function closest(selector) {
 
 	while (node) {
 		if (node.matches(selector)) return node;
-		else node = node instanceof SVGElement ? node.parentNode : node.parentElement;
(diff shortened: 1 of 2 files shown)


JakeChampion:
@Loilo Thanks for the PR, it looks to fail the tests on IE7 & IE8, would you be able to look into fixing those?

```
Failures:
 - ie/8:
    -> Element.prototype.closest should return the first ancestor that matches selectors
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.closest
       'SVGElement' is undefined
 - ie/8:
    -> Element.prototype.closest should return null if there are no matches
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.closest
       'SVGElement' is undefined
 - ie/7:
    -> Element.prototype.closest should return the first ancestor that matches selectors
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.closest
       'SVGElement' is undefined
 - ie/7:
    -> Element.prototype.closest should return null if there are no matches
       http://127.0.0.1:3000/test/tests?mode=targeted&feature=Element.prototype.closest
       'SVGElement' is undefined
```

loilo:
Oh. Right, IE < 9 does not support inline SVG at all, so I guess I'll need to exclude that case.

loilo:
Okay, I added a check for existence of `SVGElement`. IE 7/8 should be fine now.

JakeChampion (approved):

JakeChampion:
Thanks for this work @Loilo!

loilo:
My pleasure. 🙂

Issue #1429, opened 2017-12-19
Can't start server: Cannot find module './build/bindings/iltorb.node'


njbotkin:
This was first mentioned in https://github.com/Financial-Times/polyfill-service/pull/1397.

Upon `npm run start`, I receive `Error: Cannot find module './build/bindings/iltorb.node'`

For future issue-havers, please post these details:

- Node.js version: `node -v`
- npm version: `npm -v`
- Ubuntu version: `lsb_release -a`
- Kernel version: `uname -a`
- Python version: `python --version`
- gcc version: `gcc --version`

njbotkin:
My environment:

- Node.js version: `v9.2.0`
- npm version: `5.5.1`
- Ubuntu version: `Ubuntu 16.04.2 LTS`
- Kernel version: `Linux Exordium 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014 x86_64 x86_64 x86_64 GNU/Lin`ux
- Python version: `Python 2.7.12`
- gcc version: `(Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609`

My kernel version looks funky because I'm running Windows Subsystem for Linux.

njbotkin:
Got it to work.  I needed to run `npm i` in

- `node_modules/iltorb`
- `node_modules/node-zopfli`

Which compiled their bindings.

njbotkin:
Should this have happened automatically, or should this extra build step be added to the documentation?

JakeChampion:
Yes, npm should automatically build these during the installation. I'm not sure why it would not do that, I've tried on my macbook using node 9 with npm 5 and it worked correctly.

Could you try installing node-pre-gyp first and then install polyfill-service please, I'm wondering whether we need to include this as a dependency in order to find the correct pre-built binaries instead of building them on your machine manually.

njbotkin:
Huh.  I just cleared out my node_modules, reinstalled with node-pre-gyp globally installed, and everything worked great.  Then I cleared out node_modules again, reinstalled with node-pre-gyp uninstalled, and everything worked again.  Strange.

However, both times, all attempts to fetch binaries failed, no attempt appears to have been made to fetch oniguruma's binaries:

`node-pre-gyp ERR! Tried to download(403): https://node-zopfli.s3.amazonaws.com/Release/zopfli-v1.4.0-node-v59-linux-x64.tar.gz
[...]

Issue #1233, opened 2017-06-23
Add NodeList polyfills


gu-stav:
At the guardian we [realized yesterday](https://github.com/guardian/frontend/pull/17225), that there is currently no polyfill for `NodeList.forEach()`. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach) there is no support for the feature in the following browsers:

|Browser|Version| 
|---|---|
|Chrome|< 51|
|Firefox| <50
|IE|*|
|Opera|<38|
|Safari|<10|

The possible polyfill would be quite small:

```js
if (window.NodeList && !NodeList.prototype.forEach) {
    NodeList.prototype.forEach = function (callback, argument) {
        argument = argument || window;
        for (var i = 0; i < this.length; i++) {
            callback.call(argument, this[i], i, this);
        }
    };
}
```

Do you think it's worth adding this as a single polyfill, or would you rather like to polyfill all potential missing methods on `NodeList`, which are `.entries(), forEach(), item(), keys(), values()`?

I'm happy to work on a PR for this.

JakeChampion:
A single polyfill for `NodeList.prototype.foreach` would be great. If you want to add all the other methods as other individual polyfills that would be very much appreciated :-)

gu-stav:
@JakeChampion When I wanted to fill out the Spec field, I discovered this actually isn't part of any Spec. FF, Chrome and Webkit just tries to help helpful. I guess that's a reason why `.forEach()` shouldn't go in here, right? The other ones are in the Spec.

JakeChampion:
@gustavpursche Everything except `forEach` is inside the DOM specification: https://dom.spec.whatwg.org/#interface-nodelist

JakeChampion:
I think we should avoid adding `forEach` since it is not in a specification. Thanks for taking the time to find all this out.

gu-stav:
I agree. I'll ask the flow people aswell why [they put it in](https://github.com/facebook/flow/blob/v0.48.0/lib/dom.js#L482). Thanks!

jakearchibald:
@JakeChampion 

> I think we should avoid adding forEach since it is not in a specification

https://heycam.github.io/webidl/#idl-iterable says any iterable interface should have forEach. However, I believe there were some libraries that used `thing.map` etc as a way of assuming "This is an array, not a nodelist", so browsers haven't added it.

JakeChampion:
[...]

PR #1430, opened 2017-12-19
Add NodeList.prototype.forEach polyfill

njbotkin:
Implements #1233.

I couldn't find good support information for 

- ios_chr
- op_mob
- op_mini
- samsung_mob
- bb

Obviously, this assumes that the implementation of `NodeList.prototype.forEach` is identical to `Array.prototype.forEach`.  So I also copied tests directly from `Array.prototype.forEach`.  Because of #1429, I couldn't run the tests myself :(

### DIFF
--- a/polyfills/NodeList/prototype/forEach/config.json
+++ b/polyfills/NodeList/prototype/forEach/config.json
@@ -1,0 +1,19 @@
+{
+	"browsers": {
+		"ie": "*",
+		"ie_mob": "*",
+		"chrome": "<51",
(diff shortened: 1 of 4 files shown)


amarinov:
The tests use arrays, not nodelists. If the polyfill changes and is broken, the tests will continue to pass successfully. I believe the tests shouldn't be bound to what the implementation is.

JakeChampion:
This polyfill was added in #1686 👍

PR #1432, opened 2017-12-20
Rewrite Map polyfill to be more spec compliant

JakeChampion:
Inline with my other updates to polyfills, I have added the specification steps inline in the polyfill.

A nice benefit from this rewrite is on the Map.prototype.forEach method, we now only iterate the map once. We used to iterate it twice.

Another nice benefit is we don't need a `NaN` symbol marker anymore, that was only needed because we were using `Array.prototype.indexOf` to find keys instead of following the specification which states that we should iterate through the keys and use `sameValueZero` (which handles `NaN` correctly.

In a later PR we can look to optimise the size of this polyfill by creating some helpers functions around defining properties.

Update: This polyfill now supports custom iterators to be used to construct a Map object 🎉. I also made it work correctly with argument objects (for browsers which do not have iteration methods attached to those objects).

### DIFF
--- a/polyfills/Map/config.json
+++ b/polyfills/Map/config.json
@@ -23,7 +23,9 @@
 		"Symbol",
 		"Symbol.iterator",
 		"Symbol.species",
+		"Object.create",
 		"Object.defineProperty",
+		"Object.getPrototypeOf",
 		"Number.isNaN"
 	],
 	"notes": [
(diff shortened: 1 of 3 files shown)


rowanmanning (approved):
This is great! I haven't run it but looking over the code can't see any gotchas and your tests seem to cover everything. Excellent 😄 though if it were anything other than the Polyfill service I'd tell you to stop working and do Christmas things

PR #1433, opened 2017-12-21
Alias GoogleBot 2.1 to Chrome 41

### LINKED ISSUE #1405: Support Googlebot as a proper User Agent
I was just bitten by the fact that polyfill.io doesn't recognize Googlebot and, by default, sends it no polyfills. I was able to fix by adding `unknown=polyfill` to the URL, but didn't even realize that a missing polyfill was the issue because Googlebot fetch-and-render makes it hard to surface errors.

Is there a reason that Googlebot isn't handled by default? It's built on Chrome 41, so Polyfill.io should be capable of sending the polyfills it needs. Is there interest in handling Googlebot gracefully, or a specific reason it's ignored by default?

JakeChampion:
GoogleBot is currently an unknown/unsupported browser in the eyes of the polyfill-service. This causes issues for sites crawled by GoogleBot as we would not serve the correct set of polyfills to make the site work for GoogleBot. Even setting the `unknown` property to `polyfill` does not work for all cases as our `Element` polyfill seems to throw an error in browsers which support `Element` natively (Which GoogleBot does).

It turns out that GoogleBot 2.1 is using Chrome 41, we can alias this specific version of GoogleBot to that specific version of Chrome, this will make the polyfill-service start serving the correct polyfills for GoogleBot.

In another PR I will look to make the `Element` polyfill not throw an error in browsers which support it natively.

This PR fixes https://github.com/Financial-Times/polyfill-service/issues/1405.

### DIFF
--- a/lib/UA.js
+++ b/lib/UA.js
@@ -139,6 +139,10 @@ const aliases = {
 		"45": ["chrome", 58],
 		"46": ["chrome", 59],
 		"47": ["chrome", 60]
+	},
+
+	"googlebot": {
+		"2.1": ["chrome", 41]
 	}
 };
 
@@ -174,7 +178,6 @@ function UA(uaString) {
 
 		// Electron ` Electron/X.Y.Z` (see #1129)
 		uaString = uaString.replace(/ Electron\/[\d\.]+\d+/i, '');
-
 
 		this.ua = useragent.parse(uaString);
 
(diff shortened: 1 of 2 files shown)


rowanmanning (approved):

TehShrike:
I'm not familiar with the polyfill.io deployment schedule - when will this get deployed?

TehShrike:
Sorry, I just checked the readme - I'll watch the Twitter account for notifications on 3.25.

JakeChampion:
This will get promoted from qa.polyfill.io to production (cdn.polyfill.io and polyfill.io) tomorrow.

PR #1438, opened 2018-01-03
Update IntersectionObserver{,Entry} compatibility table.

dracos:
Safari has no native support at all, that I can see; Firefox is polyfilled until version 55 (it was present in 53 but behind a flag); the polyfill says it works in IE7+ and Safari 6+.

Hope I've understood this right, apologies if not!

### DIFF
--- a/docs/assets/compat.json
+++ b/docs/assets/compat.json
@@ -7073,60 +7073,6 @@
       "9.1": "native"
     }
   },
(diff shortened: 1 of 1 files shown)


JakeChampion:
Thanks for the pull-request, the codebase has changed a lot and no longer has this file.

PR #1443, opened 2018-01-03
Add Blob polyfill

JakeChampion:


polyfills/Blob/polyfill.js
```diff
@@ -0,0 +1,236 @@
+// Local modification of https://github.com/eligrey/Blob.js/blob
+
+/* Blob.js
+	 * A Blob implementation.
+	 * 2014-07-24
+	 *
```
rowanmanning: This file seems to be over-indented past line 3

rowanmanning (approved):
Small whitespace issue but otherwise good

rowanmanning (approved):

PR #1444, opened 2018-01-03
Add ECMAScript 262 abstract operations to base polyfills off of

JakeChampion:
This will make it simpler to implement polyfills as they will mostly a copy paste of the written specification.


JakeChampion:
I've split this PR into lots of smaller PRs to make reviewing simpler

PR #1486, opened 2018-01-07
Add SpeciesConstructor abstract operation

JakeChampion:


polyfills/_ESAbstract/SpeciesConstructor/polyfill.js
```diff
@@ -0,0 +1,27 @@
...
+	// 7.3.20.3 If C is undefined, return defaultConstructor.
+	if (C === undefined) {
+		return defaultConstructor;
+	}
+	// 7.3.20.4 If Type(C) is not Object, throw a TypeError exception
+	if (Type(C) !== 'object') {
+		throw new TypeError('O.constructor is not an Object');
+	}
+	// 7.3.20.5 Let S be ? Get(C, @@species).
+	var S = typeof this.Symbol === 'function' && typeof this.Symbol.species === 'symbol' ? C[this.Symbol.species] : undefined;
```
notlee, line 16: Is `this. Symbol` intentional? I don't fully understand this line. 🤔

JakeChampion: 1) `this.Symbol` can be switched to `Symbol` since `this` is the global object. Would that be more understandable?

JakeChampion: 2. This line is checking that Symbols are natively supported, if they are it will actually execute the step `Get(C, @@species)`, otherwise it will returned undefined (which is okay to be returned as the operation will return `defaultConstructor` in that case as part of step 6).

notlee: Ah sorry. I misunderstood Symbol and thought the call may be context dependent. :facepalm: 
This is 👍 .

notlee (approved):

PR #1508, opened 2018-01-08
Make Object.entries tests use ES3 syntax to enable them to run in ES3 environments

JakeChampion:
Currently these tests will never run in IE7 or IE8 due to the use of the `get` syntax being used. This causes issues for our script which generates the compatibility data.

https://github.com/Financial-Times/polyfill-service/blob/5619eb89b5fad1c79135c6c9ff08ccfdaf47080e/docs/assets/compat.json#L9204-L9209

### DIFF
--- a/polyfills/Object/entries/tests.js
+++ b/polyfills/Object/entries/tests.js
@@ -50,18 +50,26 @@ it('has length `1`', function() {
 	proclaim.equal(Object.entries.length, 1);
 });
 
-it('should terminate if getting a value throws an exception', function() {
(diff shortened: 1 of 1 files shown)


JakeChampion:
After we get this issue fixed we can look into a script to compare the compatibility data against our polyfill configurations to see if improvements can be made. I've done a first pass at this work and found many polyfill configurations (or polyfill tests/detects) that can be improved.

rowanmanning (approved):

PR #1524, opened 2018-01-10
Rewrite Array.prototype.includes to be spec compliant and use es abstract methods

JakeChampion:


notlee:
Pedantic: Shall we update the spec link to https://tc39.github.io/ecma262/#sec-array.prototype.includes?

notlee (approved):

PR #1525, opened 2018-01-10
Rewrite Array.prototype.indexOf to be spec compliant and use es abstract methods

JakeChampion:


polyfills/Array/prototype/indexOf/polyfill.js
```diff
@@ -1,20 +1,50 @@
...
+	// 4. Let n be ? ToInteger(fromIndex). (If fromIndex is undefined, this step produces the value 0.)
+	var n = ToInteger(arguments[1]);
+	// 5. If n ≥ len, return -1.
+	if (n >= len) {
+		return -1;
+	}
+	// 6. If n ≥ 0, then
+	if (n >= 0) {
+		// a. If n is -0, let k be +0; else let k be n.
+		var k = n === -0 ? 0 : n;
```
rowanmanning, line 20: In cases like this (where `var k` appears in both the `if` and `else`), should the declaration be moved out?

rowanmanning (approved):

polyfills/Array/prototype/indexOf/polyfill.js
JakeChampion: It could be moved out, do you think it should be moved out? 
As the variable declaration is done with `var`, it gets hoisted to the top of the function scope. This double declaration would get compiled away when users request the minified bundle.

PR #1527, opened 2018-01-10
Rewrite Array.prototype.lastIndexOf to be spec compliant and use es abstract methods

JakeChampion:


polyfills/Array/prototype/lastIndexOf/polyfill.js
```diff
@@ -1,20 +1,42 @@
...
-	length = Math.max(Math.min(arraylike.length, 9007199254740991), 0) || 0,
-	index = Number(arguments[1]) || 0;
-
-	index = 1 in arguments ? (index < 0 ? Math.max(length + index, 0) : index) + 1 : length;
-
-	while (--index >= 0) {
-		if (index in arraylike && arraylike[index] === searchElement) {
-			return index;
+	// 4. If fromIndex is present, let n be ? ToInteger(fromIndex); else let n be len-1.
+	var n = arguments.length > 1 ? arguments[1] : len - 1;
```
rowanmanning: Why no `ToInteger` here?

rowanmanning (approved):
One question, but otherwise good

polyfills/Array/prototype/lastIndexOf/polyfill.js
JakeChampion: Something I missed, good catch, I'll add it now 👍

PR #1533, opened 2018-01-10
Rewrite Function.prototype.bind to be spec compliant and use es abstract methods

JakeChampion:


rowanmanning (approved):
The naming in here doesn't match the spec as with other polyfills, but assuming there's a reason for that and the TODOs 🙂

JakeChampion:
It is based on https://github.com/es-shims/es5-shim/blob/d6d7ff1b131c7ba14c798cafc598bb6780d37d3b/es5-shim.js#L182 but has been modified to make use of the other polyfills we have in the service

PR #1534, opened 2018-01-10
Rewrite Map to be spec compliant and use es abstract methods

JakeChampion:


rowanmanning (approved):
Code itself looks good, there are a few whitespace issues with the `polyfills/Map/polyfill.js` (extra tabs and closing curlys in unexpected places) but nothing that should prevent merging

PR #1550, opened 2018-01-10
Rewrite Math.trunc to be spec compliant and use es abstract methods

JakeChampion:


gvonkoss (approved):

gvonkoss:
Just out of curiosity, does this not get similar comments as the rest of them?

JakeChampion:
@gvonkoss It doesn't have any explicit steps, which is why I did not add any inline comments.

PR #1587, opened 2018-01-10
Rewrite String.prototype.padStart to be spec compliant and use es abstract methods

JakeChampion:


polyfills/String/prototype/padStart/polyfill.js
```diff
@@ -1,17 +1,37 @@
...
+	// 5. If intMaxLength is not greater than stringLength, return S.
+	if (intMaxLength <= stringLength) {
+		return S;
+	}
+	// 6. If fillString is undefined, let filler be the String value consisting solely of the code unit 0x0020 (SPACE).
+	if (fillString === undefined) {
+		var filler = ' ';
+		// 7. Else, let filler be ? ToString(fillString).
+	} else {
+		filler = ToString(fillString);
```
gvonkoss: Does this need to be declared as a variable?

gvonkoss (changes requested):

gvonkoss (approved):

PR #1588, opened 2018-01-10
Rewrite String.prototype.padEnd to be spec compliant and use es abstract methods

JakeChampion:


polyfills/String/prototype/padEnd/polyfill.js
```diff
@@ -1,17 +1,37 @@
...
+	// 5. If intMaxLength is not greater than stringLength, return S.
+	if (intMaxLength <= stringLength) {
+		return S;
+	}
+	// 6. If fillString is undefined, let filler be the String value consisting solely of the code unit 0x0020 (SPACE).
+	if (fillString === undefined) {
+		var filler = ' ';
+		// 7. Else, let filler be ? ToString(fillString).
+	} else {
+		filler = ToString(fillString);
```
rowanmanning: Missing `var`? It's in the condition above.

rowanmanning (changes requested):
Small issue, think we maybe have an undefined variable

polyfills/String/prototype/padEnd/polyfill.js
rowanmanning: Also noticed that the same issue is present in `padStart`

rowanmanning (approved):

PR #1590, opened 2018-01-10
Rewrite String.prototype.endsWith to be spec compliant and use es abstract methods

JakeChampion:


notlee:
Pedantic... spec link: https://tc39.github.io/ecma262/#sec-string.prototype.endswith

notlee (approved):

Issue #1596, opened 2018-01-12
IE11 broke


kohesion:
![image](https://user-images.githubusercontent.com/1246027/34899476-225fa80a-f7c7-11e7-8a8e-84a8c5197152.png)

LOK-Soft:
can you estimate when the committed fix for this issue will be on production?
Thanks :)

JakeChampion:
Middle of next week 👍

AchimKre:
Is there a workaround for this? Because this breaks our page completely down in IE 11.

kohesion:
Just download the file in IE11 and wrap the problem bit in a try catch block

JakeChampion:
@AchimKre Yes, if you add `&excludes=Object.assign` to the end of your polyfill.io script url, that will fix the issue.

AchimKre:
@JakeChampion Thx for your hint, but it does not work for us. It seems an other polyfill needs Object.assign and only excludes=Object.assign runs into an other javascript error.

We added the polyfill Object.assign polyfill from [1] inline before include https://cdn.polyfill.io/v2/polyfill.min.js?excludes=Object.assign

That seems to work.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

JakeChampion:
Hi @kohesion, @LOK-Soft, @AchimKre, @Turbo87, @tolgabalci, @bthibault, @gavinlove. If you could please test out https://qa.polyfill.io, it contains the fix, once proven to work I can put it into production.

JakeChampion:
This is now live in production on https://polyfill.io.

AchimKre:
It works. Thank you!

PR #1560, opened 2018-01-10
Rewrite Object.assign to be spec compliant and use es abstract methods

JakeChampion:

PR #1598, opened 2018-01-15
Make Object.assign not readonly

### LINKED ISSUE #1595: IE11 Error: Assignment to read-only properties is not allowed in strict mode
After the update today:
* Polyfill v3.25.0 loads and breaks in IE 11

When using:
https://cdn.polyfill.io/v2/polyfill.js?features=default,Array.prototype.includes,Array.prototype.find,Array.prototype.findIndex,Object.values

Getting error (and page fails to load properly):
`Assignment to read-only properties is not allowed in strict mode` in the below code on line `Object.assign = require('object-assign');`

`// @remove-on-eject-begin
/**
 * Copyright (c) 2015-present, Facebook, Inc.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
// @remove-on-eject-end
'use strict';

if (typeof Promise === 'undefined') {
  // Rejection tracking prevents a common issue where React gets into an
  // inconsistent state due to an error, but it gets swallowed by a Promise,
  // and the user has no idea what causes React's erratic future behavior.
  require('promise/lib/rejection-tracking').enable();
  window.Promise = require('promise/lib/es6-extensions.js');
}

// fetch() polyfill for making API calls.
require('whatwg-fetch');

// Object.assign() is commonly used with React.
// It will use the native implementation if it's present and isn't buggy.
Object.assign = require('object-assign');

//////////////////
// WEBPACK FOOTER
// ./node_modules/react-scripts/config/polyfills.js
// module id = 392
// module chunks = 0`

TobiasBales:
this should fix #1595 and #1596

If any other library tries to polyfill Object.assign the previous change breaks in IE 11 in strict mode
since Object.defineProperty creates readonly properties.

### DIFF
--- a/polyfills/Object/assign/polyfill.js
+++ b/polyfills/Object/assign/polyfill.js
@@ -1,68 +1,65 @@
 (function() {
-
 	// 7.1.13 ToObject ( argument )
 	function toObject(argument) {
-    if (argument === null || argument === undefined) {
+		if (argument === null || argument === undefined) {
 			throw new TypeError('Cannot call method on ' + argument);
 		}
-    return Object(argument);
-  }
+		return Object(argument);
+	}
 
-	Object.defineProperty(Object, 'assign', {
-		// 19.1.2.1 Object.assign ( target, ...sources )
-		value: function assign(target, source) { // eslint-disable-line no-unused-vars
+	Object.assign = function assign(target, source) {
+		// eslint-disable-line no-unused-vars
 
-			// 1. Let to be ? ToObject(target).
-			var to = toObject(target);
+		// 1. Let to be ? ToObject(target).
+		var to = toObject(target);
 
-			// 2. If only one argument was passed, return to.
-			if (arguments.length === 1) {
-				return to;
-			}
-
-			// 3. Let sources be the List of argument values starting with the second argument
-			var sources = Array.prototype.slice.call(arguments, 1);
-
-			// 4. For each element nextSource of sources, in ascending index order, do
-			var index1;
-			var index2;
-			var keys;
-			var key;
-			var from;
-			for (index1 = 0; index1 < sources.length; index1++) {
-				var nextSource = sources[index1];
-				// 4a. If nextSource is undefined or null, let keys be a new empty List.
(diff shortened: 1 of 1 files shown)


polyfills/Object/assign/polyfill.js
```diff
@@ -1,68 +1,65 @@
...
 		}
-    return Object(argument);
-  }
+		return Object(argument);
+	}
 
-	Object.defineProperty(Object, 'assign', {
-		// 19.1.2.1 Object.assign ( target, ...sources )
-		value: function assign(target, source) { // eslint-disable-line no-unused-vars
+	Object.assign = function assign(target, source) {
```
JakeChampion, line 10: This will also make it enumerable, it should not be enumerable.

JakeChampion:
Thanks for taking the time to create the pull-request. We already have a [pull-request](https://github.com/Financial-Times/polyfill-service/pull/1560) which fixes this issue.

2ECC71:
@JakeChampion Is there a timeframe for fixing this? Huge issue for me.

JakeChampion:
@loxtank It is being review now and we will make a release this week

loxtank:
@JakeChampion ok, is there anyway to get an old version in the meantime?

prothweil:
Is there a possible workaround until a fix is released?

JakeChampion:
@prothweil Yes, adding `&excludes=Object.assign` to the end of your polyfill.io script url will fix the issue for you.

TobiasBales:
At least for us that broke with `Object.assign` is not a function in ie, even though create-react-app ships it's own polyfill. We reverted immediately but our assumption was, that it broke inside the polyfills (polyfills.io ones) before the other one got loaded. Not 100% sure though.

JakeChampion:
@TobiasBales Yes, if you load polyfill.io before loading your own Object.assign polyfill, that would likely happen

TobiasBales:
Yeah and it makes sense, just wanted to point that out.
And loading them in reverse order my break too since then you code might be missing polyfills and already executing.

PR #1664, opened 2018-02-12
Improve number.parseint polyfill and tests

JakeChampion:


JakeChampion:
Moved to https://github.com/Financial-Times/polyfill-library/pull/81

PR #1684, opened 2018-03-15
Update docs.js ---- use req.protocol instead of fixed 'https'

xboy2012:
use req.protocol instead of fixed 'https'


xboy2012:
I open http://localhost:3000/v2/docs in my browser
But the code uses 'https:' which causes the url become

https://localhost:3000/v2/docs

It's obviously wrong url.
After inspecting the code. I found where the problem is.

Please merge the request.

JakeChampion (approved):

JakeChampion:
Closed via #1694

PR #1687, opened 2018-03-24
Add option to skip mounting tests

gtd:
What do you think about giving the option not to mount the tests?

This came up because I have security scanners that are flagging the tests for XSS issues.  I haven't dug in to assess the specifics of those vectors, but in any case, I don't really want people running the tests on my production instance anyway.

Would this option be acceptable to you?

### DIFF
--- a/service/index.js
+++ b/service/index.js
@@ -65,7 +65,10 @@ app.use((req, res, next) => {
 
 app.use(require('./routes/api.js'));
 app.use(require('./routes/meta.js'));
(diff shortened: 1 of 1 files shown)


JakeChampion (approved):

JakeChampion:
Thanks for the pull-request! This is a sensible idea, I imagine lots of self-hosting users would prefer not to have the test-runner mounted.

gtd:
Thanks for the quick turn around!

PR #1697, opened 2018-04-10
serve `URL` polyfill for BlackBerry devices

mcfog:
According to caniuse, it's not supported on all blackberry devices.  https://caniuse.com/#feat=url

### DIFF
--- a/packages/polyfill-library/polyfills/URL/config.json
+++ b/packages/polyfill-library/polyfills/URL/config.json
@@ -18,6 +18,7 @@
 		"op_mini": "*",
 		"android": "*",
 		"samsung_mob": "<5",
+		"bb": "*",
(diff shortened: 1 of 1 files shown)


JakeChampion:
@mcfog Are you able to prove that this polyfill works on BlackBerry devices at all? A picture of a BlackBerry device loading the polyfill and passing all the tests in the test suite would be the best way to prove it is working :-)

mcfog:
@JakeChampion  Sorry but I don't have such device in hand. I just find some related error log on blackberry, and fixed them (at least the error log) by adding `URL|gated|always` in features parameter

JakeChampion:
This has moved to https://github.com/Financial-Times/polyfill-library/pull/67

PR #1699, opened 2018-04-13
polyfill Set for samsung <3

skv-headless:
based on our error reports


JakeChampion:
@skv-headless Thanks for the pull-request. Unfortunately we can not merge this as we do not offer support for Samsung Browser versions older than version 4. The reason we do not offer support for older Samsung Browser versions is because it would be too much of a maintenance burden manually testing across each version whenever a change to a polyfill is made.

skv-headless:
Thank for the answer

PR #1701, opened 2018-04-15
Update Yaku version to support Promise polyfilling in Web Worker

boriskor:
Hi! First of all, thank you very much for a great service!
I've encountered a problem while trying to load the `polyfill.js` inside of Web Worker of IE11 (`importScripts`). While running the Promise polyfill, it crashed, because the Web Worker was not identified as a browser and it ran the code containing `process.nextTick`. It was fixed in https://github.com/ysmood/yaku/pull/57 and released in `yaku@0.18.6` =)

### DIFF
--- a/packages/polyfill-library/package-lock.json
+++ b/packages/polyfill-library/package-lock.json
@@ -4835,9 +4835,9 @@
       }
     },
     "yaku": {
-      "version": "0.17.11",
-      "resolved": "https://registry.npmjs.org/yaku/-/yaku-0.17.11.tgz",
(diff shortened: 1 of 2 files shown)


JakeChampion (approved):
Thanks for the pull request @boriskor!

boriskor:
Thanks, @JakeChampion! When do you plan to release v3.26? 🙂

gozno42:
@JakeChampion I just ran into this issue.  When will 3.26 be released?

PR #1717, opened 2018-05-09
Update Promise.prototype.finally browser versions

Gidgidonihah:
Currently many of the latest browsers support `finally` natively, but the polyfill is still setup to be used.

As of this writing, [MDN[(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally_ is outdated on Safari, but [Kangax](http://kangax.github.io/compat-table/es2016plus/#test-Promise.prototype.finally) is up to date on ` Promise.prototype.finally ` browser support.


JakeChampion:
This has moved to https://github.com/Financial-Times/polyfill-library/pull/66

PR #1730, opened 2018-06-15
Add cache for file/dir reader

sleagon:
High io may occur under high concurrency, a basic lru-cache may be useful.


sleagon:
@johnmichel could any guy just take a look?  a simple cache for file reader added.

PR #1732, opened 2018-06-16
Add DOMRect polyfill

brandonpayton:
This is a PR to add a simple DOMRect polyfill.

Spec: https://www.w3.org/TR/geometry-1/#DOMRect
MDN: https://developer.mozilla.org/en-US/docs/Web/API/DOMRect

We currently need a DOMRect polyfill for IE/Edge support in WordPress' new [block-based approach to editing](https://github.com/WordPress/gutenberg/), and it seems better to submit it to a polyfill library we appreciate than to maintain it privately.

### DIFF
--- a/packages/polyfill-library/polyfills/DOMRect/config.json
+++ b/packages/polyfill-library/polyfills/DOMRect/config.json
@@ -1,0 +1,18 @@
+{
+	"aliases": [],
+	"browsers": {
+		"ie": "*",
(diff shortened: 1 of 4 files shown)


brandonpayton:
Hi @JakeChampion, would you be up for taking a look at this?

JakeChampion:
Moved to https://github.com/Financial-Times/polyfill-library/pull/82

PR #1738, opened 2018-07-12
Add undefined check to Symbol setDescriptor

troys-code:
In an IE 11 environment, the `polyfill.js` file was throwing the error `Invalid descriptor for property '0'`, in the Symbol polyfill: https://github.com/Financial-Times/polyfill-service/blob/8717a9e04ac7aff99b4980fbedead98036b0929a/packages/polyfill-library/polyfills/Symbol/polyfill.js#L227

This line is being called with these arguments:
```javascript
defineProperty(ObjectProto, '0', undefined); // should not be called with undefined
```

In this pull request I have added an `undefined` check before calling `defineProperty`.

### DIFF
--- a/packages/polyfill-library/polyfills/Symbol/polyfill.js
+++ b/packages/polyfill-library/polyfills/Symbol/polyfill.js
@@ -223,7 +223,7 @@
 		var protoDescriptor = gOPD(ObjectProto, key);
 		delete ObjectProto[key];
 		defineProperty(o, key, descriptor);
-		if (o !== ObjectProto) {
+		if (o !== ObjectProto && protoDescriptor) {
 			defineProperty(ObjectProto, key, protoDescriptor);
 		}
 	};


JakeChampion:
@troy-lamerton Thanks for the pull-request! Could you explain/show how you came across this situation please? I'm surprised that this hasn't been an issue before. 
If you could post a small reproducible test case on a site such as https://codepen.io or https://jsbin.com that would help us confirm what exactly is causing the issue. The test case should load only the https://polyfill.io script, and then any code required to trigger the problem (which should hopefully be no more than a few lines).

Again, thanks for taking the time to open a pull-request, it is great to see help from the community :-)

troys-code:
The call stack when the page hits the error:
screen shot 2018-07-18 at 12 07 06

The erroneous function is called by the `DOMTokenList` polyfill. 

Part of the way through the callstack I found this:
```javascript
var $defineProperty = function defineProp(o, key, descriptor) {
  var uid = '' + key;
  if (onlySymbols(uid)) { // onlySymbols returns true, and enters this block -- is that intended?
    setDescriptor(o, uid, descriptor.enumerable ?
      copyAsNonEnumerable(descriptor) : descriptor);
    addInternalIfNeeded(o, uid, !!descriptor.enumerable);
  } else {
    defineProperty(o, key, descriptor);
  }
  return o;
};
```
For the property, onlySymbols returns true, is that intended?

I'm not sure how else to narrow down this error. The project that hits this error is big and difficult to seperate.

JakeChampion:
Hi @jiayihu, I'm struggling to replicate the issue myself.

If you could post a small reproducible test case on a site such as https://codepen.io or https://jsbin.com that would help us confirm what exactly is causing the issue. The test case should load only the https://polyfill.io script, and then any code required to trigger the problem (which should hopefully be no more than a few lines).

JakeChampion:
The code that this PR was changed has been moved to https://github.com/Financial-Times/polyfill-library

PR #1750, opened 2018-08-15
fix(polyfills/intl): enable for IE11

buschtoens:
IE11 actually does not fully support `Intl`.

For instance, [`DateTimeFormat#formatToParts`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/formatToParts#Syntax) is not available in IE11, but is in Chrome and Firefox. It is also included in the polyfill.

This PR therefore updates the polyfillable range for IE to `9 - 11` (from `9 - 10`) and for IE mobile to `10 - 11` (from just `10`).

I hereby agree to the [Contribution Terms](https://cdn.polyfill.io/v2/docs/contributing#contribution-terms).

### DIFF
--- a/packages/polyfill-library/polyfills/Intl/config.json
+++ b/packages/polyfill-library/polyfills/Intl/config.json
@@ -1,8 +1,8 @@
 {
 	"browsers": {
 		"android": "<=4.3",
-		"ie": "9 - 10",
-		"ie_mob": "10",
+		"ie": "9 - 11",
+		"ie_mob": "10 - 11",
 		"firefox": "<29",
 		"opera": "<15",
 		"chrome": "<=34",


buschtoens:
Damn. Merging this won't have any effect just yet, because [`Intl.js` itself is gated](https://github.com/andyearnshaw/Intl.js/blob/master/src/browser-main.js) and "incorrectly" does not patch missing methods, if `Intl` is present: 
https://github.com/andyearnshaw/Intl.js/issues/260

Also relevant: https://github.com/andyearnshaw/Intl.js/issues/242

We could work around this with executing `delete window.Intl` before including the polyfill in the IE11 version.

buschtoens:
I can confirm that the following code successfully loads the polyfill in IE11, including among others [`DateTimeFormat#formatToParts`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/formatToParts#Syntax):

```html
<script>delete window.Intl;</script>
<script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=Intl.~locale.de-DE|always"></script>

```

buschtoens:
So until andyearnshaw/Intl.js#260 or andyearnshaw/Intl.js#242 are fixed, I guess serving the 
`delete window.Intl;` flavor of the polyfill to IE11 is the best / only solution.

buschtoens:
There _is_ another solution: [tc39/proposal-intl-formatToParts](https://github.com/tc39/proposal-intl-formatToParts) has a polyfill for just `formatToParts`, that patches the native `Intl` implementation.

So instead of forcing the full `Intl` polyfill onto IE11 users, we could add this polyfill.

JakeChampion:
Closing this as code has moved to https://github.com/Financial-Times/polyfill-library/
I will make an issue for adding `DateTimeFormat.prototype.formatToParts` and `NumberFormat.prototype.formatToParts` 👍 
Issue - https://github.com/Financial-Times/polyfill-library/issues/109

buschtoens:
Thanks for transferring! :)

RichiCoder1:
@buschtoens I'd love to push a fix forward if you'd still like to do this. My comment on the new issue here: https://github.com/Financial-Times/polyfill-library/issues/109#issuecomment-464142663

PR #1754, opened 2018-08-23
feat: do not overwrite the flags of features(&alias)

sleagon:
the polyfill-liabrary will always merge the flags from feature and alias:

a query like this: ```default|gated,Promise|always``` , it is same as ```default|gated,Promise|always|gated```. Actually, i do not want the gated flag, i want to overwrite Promise anyway. There is no way to do things like this. I think we should not merge the flags if there is already flags assigned to feature (like, Promise).

### DIFF
--- a/packages/polyfill-library/lib/aliases.js
+++ b/packages/polyfill-library/lib/aliases.js
@@ -69,7 +69,11 @@ async function applyResolverToIdentifiers(featureSet, nameResolverFunction) {
 					// Merge properties of source into properties of target
 					// Eg. 'es6|always' becomes 'Set|always', 'Map|always' etc.
 					// Note: optimisation here is important, a previous slower merge technique used to block the event loop for long periods
-					['flags', 'aliasOf'].forEach(prop => {
+					// Note: do not overwrite flags of feature
+					if (!featureSet[newName].flags || !featureSet[newName].flags.size) {
+						featureSet[newName].flags = props.flags;
+					}
+					['aliasOf'].forEach(prop => {
 						for (const v of props[prop]) {
 							if (!(prop in featureSet[newName])) {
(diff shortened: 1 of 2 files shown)


JakeChampion:
Hi @sleagon, thank you for the pull-request. After speaking with the team, we can't figure out what your need is. 

May I ask why you are wanting to always apply `Promise` polyfill to all browsers? Is there a bug in some browsers native `Promise` and we are not serving those browsers the polyfill?

If we do add this feature as is, it would be a breaking change and require a new `/v4` endpoint. Do you think there is a way we can add this feature without it being a breaking change?

sleagon:
> Hi @sleagon, thank you for the pull-request. After speaking with the team, we can't figure out what your need is.
> 
> May I ask why you are wanting to always apply `Promise` polyfill to all browsers? Is there a bug in some browsers native `Promise` and we are not serving those browsers the polyfill?
> 
> If we do add this feature as is, it would be a breaking change and require a new `/v4` endpoint. Do you think there is a way we can add this feature without it being a breaking change?

Promise is natively supported in firefox, however window.onunhandledrejection which depends on implement of Promise is not. You want to catch the unhandledrejection event in firefox, you need overwrite the native promise. Promise is a simple example, the problem is when you want to add flags to some specific polyfill and use preset(s) (alias) which contains the polyfill at the same time, there is no way to do this, you have to write all the features in the url. Anyway, what do you want if you write some code like this `default|gated,Promise|always|gated`, I do think flags added to Promise should get higher priority than flags added to default. Feel free to close it anyway.

JakeChampion:
I've moved this to the new polyfill-library repository --> https://github.com/Financial-Times/polyfill-library/pull/165

PR #1768, opened 2018-09-30
Fallback to closest known user agent for mapped browsers

kaskar2008:
Service can now find a closest match for known browsers (that have been mapped)


JakeChampion:
Sorry I took so long to get to this pull-request. This file no longer exists, we have a new project which does the user-agent normalisation in both JavaScript and VCL (Varnish Configuration Language) -- https://github.com/Financial-Times/polyfill-useragent-normaliser/

I would gladly accept a pull-request there for this work. I have already opened one for the new Yandex browser aliases you added.

PR #1777, opened 2018-10-16
Beta V3 API -- Moving from Heroku to AWS+Serverless framework 

JakeChampion:
## What change have been made?
- Rebuilt the polyfill-service codebase from the ground-up, scrapping ExpressJS and using the [Serverless](https://serverless.com/) framework.
- Enabled the entire stack (DNS, AWS, Fastly) to be recreated automatically, enabling full-stack testing to be done for every pull-request.
- Created a V3 of the API. When an unknown user-agent is detected, it will return a bundle of polyfills wrapped in feature detects.
- The version of the polyfill-library can now be configured as part of the V3 API. This ensures user can stick to a particular version and/or upgrade to newer version in their own time.
- Polyfill bundles are generated once and stored in AWS S3, this ensures that when the same bundle is requested in the future, we serve the exact same content. Combining this feature with a specific version of the polyfill-library can enable the use of Sub-Resource-Integrity (SRI).
- The `/v3/normalizeUa` route is now done completely at the edge of Fastly's network instead of done at the application's origin.
- Testing against all browser versions in BrowserStack that the service officially supports. This has caught a few bugs in polyfills, which have been updated in this pull-request.
- Raised the baseline browser version support of the service, removing Firefox 3.1 and other very old browsers. 
- Website is now a static website hosted on AWS S3.
- Lots of unit and integration tests have been added to cover the codebase.
- Use of the FT's [secret-squirrel](https://github.com/Financial-Times/secret-squirrel) project to ensure no API keys or other secrets are committed into the Git repository.


chee:
LGTM 👍

packages/polyfill-library/lib/index.js
```diff
@@ -50,34 +50,29 @@ const PolyfillLibrary = class PolyfillLibrary {
 	 * @param {Array<String>} [opts.excludes=[]] - Which features should be excluded from the returned object.
 	 * @param {String} [opts.uaString=''] - The user-agent string to check each feature against.
 	 * @param {Boolean} [opts.rum=false] - Whether to add a script to the polyfill bundle which reports anonymous usage data
-	 * @return {Object} - opts merged with the defaults opts values.
+	 * @return {Object} options - opts merged with the defaults opts values.
```
JakeChampion:
```suggestion
	 * @return {Object} options - opts merged with the defaults option values.
```

packages/polyfill-library/polyfills/Array/prototype/find/config.json
```diff
@@ -7,15 +7,17 @@
 	"browsers": {
 		"bb": "*",
 		"chrome": "<=45",
-		"firefox": "3.6 - 24",
+		"edge": "*",
+		"edge_mob": "*",
+		"firefox": "<24",
```
JakeChampion, line 12:
```suggestion
		"firefox": "<25",
```

packages/polyfill-library/polyfills/Array/prototype/find/config.json
```diff
@@ -7,15 +7,17 @@
...
 		"ie": "*",
 		"ie_mob": "10 - *",
 		"opera": "<32",
 		"safari": "* - 7.1",
 		"ios_saf": "* - 8.0",
 		"ios_chr": "* - 8.0",
 		"android": "*",
-		"firefox_mob": "3.6 - 24",
+		"firefox_mob": "<24",
```
JakeChampion, line 20:
```suggestion
		"firefox_mob": "<25",
```

packages/polyfill-library/polyfills/Symbol/replace/config.json
```diff
@@ -4,7 +4,9 @@
 		"es2015"
 	],
 	"browsers": {
-		"ie": "< 13",
+		"edge": "*",
```
JakeChampion, line 7: This is likely incorrect, but I would rather serve a polyfill to a browser that doesn't need it than to not serve a polyfill to a browser that does. In a subsequent pull-request I will add a script that checks if polyfills are being served to browsers which do not require them, and I will have the script run in CircleCI, causing builds to break if it finds such a situation.

packages/polyfill-library/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "polyfill-library",
-  "version": "3.25.1",
+  "version": "0.0.0",
```
jonathantneal, line 3: These are being reset because they were never published?

JakeChampion: They were reset because the publishing workflow was going to use https://github.com/matthew-andrews/npm-prepublish#npm-prepublish. We have since moved away from this and we now use this -- https://github.com/Financial-Times/polyfill-service/tree/master/packages/polyfill-service#release-process

PR #1794, opened 2018-11-02
fix: replace object.assign by babel object.assign

ephor:
connected to #1762


JakeChampion:
The polyfill-library has been moved to it's own repo -- https://github.com/Financial-Times/polyfill-library

PR #1801, opened 2018-11-15
Update polyfill and add tests for Math.imul

JakeChampion:


JakeChampion:
Moved to https://github.com/Financial-Times/polyfill-library/pull/83

PR #1820, opened 2018-11-22
Versioned libraries

JakeChampion:
Problem:
The implementation for versioned libraries was complex and cumbersome, requiring lots of scripts to build the polyfills from previous library versions.

Solution:
Use yarn instead of npm to give us the ability to have multiple versions of the same package as a dependency under different aliases. 

`/v3/polyfill.js` will use the latest version of the polyfill-library module
`/v3/polyfill.js?version=3.25.2` will use the version `3.25.2` of the polyfill-service module (this is because version 3.25.2 is the last version where the polyfill-library and the polyfill-service were under a single module.
`/v3/polyfill.js?version=xxx` will return a 400 because we do not have a version of the polyfill-library which matches "xxx".

When we release newer versions of the polyfill-library, we can add the previous version as an alias and add it to the switch statement in `packages/polyfill-service/server/routes/v3/polyfill.js`.


package.json
```diff
@@ -5,7 +5,7 @@
   "description": "A polyfill combinator",
   "scripts": {
     "version": "npm version $npm_package_version --prefix packages/polyfill-library && npm version $npm_package_version --prefix packages/polyfill-service && git add packages/polyfill-library/package.json packages/polyfill-service/package.json",
-    "install": "npm ci --prefix packages/polyfill-library && npm ci --prefix packages/polyfill-service",
+    "install": "cd packages/polyfill-library && yarn install --frozen-lockfile && cd - && cd packages/polyfill-service && yarn install --frozen-lockfile && cd -",
```
chee, line 8: i have never seen `cd -` before. i've been using pushd/popd like a chump

packages/polyfill-service/server/routes/v3/polyfill.js
```diff
@@ -5,22 +5,58 @@ const compressBundle = require("../../lib/compress-bundle");
...
-		response.set({
-			"Cache-Control": "public, s-maxage=31536000, max-age=604800, stale-while-revalidate=604800, stale-if-error=604800",
-			"Content-Encoding": params.compression,
-			"Content-Type": "application/javascript; charset=utf-8",
-			Etag: etag,
-			"surrogate-key": "polyfill-service"
-		});
-		response.send(file);
+		switch (params.version) {
+			case latestVersion: {
```
notlee, line 26: ~Should this case build any version since `3.25.2` rather than just the latest?~

chee:
for the 46,000 line diff this is a relatively small pull request.

it doesn't look like it too heavily ties to yarn, which is nice, so leaves the option open for another aliasing approach should one ever exist.

seems p good.

are any of the [things in the switch statement](https://github.com/Financial-Times/polyfill-service/blob/26d01d0b42d73545121ef2e09a8449616dab4c00/packages/polyfill-service/server/routes/v3/polyfill.js#L14-L50) generic enough to live in a function?

packages/polyfill-service/server/routes/v3/polyfill.js
notlee: >When we release newer versions of the polyfill-library, we can add the previous version as an alias and add it to the switch statement in packages/polyfill-service/server/routes/v3/polyfill.js.

Nope, got it, I should read 😅

PR #1822, opened 2018-11-28
Set polyfill: missing opera config 

christophehurpeau:
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Set

Set is supported starting version 25


christophehurpeau:
new pr: https://github.com/Financial-Times/polyfill-library/pull/9

PR #1833, opened 2018-12-14
add AbortController

caub:
I'd like to add `AbortController`, I'm just a little confused on how to add it, it seems polyfills are inlined in https://github.com/Financial-Times/polyfill-service/blob/master/packages/polyfill-library/polyfills directories, so it must part of a script?

repo: https://github.com/mo/abortcontroller-polyfill/
polyfill: https://unpkg.com/abortcontroller-polyfill

Thanks for the pointers


caub:
@JakeChampion I added https://unpkg.com/abortcontroller-polyfill@1.2.1/dist/umd-polyfill.js (copy-pasted in polyfill.js) I'm not sure how the config.json install field would work

caub:
rebased

JakeChampion:
This is now open at https://github.com/Financial-Times/polyfill-library/pull/59

PR #1866, opened 2019-01-16
commas in the bignumber

ka-tsu-ra:


JakeChampion (approved):
Nice, I did not know Node had this function!

PR #1871, opened 2019-01-24
use the @financial-times/polyfill-useragent-normaliser module

JakeChampion:
Using this module as it contains the exact same normalisation in JS and VCL. The JS will be used by polyfill-library and the VCL will be used by polyfill-service.


chee:
([financial-times/polyfill-useragent-normaliser](/financial-times/polyfill-useragent-normaliser))

PR #1899, opened 2019-03-04
move the website to being completely static and use a static server 

JakeChampion:


src/assets/css/index.scss
```diff
@@ -1,102 +1,102 @@
...
+	@include oButtons($theme: 'mono');
+}
+
+.core .o--if-js,
+.enhanced .o--if-no-js {
+	display: none !important;
+}
+
+[aria-hidden=true] {
+	display: none !important;
```
notlee, line 46: This could have unintended consequences. Elements which are hidden to screen readers are not always intended to be visually hidden. Could this use a more specific selector instead?

JakeChampion: Fixed

PR #1921, opened 2019-03-18
Add a regression test covering Vary headers where a Fastly-FF header is present

rowanbeentje:
There was a change (https://github.com/Financial-Times/polyfill-service/blame/839da3d75256b82a23393932182cd24b59b8c3a5/fastly/vcl/polyfill-service.vcl#L193-L201) moving some logic to behind a `Fastly-FF` check to support shielding.  Unfortunately there are third-party consumers of this project (including the FT) who use Fastly as a proxy to the polyfill service, so the `Fastly-FF` header is also set there, which can cause issues.

This adds a regression test looking at one of the externally available headers which could act as a canary for this.

### DIFF
--- a/test/integration/regression/external-fastly-proxy.test.js
+++ b/test/integration/regression/external-fastly-proxy.test.js
@@ -1,0 +1,21 @@
+/* eslint-env mocha */
+
+"use strict";
+
+const request = require("supertest");
+const assert = require("proclaim");
+const host = require("../helpers").host;
+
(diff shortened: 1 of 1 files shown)


rowanbeentje:
(This could be added to https://github.com/Financial-Times/polyfill-service/pull/1919 but I don't have the access for that :) )

JakeChampion (approved):

PR #1942, opened 2019-03-27
Show aliases as part of the url builder

JakeChampion:
This is a first pass at showing the aliases on the url builder page, it could do with more ui/ux thought going into it.
image

### DIFF
--- a/src/assets/css/index.scss
+++ b/src/assets/css/index.scss
@@ -98,6 +98,7 @@ $o-tooltip-is-silent: false;
 .o-tooltip .o-tooltip-content {
 	hyphens: unset;
 	white-space: nowrap;
+	max-height: 50vh;
 
 	> :last-child {
 		margin-bottom: 8px;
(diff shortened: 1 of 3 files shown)


azizhk:
Does `alias` over here also mean dependencies?
Like will `requestAnimationFrame` have `Date.now`?

JakeChampion:
@azizhk `alias` does not include dependencies for polyfills, `alias` is this json file --> https://unpkg.com/polyfill-library@3.31.1/polyfills/__dist/aliases.json

Did you want it to show dependencies of polyfills as well?

azizhk:
> Did you want it to show dependencies of polyfills as well?

Well not my end use case per se.
My use case is that if I check `default` then `Array.prototype.filter` should automatically get checked. Like checked by default, which cannot be unchecked. Which on hover can say "Included as you have selected `default` bundle.

I know maybe too much trouble to implement 😅 . Can ignore.

JakeChampion:
@azizhk That is a good feature request, could you open a new issue for it please? :-)

PR #1962, opened 2019-05-14
Add Docker support

revolunet:
Thanks for this handy service; 

Here's a Dockerfile to run the service directly

May be nice to publish the image to dockerhub so users wont need to build the image by themself.

Thanks !

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:10
+
+WORKDIR /app
+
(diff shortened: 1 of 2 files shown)


JakeChampion:
Hi, thanks for the pull-request @revolunet, we won't be accepting this pull-request however as we do not want to take on the extra maintenance of keeping a Docker image up-to-date whenever any vulnerabilities in any of the layers within the Docker image are found.

PR #1992, opened 2019-07-18
Update useragent normaliser to version which detects WebPageTest as Chrome

JakeChampion:
Resolves https://github.com/Financial-Times/polyfill-useragent-normaliser/issues/13


chee (approved):
nice, thanks @ianfeather

PR #2000, opened 2019-07-29
Add polyfill.io script tag to make website work on older browsers

### LINKED ISSUE #1964: Add polyfill on url builder itself, attempt 2
Considering you already have the polyfill, can you add it on your page itself? Specifically, on https://cdn.polyfill.io/v3/url-builder with Chrome 55 I get the bundle url as https://polyfill.io/v3/polyfill.min.js?%5Bobject+Object%5D= regardless of what I select. I suppose you are using some feature not available in it ...

Note that https://cdn.polyfill.io/static/polyfills.js is returning 404 and is referenced on that page ... and I also see this error:
```
jquery.js:19 Refused to apply inline style because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-0wIoD60yL42+1XJUY22zM8LflSmtzRyZIjM0qasci88='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.
jquery.js:19 Refused to execute inline script because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-+1gZjPThUpVdzUAk0dH9fwK52R9/GjM+bgt7zsawajs='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.
```

Note that I already submitted this as #1873 but someone else commented on that bug and you fixed just what he reported and not my original complain.

JakeChampion:
Fixes https://github.com/Financial-Times/polyfill-service/issues/1964


chee (approved):
🎉 2000 🎊

PR #2034, opened 2019-11-08
update polyfill-library to 3.42 and add 3.40 and 3.41 as historical versions that can still be used

JakeChampion:
3.42 release notes:
Added missing Type dependency for Object.preventExtensions

Don't serve object.preventextensions to browsers which support the es2015 defined behavior
as defined from here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions#Browser_compatibility

Serve URL/URLSearchParams polyfill to Firefox versions below 44

Added element.scroll* polyfills

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,5 +94,5 @@
     "sinon": "^7.3.2",
     "supertest": "^4.0.2"
   },
-  "version": "4.13.0"
+  "version": "4.14.0"
 }
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "dotenv": "^8.0.0",
     "express-extractheaders": "^3.0.0",
     "iltorb": "^2.4.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "dotenv": "^8.0.0",
     "express-extractheaders": "^3.0.0",
     "iltorb": "^2.4.3",
-    "polyfill-library": "^3.40.0",
+    "polyfill-library": "^3.42.0",
```
chee, line 48: what happened to 3.41 !

chee (approved):

package.json
JakeChampion: It exists, we just didn't release a version of polyfill.io with it.

PR #2035, opened 2019-11-09
Update the process for reporting security bugs

### LINKED ISSUE #1864: Where can I disclose a security bug?
On the Polyfill.io site it mentions the following at https://polyfill.io/v3/report-a-bug#security-bugs.

> Polyfill.io has a bounty program for the safe disclosure of security bugs. With that in mind, please do not file public issues; go through the process outlined on that page.

But it's not actually clear _where_ the mentioned page is.

A direct link to the HackerOne setup would be helpful.

JakeChampion:
Resolves #1864


chee (approved):
it's a Saturday

PR #2074, opened 2020-01-15
Dereference the tag recursively until a non-tag object is found

JakeChampion:
so that we do not try to write a non-commit object to the master branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
       - "checkout"
       - run: "sudo npm i -g npm && npm ci"
       - run: "bash ~/project/.circleci/create_netrc_file.sh"
-      - run: "git push https://git.heroku.com/origami-polyfill-service-int.git HEAD:master --force"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - "checkout"
       - run: "sudo npm i -g npm && npm ci"
       - run: "bash ~/project/.circleci/create_netrc_file.sh"
-      - run: "git push https://git.heroku.com/origami-polyfill-service-int.git HEAD:master --force"
+      - run: "git push https://git.heroku.com/origami-polyfill-service-int.git HEAD^{}:refs/heads/master --force"
```
notlee, line 35: This looks reasonable, approved! 

_except I don't don't what the purpose of `^{}` is or when writing a tag object is a problem?_ 😳

notlee (approved):

.circleci/config.yml
JakeChampion: I think I don't need to do this actually, I think Heroku can instead reset their git repo and that should resolve the issue.

PR #2082, opened 2020-01-17
Origami major update

JakeChampion:
This upgrades all the Origami components to their latest versions and also adds a script to detect if multiple versions of a component have been installed by mistake

I also added the hero svg by @gvonkoss  because I think it is really nice :-)


src/assets/url-builder.njk
```diff
@@ -5,84 +5,78 @@ oLayoutStyle: "o-layout--query"
...
+					<input name="filter" id="filter" type="text" class="o-forms__text" placeholder="Type to filter by name">
+			</span>
+		</label>
+		<label class="o-forms-field o-forms-field--optional">
+			<span class="o-forms-title">
+				<span class="o-forms-title__main">Callback</span>
+				<span class="o-forms-title__prompt">Name of the function to call after the polyfills are loaded.</span>
+			</span>
+			<span class="o-forms-input o-forms-input--text">
+					<input name="callback" id="callback" type="text" class="o-forms__text" placeholder="Type to filter by name">
```
notlee, line 39: Doesn't look like this placeholder matches the input
```suggestion
					<input name="callback" id="callback" type="text" class="o-forms__text">
```

notlee:
Did you mean to remove the RUM, feature detect, etc options from the url builder?

notlee:
FYI looks like the spacing under the polyfill checkboxes is a little tight now too -- thanks percy!

JakeChampion:
> Did you mean to remove the RUM, feature detect, etc options from the url builder?

Yes I did, we don't support RUM at all 😂 and the other flags are really to not be used unless debugging so removing them from the url builder and keeping them in the api docs page hopefully will prevent people from using them on production websites because they really shouldn't be used in production

JakeChampion:
> FYI looks like the spacing under the polyfill checkboxes is a little tight now too -- thanks percy!

Yeah I saw that, not sure what introduced it though

PR #2129, opened 2020-02-18
append tooltip to body

JakeChampion:
Without this, the tooltip is not visible:

before:
image

after:
image


chee (approved):

notlee:
Sorry this is because I added `overflow: hidden` on o-layout. I'll remove it in case anyone else has a similar issue. It's not needed for the Chrome 80 fix

Edit: oh, actually I already did this https://github.com/Financial-Times/o-layout/pull/130/files

PR #2147, opened 2020-02-26
Add the Intl locales to the url builder page

### LINKED ISSUE #1911: Intl and Locale Documentation missing in v3 website
Feature Request

## What

Intl and Locale Documentation missing in v3 website

## Details

There used to be a line in v2 docs: 
> Locales must be specified separately by prefixing the locale name with Intl.~locale, eg Intl.~locale.en-GB.

Might help people looking for it. I can add it, if you can tell what would be a good place.

JakeChampion:
This is one way to resolve https://github.com/Financial-Times/polyfill-service/issues/1911

If someone has another solution to this problem, please comment on this PR :-)
image

### DIFF
--- a/src/data/polyfills.js
+++ b/src/data/polyfills.js
@@ -38,7 +38,7 @@ module.exports = async () => {
 
 		for (const polyfill of await polyfillLibrary.listAllPolyfills()) {
 			// Polyfills which start with _ are internal functions used by other polyfills, they should not be displayed on the website.
-			if (!polyfill.startsWith("_") && !polyfill.startsWith("Intl.~locale")) {
+			if (!polyfill.startsWith("_")) {
 				const polyfillInfo = Object.assign(
 					{
 						name: polyfill,


chee:
a long cat with all the internationalisation settings

notlee:
Seems like the old solution was reasonable, to show one result with a little note
Screenshot 2020-02-26 at 13 54 32

JakeChampion:
@notlee It was okay but it did not list all the locales the user can choose from, they would have to hazard a guess

notlee:
> @notlee It was okay but it did not list all the locales the user can choose from, they would have to hazard a guess

How many are there? I wonder if they could be listed in the note. Not searchable but avoids the long long list so other available polyfills can be scanned

JakeChampion:
> > @notlee It was okay but it did not list all the locales the user can choose from, they would have to hazard a guess
> 
> How many are there? I wonder if they could be listed in the note. Not searchable but avoids the long long list so other available polyfills can be scanned

716

chee:
Could we have it show as something like `Intl.PluralRules.~locale.LOCALE` and have the word `LOCALE` be bold and the hover note says `For valid locales see <the supported locales page>` and we have another page that lists all the valid locales at like `https://polyfill.io/v3/supported-locales`

chee:
Actually, that does kind of interrupt the ability to use the polyfill builder.
What about if we have a `▶ Intl.PluralRules.~locale` expander, and all the locale checkboxes appear loaded in when you toggle the expander?

PR #2148, opened 2020-02-26
Replace custom VCL logic with new Fastly features

JakeChampion:
Replace custom solution for figuring out whether it is the first time the request has entered the service with the new Fastly provided solution


chee (approved):
Let's make sure to monitor this as it goes out

PR #2236, opened 2020-05-13
Move to Docker deployments for Heroku

JakeChampion:
Our Heroku deployments takes 15-20 minutes to happen for each application. Because of our multi-region setup making deployments sequentially (and each deployment builds the artefact again), our deploy times take twice as long, 30-40 minutes.

If we move to deploying a prebuilt artefact, the deploys should happen in a few minutes each, whilst the building of the artefact will be several minutes but only need to happen once.

We could also look at distributing a polyfill-service docker image to the DockerHub public registry to allow self-hosting to happen via Docker for those users which use Docker :-)

### DIFF
--- a/.github/workflows/deploy-to-dev-and-test.yml
+++ b/.github/workflows/deploy-to-dev-and-test.yml
@@ -15,9 +15,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: git fetch --prune --unshallow
-      - uses: actions/setup-node@v1
-        with:
-          node-version: 12.x
       - name: 'Terraform Format'
         uses: hashicorp/terraform-github-actions@v0.8.0
         with:
@@ -29,6 +26,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Remove the staging and production terraform configuration overrides
         run: rm -f terraform.tfstate fastly/terraform/qa_override.tf fastly/terraform/production_override.tf fastly/terraform/domains_override.tf
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - run: npm ci
       - name: Login to Heroku Container registry
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
--- a/.github/workflows/deploy-to-dev-and-test.yml
+++ b/.github/workflows/deploy-to-dev-and-test.yml
@@ -33,10 +33,12 @@ jobs:
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
         run: heroku container:login
-      - name: Build and push
+      - name: Build Docker image
+        run: docker build . --tag registry.heroku.com/origami-polyfill-service-int/web
+      - name: Push Docker image to Heroku
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
-        run: heroku container:push -a origami-polyfill-service-int web
+        run: docker push registry.heroku.com/origami-polyfill-service-int/web
       - name: Release
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
--- a/.github/workflows/deploy-to-dev-and-test.yml
+++ b/.github/workflows/deploy-to-dev-and-test.yml
@@ -29,7 +29,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Remove the staging and production terraform configuration overrides
         run: rm -f terraform.tfstate fastly/terraform/qa_override.tf fastly/terraform/production_override.tf fastly/terraform/domains_override.tf
-      - name: Deploy to Heroku
       - name: Login to Heroku Container registry
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
--- a/.github/workflows/deploy-to-dev-and-test.yml
+++ b/.github/workflows/deploy-to-dev-and-test.yml
@@ -18,7 +18,6 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 12.x
-      - run: npm ci
       - name: 'Terraform Format'
         uses: hashicorp/terraform-github-actions@v0.8.0
         with:
@@ -31,7 +30,18 @@ jobs:
       - name: Remove the staging and production terraform configuration overrides
         run: rm -f terraform.tfstate fastly/terraform/qa_override.tf fastly/terraform/production_override.tf fastly/terraform/domains_override.tf
       - name: Deploy to Heroku
-        run: git push https://heroku:${{ secrets.HEROKU_AUTH_TOKEN }}@git.heroku.com/origami-polyfill-service-int.git HEAD:master --force
+      - name: Login to Heroku Container registry
+        env: 
+          HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
+        run: heroku container:login 
+      - name: Build and push
+        env:
+          HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
+        run: heroku container:push -a origami-polyfill-service-int web 
+      - name: Release
+        env:
+          HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
+        run: heroku container:release -a origami-polyfill-service-int web 
       - name: 'Terraform Init'
         uses: hashicorp/terraform-github-actions@v0.8.0
         with:
--- a/.github/workflows/deploy-to-dev-and-test.yml
+++ b/.github/workflows/deploy-to-dev-and-test.yml
@@ -27,17 +27,17 @@ jobs:
         run: rm -f terraform.tfstate fastly/terraform/qa_override.tf fastly/terraform/production_override.tf fastly/terraform/domains_override.tf
       - name: Deploy to Heroku
       - name: Login to Heroku Container registry
-        env: 
+        env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
-        run: heroku container:login 
+        run: heroku container:login
       - name: Build and push
         env:
           HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
-        run: heroku container:push -a origami-polyfill-service-int web 
+        run: heroku container:push -a origami-polyfill-service-int web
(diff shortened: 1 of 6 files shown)


chee:
this is a great idea!

Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM node:12-alpine AS build
+# packages needed for node-gyp and dupe-krill
+RUN apk add --update --no-cache python make g++ rust cargo
+# use a non-root user to make commands such as npm post-install scripts safer
+USER node
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PATH=$PATH:/home/node/.npm-global/bin:/home/node/.cargo/bin
+# RUN cargo install dupe-krill
+# copy by default will make the root user own the files, we want the node user to own the files
+COPY --chown=node:node package*.json .snyk /home/node/app/
```
notlee, line 10: I need to understand users/groups better

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR /home/node/app
+RUN npm install
+COPY --chown=node:node . /home/node/app/
+RUN npm run build
+RUN npm prune --prod
+# hardlink duplicate files to save space (saves 50% for polyfill-service)
+# RUN dupe-krill .
```
notlee, line 17: I'm not sure what these comments refer to?

test/.dockerignore
```diff
@@ -0,0 +1 @@
+node_modules
```
notlee, line 1: wouldn't the docker image be able to ignore everything under test?

notlee (approved):
looks great, nice work :D

.github/workflows/deploy-to-dev-and-test.yml
```diff
@@ -30,8 +26,24 @@ jobs:
...
-      - name: Deploy to Heroku
-        run: git push https://heroku:${{ secrets.HEROKU_AUTH_TOKEN }}@git.heroku.com/origami-polyfill-service-int.git HEAD:master --force
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - run: npm ci
+      - name: Login to Heroku Container registry
+        env:
+          HEROKU_API_KEY: ${{ secrets.HEROKU_AUTH_TOKEN }}
+        run: heroku container:login
```
jamesarosen, line 36: Do you need `uses: "actions/heroku@master"` for this command to work?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PATH=$PATH:/home/node/.npm-global/bin:/home/node/.cargo/bin
+# RUN cargo install dupe-krill
+# copy by default will make the root user own the files, we want the node user to own the files
+COPY --chown=node:node package*.json .snyk /home/node/app/
+WORKDIR /home/node/app
+RUN npm install
+COPY --chown=node:node . /home/node/app/
+RUN npm run build
+RUN npm prune --prod
```
jamesarosen, line 15: There is a `FROM` below, so it _probably_ doesn't matter, but you may consider combining these two `RUN` statements.

See https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-what-is-better

Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM node:12-alpine AS build
+# packages needed for node-gyp and dupe-krill
+RUN apk add --update --no-cache python make g++ rust cargo
+# use a non-root user to make commands such as npm post-install scripts safer
+USER node
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PATH=$PATH:/home/node/.npm-global/bin:/home/node/.cargo/bin
```
jamesarosen, line 7: You could use `NPM_CONFIG_PREFIX` when computing `PATH`.

Dockerfile
jamesarosen: When I run this, `/home/node/` exists, but `/home/node/app/` doesn't at this point.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# RUN cargo install dupe-krill
+# copy by default will make the root user own the files, we want the node user to own the files
+COPY --chown=node:node package*.json .snyk /home/node/app/
+WORKDIR /home/node/app
+RUN npm install
+COPY --chown=node:node . /home/node/app/
+RUN npm run build
```
jamesarosen, line 14: The `build` task uses [`run-s`](https://github.com/mysticatea/npm-run-all/blob/master/docs/run-s.md) to be on the path. I don't see where you've installed it.

jamesarosen: It also depends on the `production:eleventy` task, which requires [`eleventy`](https://github.com/11ty/eleventy) to be on the path.

Dockerfile
jamesarosen: You'll need to copy `.eleventy.js`, `src/`, and `node_modules/` as well to get `production:eleventy` to run.

Dockerfile
jamesarosen: `.eleventy.js` depends on `fast-glob`, which isn't listed in the dependencies. #2237 will resolve that.

JakeChampion: they are installed on like 12 `RUN npm install`

Dockerfile
JakeChampion: Thanks, that's a great piece of information, I'll do it :-)

.github/workflows/deploy-to-dev-and-test.yml
JakeChampion: Heroku is installed on the GitHub Actions machine by default :-)

Dockerfile
JakeChampion: That happens on like 13 ` COPY --chown=node:node . /home/node/app/ `

Dockerfile
jamesarosen: `npm-run-all` (the thing that defines `run-s`) isn't in the project's `devDependencies`. Even if it were, `npm install` would install it to `./node_modules/` and its executable to `./node_modules/.bin/`, not to `/home/node/.npm-global/bin`. `npm install -g npm-run-all` would fix it. So would adding it to `devDependencies` and running it as `./nodes_modules/.bin/run-s …`. For that matter, so would adding `./node_modules/.bin/` to `$PATH`.

`@11ty/eleventy` is in `devDependencies`, but the rest of the above applies to its executable `eleventy` as well.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apk add --update --no-cache python make g++ rust cargo
+# use a non-root user to make commands such as npm post-install scripts safer
+USER node
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PATH=$PATH:/home/node/.npm-global/bin:/home/node/.cargo/bin
+# RUN cargo install dupe-krill
+# copy by default will make the root user own the files, we want the node user to own the files
+COPY --chown=node:node package*.json .snyk /home/node/app/
+WORKDIR /home/node/app
+RUN npm install
```
jamesarosen, line 12: When I build this Dockerfile locally, this installs `polyfill-library-3.27.4`, but `node_modules/polyfill-library-3.27.4/polyfills/__dist/` doesn't exist, which causes `npm run build` to fail:
```
Step 11/12 : RUN find node_modules -name "__dist" | sort
 ---> Running in cdab82d1db70
node_modules/polyfill-library-3.25.3/polyfills/__dist
node_modules/polyfill-library-3.34.0/polyfills/__dist
node_modules/polyfill-library-3.35.0/polyfills/__dist
node_modules/polyfill-library-3.36.0/polyfills/__dist
node_modules/polyfill-library-3.37.0/polyfills/__dist
node_modules/polyfill-library-3.38.0/polyfills/__dist
node_modules/polyfill-library-3.39.0/polyfills/__dist
node_modules/polyfill-library-3.40.0/polyfills/__dist
node_modules/polyfill-library-3.41.0/polyfills/__dist
node_modules/polyfill-library-3.42.0/polyfills/__dist
node_modules/polyfill-library-3.43.0/polyfills/__dist
node_modules/polyfill-library-3.44.0/polyfills/__dist
node_modules/polyfill-library-3.45.0/polyfills/__dist
node_modules/polyfill-library-3.46.0/polyfills/__dist
node_modules/polyfill-library-3.48.0/polyfills/__dist
node_modules/polyfill-library-3.49.0/polyfills/__dist
node_modules/polyfill-library-3.50.2/polyfills/__dist
node_modules/polyfill-library-3.51.0/polyfills/__dist
node_modules/polyfill-library-3.52.0/polyfills/__dist
node_modules/polyfill-library-3.52.1/polyfills/__dist
node_modules/polyfill-library-3.52.2/polyfills/__dist
node_modules/polyfill-library-3.52.3/polyfills/__dist
node_modules/polyfill-library-3.53.1/polyfills/__dist
node_modules/polyfill-library-3.89.4/polyfills/__dist
node_modules/polyfill-library/polyfills/__dist
```

This is also true for `3.25.1` and `3.28.1`. I opened Financial-Times/polyfill-library#639 for the issue. I still haven't quite narrowed down the cause.

jamesarosen: I think the problem is something to do with system permissions. `npm install --unsafe-perm` seems to fix it!

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY --chown=node:node . /home/node/app/
+RUN npm run build
+RUN npm prune --prod
+# hardlink duplicate files to save space (saves 50% for polyfill-service)
+# RUN dupe-krill .
+
+FROM node:12-alpine
+COPY --chown=node:node --from=build /home/node/app /home/node/app/
+USER node
+EXPOSE 8080
```
jamesarosen, line 22: Heroku's [Docker docs](https://devcenter.heroku.com/articles/container-registry-and-runtime) say

> EXPOSE - While EXPOSE can be used for local testing, it is not supported in Heroku’s container runtime. Instead your web process/code should get the $PORT environment variable.

I guess `@financial-times/origami-service` will use 8080 in local development and `$PORT` in Heroku. Nice!

test/.dockerignore
jamesarosen: And `fastly/`

Dockerfile
JakeChampion: the like 12 runs npm install which installs devdependencies and the build command is called via `npm run` which adds `./node_modules/.bin/ to the PATH, which is why this is working corectly

jamesarosen: I just tried again with your latest and no local changes. It works. I don't know what I was doing wrong. Resolved. 👍

Dockerfile
jamesarosen: I just tried again with your latest and no local changes. It works. I don't know what I was doing wrong. Resolved. 👍

Issue #2253, opened 2020-05-29
A feature request: possibility to add own basepath to url router?


maapteh:
Awesome service.

We want to use it in another route path. So not snip.com/v3/polyfill.js but something like snip.com/foo/polyfill/polyfill.js because we deploy inside container but cant enroll on just /v3

I see the routes are now set here:
https://github.com/Financial-Times/polyfill-service/blob/master/server/routes/v3/polyfill.js#L74

```
app.get(["/v3/polyfill.js", "/v3/polyfill.min.js"], async (request, response, next) => {
```

Now i can make it an optional ENV var, but would like to know if FT is willing to see this as a feature request to start with :)

Maarten

JakeChampion:
@maapteh good feature request, we would definitely accept it

maapteh:
Thanks, i added the PR. Tested the app with and without (default) process.env.CONTEXT_PATH being set.

JakeChampion:
resolved via https://github.com/Financial-Times/polyfill-service/pull/2254

PR #2254, opened 2020-05-29
feature(context): add path to url exposed

maapteh:
PR for setting own context:

- foo.com/v3/polyfill.js is the default, so non breaking for sure :)

In your container you can now set ENV variable (will be picked up in polyfill route)
`ENV CONTEXT_PATH = '/bar' `

This now will expose it on:
- foo.com/bar/v3/polyfill.js

i choose for docs to leave out

belongs to #2253

testing fork now in own k8s container


JakeChampion (approved):

JakeChampion (approved):

maapteh:
awesome @JakeChampion ! Thanks! Do you perhaps know where:

`polyfill-service configured (graphite=false logging=true sentry=false)`

comes from? i want to set logging too :)

JakeChampion:
It comes from the origami-service package which is used to create the ExpressJS server, here is the documention for how to configure it :-) -- https://github.com/Financial-Times/origami-service/#options

PR #2279, opened 2020-06-25
Correct robots.txt syntax

chenrui333:
Just found out that your robots.txt syntax is not correct. And I have referenced it with my VCL code too. 

Thus, filing this PR to fix it.

### DIFF
--- a/fastly/vcl/synthetic-responses.vcl
+++ b/fastly/vcl/synthetic-responses.vcl
@@ -84,7 +84,7 @@ sub vcl_error {
 		set obj.response = "OK";
 		set obj.http.Content-Type = "text/plain; charset=utf-8";
 		synthetic {"User-agent: *
-Disallow:"};
+Disallow: /"};
 		return (deliver);
 	}
 
--- a/fastly/vcl/synthetic-responses.vcl
+++ b/fastly/vcl/synthetic-responses.vcl
@@ -84,7 +84,7 @@ sub vcl_error {
 		set obj.response = "OK";
 		set obj.http.Content-Type = "text/plain; charset=utf-8";
 		synthetic {"User-agent: *
-Disallow: /"};
+Allow: /"};
 		return (deliver);
 	}
 


chee:
thank you !

chee:
@Financial-Times/origami-core it looks like outside contributions can't be deployed to heroku staging?

JakeChampion:
Yes, outside contributions do not get access to secrets in our GitHub CI, they didn't get them in CircleCI either. We need to review the code and check it doesn't do anything bad like steal our secrets or leak them, once that is done we then need to push the commits to our origin and then CI will run with our secrets

chenrui333:
This is actually a very simple robots.txt syntax fix. :D

JakeChampion (commented):
We want robots to crawl our website which is why the disallow is empty. I think the change suggested here would make robots not crawl any page of our website.

chenrui333:
> We want robots to crawl our website which is why the disallow is empty. I think the change suggested here would make robots not crawl any page of our website.

Yeah, you are right, I thought you want to disallow the web crawlers for the polyfill service though.

If you want to allow every UA, then you should use:

```
User-agent: *
Allow: /
```

The syntax there is just incorrect.

chenrui333:
@JakeChampion, sorry, you are right. (just did a test with https://en.ryte.com/free-tools/robots-txt/?refresh=1&url=Polyfill.io&useragent=Googlebot&submit=Evaluate)

```
User-agent: *
Disallow:
```

is equivalent to 

```
User-agent: *
Allow: /
```

I still feel the former syntax is little bit weird though.

chee:
aren't they also both equivalent to not having a `robots.txt` at all?

chee:
anyway, it looks like this is working as expected. thank you for looking out for us, @chenrui333

chenrui333:
> aren't they also both equivalent to not having a `robots.txt` at all?

yeah, I think so.

chenrui333:
> anyway, it looks like this is working as expected. thank you for looking out for us, @chenrui333

np at all, happy to help.

PR #2457, opened 2020-11-27
feat(polyfill.js): add the bundle cache

### LINKED ISSUE #2452: Can I use the cache in self-hosting environment?
## What

Can I use the cache in self-hosting environment?
## Details

The server uses the UserAgent and Request URL QueryString to generate a key that can identify the request and stores the response data along with the key, and the response time is greatly reduced when the server responds with the stored data.

I think the implementation for this cache feature is very simple, but I haven't been able to find any implemented function. 
Can i use the cache in the self-hosting environment?

stegano:
Add cache to improve responsive performance and reduce load on self-hosting

related: #2452


stegano:
Hi, @JakeChampion 

Could you please give me some advice on how to test this feature? 😰

I implemented the function in my browser and conducted a simple test.
(I checked that the normal responseBody data and response time reduced after changing the URL and User-Agent in the browser.)

PR #2497, opened 2021-01-14
update to latest version of polyfill-library

JakeChampion:
This version fixes a bug in the Intl.PluralRules polyfill where it was throwing an error when it shouldn't.


JakeChampion:
@aronwoost, once this is merged, we will cut a release to production 👍

PR #2565, opened 2021-03-22
Update to latest version of terraform and terraform plugins and use a dynamic vcl block to reduce the amount of repetition in the terraform config

JakeChampion:
When working on this project I noticed that the versions of terraform and the terraform fastly prodiver plugin were quite old.

I've updated the project to use the latest versions and have made use of a new terraform feature known as dynamic blocks to help reduce on some repetition when defining the vcl files.

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -3,7 +3,7 @@ on:
   release:
     types: [created]
 env:
-  terraform_version: '0.12.29'
+  terraform_version: '0.14.8'
   terraform_working_dir: 'fastly/terraform/'
   fastly_service_id: ${{ secrets.FASTLY_SERVICE_ID_PROD }}
   domain: cdn.polyfill.io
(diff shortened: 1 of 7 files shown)


fastly/terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
+# This file is maintained automatically by "terraform init".
```
JakeChampion, line 1: This is a lockfile, the same concept as npm's `package-lock.json` but for terraform, it ensures we use the same versions of the fastly provider plugin

notlee:
This looks good to me :) Does the terraform version need to be updated in the workflows?
https://github.com/Financial-Times/polyfill-service/blob/c116019a94b41be44b5bc933d26285f9e199613f/.github/workflows/deploy-to-production.yml#L6

notlee (approved):

fastly/terraform/.terraform.lock.hcl
alexwilson: Nice!

fastly/terraform/main.tf
```diff
@@ -1,65 +1,31 @@
...
-
-  vcl {
-    name    = "top_pops.vcl"
-    content = file("${path.module}/../vcl/top_pops.vcl")
+  dynamic "vcl" {
+    for_each = local.vcl
+    content {
+      name    = vcl.value
+      main    = vcl.value == "main.vcl" ? true : false
+      content = templatefile("${path.module}/vcl/${vcl.value}", {})
```
alexwilson: You *might* not need templatefile here, but I guess it's handy!

.github/workflows/deploy-to-production.yml
```diff
@@ -3,7 +3,7 @@ on:
   release:
     types: [created]
 env:
-  terraform_version: '0.12.29'
+  terraform_version: '0.14.8'
```
alexwilson, line 6: Nice, guess this plus the lockfile guarantees that things are updated together?

alexwilson (approved):
Looks good, and a nice reduction in code!

.github/workflows/deploy-to-production.yml
JakeChampion: Exactly :-D

fastly/terraform/main.tf
JakeChampion: Yes, we don't really need it as we are not providing any variables to the vcl files right now but I figured I'd keep it around to keep it similar to other Fastly configs in the company.

PR #2566, opened 2021-03-22
Add eclint to the format script to ensure the project adheres to our editorconfig settings

JakeChampion:
When working on this project and looking at some of the VCL files, I noticed they were using a mixture of spaces and tabs for indentation.

I saw the project has an editorconfig file but it was not being used to automatically format the files.

I've added `eclint` to the `npm run fmt` command, it is an editorconfig linter. I then ran the command to fix all the editorconfig issues in the project.


chee (approved):
oh how nice, i didn't know about eclint

PR #2644, opened 2021-09-22
polyfill url builder fixes

romainmenke:
- latest versions weren't showing up
- filter now works after changing version
- last (default) version is now also show explicitly, enabling pinned versions of the last version in the url.

Screenshot 2021-09-22 at 09 12 44
Screenshot 2021-09-22 at 09 12 52

### DIFF
--- a/src/assets/js/index.js
+++ b/src/assets/js/index.js
@@ -122,13 +122,6 @@ const defaultPolyfillBundleOptions = {
 
 const polyfillBundleOptions = Object.assign({}, defaultPolyfillBundleOptions);
 
-const featuresCache = {};
-const featuresRows = [...document.querySelectorAll("#features-list [data-feature-name]")];
-
-featuresRows.forEach(node => {
-	featuresCache[node.dataset.featureName.toLowerCase()] = node;
-});
-
 const renderFeatureList = ({ polyfillAliases, polyfills }) => {
 	let html = "";
 	for (const item of polyfillAliases) {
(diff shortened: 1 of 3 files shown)


src/assets/js/index.js
```diff
@@ -122,13 +122,6 @@ const defaultPolyfillBundleOptions = {
 
 const polyfillBundleOptions = Object.assign({}, defaultPolyfillBundleOptions);
 
-const featuresCache = {};
-const featuresRows = [...document.querySelectorAll("#features-list [data-feature-name]")];
-
-featuresRows.forEach(node => {
-	featuresCache[node.dataset.featureName.toLowerCase()] = node;
-});
```
romainmenke, line 130: This was keeping references to the original list, breaking the filter after changing the version

src/assets/js/index.js
```diff
@@ -196,12 +189,15 @@ if (libraryVersionInput) {
 			polyfillBundleOptions.version = version;
 			updatePolyfillBundle(polyfillBundleOptions);
 			createTooltips();
+
+			// reset the polyfill name filter.
+			document.querySelector("#filter").value = '';
```
romainmenke, line 194: filter reset as it is not applied automatically after version switch

src/assets/v3/library.11ty.js
```diff
@@ -60,7 +62,7 @@ class Test {
 module.exports = Test;
 
 async function getPolyfillNamesFrom(libraryVersion) {
-	const library = require(`polyfill-library-${libraryVersion}`);
+	const library = libraryVersion === latestLibraryVersion ? require(`polyfill-library`) : require(`polyfill-library-${libraryVersion}`);
```
romainmenke, line 65: This threw an error as `polyfill-library-3.108.0` was never found.

romainmenke: Just saw that `"polyfill-library-3.108.0": "npm:polyfill-library@3.108.0",` was added to package.json.

So it can be removed again.

JakeChampion: This is a good addition to keep around 👍

JakeChampion (approved):
Thanks for all your contributions to polyfill.io and related projects @romainmenke 🙌

chee:
> Thanks for all your contributions to polyfill.io and related projects @romainmenke raised_hands

truly!

PR #2649, opened 2021-10-27
Add a PurgeGroup surrogate-key to every response and purge on deploy

JakeChampion:
This is an implementation of https://yann.mandragor.org/posts/purge-group-pattern/

This adds 1 of 1000 possible PurgeGroup surrogate keys to the response, which will then be used during deployment to slowly purge the cache and avoid a thundering-herd of requests going to the polyfill.io origin servers.

When a deployment to production happens:
- Scale the backends up
- Soft-purge each purge-group and sleep inbetween the purges to avoid the thundering-herd
- Sleep some more to let the caches build back up
- Scale the backends to original size and amount

### DIFF
--- a/fastly/vcl/polyfill-service.vcl
+++ b/fastly/vcl/polyfill-service.vcl
@@ -157,6 +157,13 @@ sub vcl_fetch {
 		call breadcrumb_fetch;
 	}
 
+	# https://yann.mandragor.org/posts/purge-group-pattern/
+	if (beresp.http.Surrogate-Key && !std.strstr(beresp.http.Surrogate-Key, "PurgeGroup")) {
+		set beresp.http.Surrogate-Key = beresp.http.Surrogate-Key " PurgeGroup" randomint(0, 999);
+	} else if (!beresp.http.Surrogate-Key) {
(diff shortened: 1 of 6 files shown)


fastly/vcl/polyfill-service.vcl
```diff
@@ -157,6 +157,13 @@ sub vcl_fetch {
 		call breadcrumb_fetch;
 	}
 
+	# https://yann.mandragor.org/posts/purge-group-pattern/
+	if (beresp.http.Surrogate-Key && !std.strstr(beresp.http.Surrogate-Key, "PurgeGroup")) {
+		set beresp.http.Surrogate-Key = beresp.http.Surrogate-Key " PurgeGroup" randomint(0, 999);
+	} else if (!beresp.http.Surrogate-Key) {
+		set beresp.http.Surrogate-Key = "PurgeGroup" randomint(0, 999);
```
chee, line 164: i had to check the fastly docs to see if `randomint` was inclusive

chee (approved):

chee (approved):

PR #2658, opened 2021-12-21
Disable request logging when DISABLE_REQUEST_LOGGING is set

Menci:
Currently even if I set `NODE_ENV` to `production`, it logs each request. I think it's unnecessary.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -9,9 +9,12 @@ dotenv.config();
 const options = {
 	log: console,
 	name: "Origami Polyfill Service",
-	requestLogFormat: process.env.DISABLE_REQUEST_LOGGING ? null : undefined,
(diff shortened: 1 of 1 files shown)


JakeChampion (changes requested):
Hi @Menci - Could you change the code to only set `requestLogFormat` to null if another environment variable is set and not use NODE_ENV, for example we could use `process.env.DISABLE_REQUEST_LOGGING` instead as that is more explicit in it's intent

E.G. Something like:
```js
const options = {
...
};

if (process.env.DISABLE_REQUEST_LOGGING) {
    options.requestLogFormat = null;
}
```

JakeChampion (approved):

Menci:
The CI checks failed since `npm ci` errors. Could you please re-run the checks?

PR #2662, opened 2022-01-20
remove polyfill-library versions which are not being requested

JakeChampion:
We are hitting the size limit for an app of heroku and need to save space
We checked the requests to see what versions were being requested and remove all the ones which got 0 requests


fastly/terraform/dev_override.tf
```diff
@@ -40,7 +40,7 @@ resource "fastly_service_v1" "app" {
     first_byte_timeout    = 120000
     between_bytes_timeout = 120000
     error_threshold       = 0
-    shield                = "dca-dc-us"
+    shield                = "iad-va-us"
```
JakeChampion, line 43: This change had to be include because DCA no longer exists - the original PR for the DCA change was https://github.com/Financial-Times/polyfill-service/pull/2659

PR #2663, opened 2022-01-21
add latest version of polyfill-library

### LINKED ISSUE #1137: Failed to execute 'appendChild' on 'Node'
```
// generate sandboxed iframe
	var vbody = document.appendChild(document.createElement('body'));
	var frame = vbody.appendChild(document.createElement('iframe'));
```

```
Uncaught DOMException: Failed to execute 'appendChild' on 'Node': Only one element on document allowed.
```

An error is occur when using the polyfill.

### LINKED ISSUE #2652: Service 503-ing for hasOwnProperty
## What

It appears that while the basic polyfill returns successfully, polyfills containing `hasOwnProperty` are 503-ing.

## Details

Visit:
https://polyfill.io/v3/polyfill.min.js?features=hasOwnProperty

JakeChampion:
Fixes https://github.com/Financial-Times/polyfill-service/issues/2652

This updates includes these changes from polyfill-library:

    Use the private _setter and _getter methods for TypedArrays when in IE8 because we can't use add square-bracket notation support due to the lack of getters and setters in IE8 ab4868c
    Adds Node.prototype.isSameNode polyfill 372d401
    fix arraybuffer polyfill to not error in ie8 95d1b47
    use an object which has no prototype so that the only properties on the object are it's own and not those from Object.prototype -- Fixes #1137 #1137
    Adds Promise.any, Promise.allSettled, and AggregateError polyfills 21967a1
    Adds Array.prototype.at, String.prototype.at, and TypedArray.prototype.at polyfills 29c6814
    Improves iterator tests for Promise.any, Promise.allSettled, and AggregateError polyfills 999dbce


src/assets/v3/library.11ty.js
```diff
@@ -3,7 +3,7 @@
 const _ = require("lodash");
 const snakeCase = _.snakeCase;
 
-const latestLibraryVersion = "3.109.0";
+const latestLibraryVersion = require("polyfill-library/package.json").version;
```
JakeChampion, line 6: This change means we won't need to update this line when we come to update the version of polyfill-library in the future

notlee (approved):

PR #2674, opened 2022-03-25
Upgrade polyfill-library to 3.111.0

jdeniau:
🤞


server/routes/v3/polyfill.js
```diff
@@ -31,6 +31,7 @@ const polyfillio_3_103_0 = require("polyfill-library-3.103.0");
 const polyfillio_3_104_0 = require("polyfill-library-3.104.0");
 const polyfillio_3_108_0 = require("polyfill-library-3.108.0");
 const polyfillio_3_109_0 = require("polyfill-library-3.109.0");
+const polyfillio_3_110_1 = require("polyfill-library-3.110.1");
```
JakeChampion, line 34: This new variable will need to be added into the array on line 104 of this file 👍

JakeChampion (changes requested):
Thanks for much for this @jdeniau, there is one blocking change which will need to be done before this can be merged and released 👍 

https://github.com/Financial-Times/polyfill-service/pull/2674/files#r835157065

server/routes/v3/polyfill.js
jdeniau: Done. 
It seems that sometime the older version is not set in the files so I tried to add 3.110.1 everywhere, but I don't know if it is really expected ¯\_(ツ)_/¯

JakeChampion (approved):

server/routes/v3/polyfill.js
JakeChampion: That would have been an error from me 😔 thank you for finding and fixing all those

jdeniau:
@JakeChampion Is there something that I can do to release this onto the cdn ? 🤔 (It seems that is still the 3.110.1 in production)

JakeChampion:
> @JakeChampion Is there something that I can do to release this onto the cdn ? 🤔 (It seems that is still the 3.110.1 in production)

I've started the process to release this to production now 👍

PR #2700, opened 2022-04-13
move away from using a director on c@e

JakeChampion:
directors are going to be disabled on compute-at-edge as they are
not yet actually meant to be used on compute-at-edge


chee (approved):
ok they DEFINITELY asked you to do this

PR #2710, opened 2022-09-20
upgrade compute at edge library

wheresrhys:
to fix vulnerability SEC-35031

### DIFF
--- a/.github/workflows/deploy-to-dev-and-test.yml
+++ b/.github/workflows/deploy-to-dev-and-test.yml
@@ -42,7 +42,7 @@ jobs:
             ${{ runner.os }}-
       - run: npm ci
         if: steps.witness.outputs.cache-hit != 'true'
-      - run: npm ci && npm run build
+      - run: npm ci && npm run build && pwd && ls && ls ./pkg
         working-directory: fastly/c-at-e
       - name: 'Terraform Format'
         if: steps.witness.outputs.cache-hit != 'true'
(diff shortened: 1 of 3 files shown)


notlee (approved):

notlee:
Ah so we're using `@fastly/js-compute` to build a package to `fastly/c-at-e/pkg`:
https://github.com/Financial-Times/polyfill-service/blob/d1cad0a5f0617133edec2302a48deb8e183a509a/fastly/c-at-e/package.json#L8

But that was removed in 0.3.0:
https://github.com/fastly/js-compute-runtime/pull/108

Looks like we can use the fastly cli instead:
https://developer.fastly.com/reference/cli/compute/

notlee:
Good morning @JakeChampion. Do you mind if I ask you a polyfill.io / Fastly c@e question?

Looks like to upgrade `js-compute-runtime` we're going to need to update the [build script](https://github.com/Financial-Times/polyfill-service/blob/d1cad0a5f0617133edec2302a48deb8e183a509a/fastly/c-at-e/package.json#L8) so it uses the fastly cli, since creating a tarball was removed from [js-compute-runtime](https://github.com/fastly/js-compute-runtime/pull/108).

Running `fastly compute build` locally within `fastly/c-at-e` creates `pkg/polyfill-service-c-at-e.tar.gz` 🙌 So I'm thinking I'll update the Github action to use `fastly/compute-actions/setup@main` so fastly's cli is installed, and update the build script there. Just wanted to check with you that's sensible? I'm not sure its output is equivalent to `js-compute-runtime bin/index.js bin/main.wasm`.

notlee:
See https://github.com/Financial-Times/polyfill-service/pull/2712/

PR #2712, opened 2022-09-23
move to honojs instead of flight-path

JakeChampion:
[honojs](https://honojs.dev/) is a small and very fast web-server library built primarily for edge-compute platforms such as Fastly and Cloudflare Workers.

I have a lot more experience using honojs now than I do flight-path and thought I'd take a moment to migrate polyfill.io over to it as the migration is rather low effort due to flight-path having a similar API ☺️

### DIFF
--- a/fastly/c-at-e/src/index.js
+++ b/fastly/c-at-e/src/index.js
@@ -1,115 +1,87 @@
 /// <reference types="@fastly/js-compute" />
-import { Router } from "flight-path";
 import UA from "@financial-times/polyfill-useragent-normaliser/lib/normalise-user-agent-c-at-e.js";
 import { normalise_querystring_parameters_for_polyfill_bundle } from "./normalise-query-parameters.js";
 import useragent_parser from "@financial-times/useragent_parser/lib/ua_parser-c-at-e.js";
 import fetchWithFailover from "./fetch-with-failover.js";
+import { Hono } from 'hono'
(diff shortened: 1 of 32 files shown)


fastly/c-at-e/src/index.js
```diff
@@ -1,115 +1,87 @@
...
-      method: request.method
-    });
-    return response.send(backendResponse);
-  }
-
-  const host = request.headers.host;
+app.get("*", handler);
+app.head("*", handler);
+app.options("*", handler);
+app.all("*", c => {
```
notlee, line 77: Looks like this will return 405 for "FASTLYPURGE" & "PURGE" methods now ([previously allowed methods](https://github.com/Financial-Times/polyfill-service/pull/2712/files#diff-2d09e3e5de1c0abcc550bfdda8ffa7ab8cb29460637152b05b22f01a0c3c7ed9L8)), is that true/intentional?

JakeChampion: That's true, I can add back support for individual URL purges if we want them. The way I've been purging the site is via the fastly purge API on api.fastly.com and not with the purge http methods

notlee: Happy to purge through the fastly api :)

JakeChampion:
@notlee everything is passing now on ci 🎉

notlee (approved):

PR #2774, opened 2023-10-20
Rewrite polyfill.io from js to rust

JakeChampion:
This drastically reduces our memory and cpu consumption. With this rewrite, memory usage is now ~10 times lower and cpu usage is ~25 times lower than the javascript version.


Lanchili:
Rust is good， but CONTRIBUTING Guide also need update🤣

PR #2790, opened 2023-11-30
Record RTT and conn_type for globe visualisation

triblondon:
In conversation with Hooman, we're trying to reduce noise from the stream of data we get for the globe visualisation.  This change allows us to record whether the request was over a mobile connection and it's minimum RTT so we can filter out egregious requests where the min RTT is not possible over the distance reported.

### DIFF
--- a/fastly/terraform/modules/service-vcl/vcl/fastly-devhub-globe.vcl
+++ b/fastly/terraform/modules/service-vcl/vcl/fastly-devhub-globe.vcl
@@ -24,11 +24,13 @@ table globeviz_known_browsers {
 //    LCY             Name of Fastly POP (3 char)
 //    51.43,-0.23     Client location (var length)
 //    17234           Duration (var length)
+//    42426           Minimum observed RTT (var length)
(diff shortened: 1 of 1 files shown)


JakeChampion (approved):

fastly/terraform/modules/service-vcl/vcl/fastly-devhub-globe.vcl
```diff
@@ -55,7 +58,8 @@ sub vcl_log {
       substr(fastly_info.state, 0, 1) " "
       regsuball(req.proto, "[^\d.]", "") " "
       regsuball(tls.client.protocol, "[^\d.]", "") " "
-      table.lookup(globeviz_known_browsers, client.browser.name, "Z")
+      table.lookup(globeviz_known_browsers, client.browser.name, "Z") " "
+      if (client.socket.conn_type == "mobile", "M", "-")
```
JakeChampion, line 62: it's located at `client.geo.conn_type` -- https://developer.fastly.com/reference/vcl/variables/geolocation/client-geo-conn-type/

